PR #41, opened 2023-04-27
Enhance filter parser

yhabteab:
After all the testing we did last week, it turns out that the current implementation of the filter `Parser` is very buggy and sometimes crashes with very simple filter strings. This PR solves this by using [`goyacc`](https://pkg.go.dev/golang.org/x/tools/cmd/goyacc) generated `Parser`. This is similar to how we parse Icinga2 `DSL` with `flex/bison`. The grammar rules can be found in the `parser.y` file and you can find the `Lexer` and other parsing utils in the `lexer.go` file.

For local testing/debugging purposes you can re-build the parser with the following command (`go generate ./...`). Since [Go 1.24](https://tip.golang.org/doc/go1.24#tools), you don't even need to pre-install the `goyacc` tool into your GOBIN, instead it's been added as a tool dependency to the go.mod file, so you can just run `go mod tidy` to download it like any other missing dependency. It won't install it to your GOBIN, but it will be available in the module cache and can be used to generate the parser.

The `go generate` command will also generate a `parser.output` file which contains all the parser states and is useful to resolve some `shift/reduce` conflicts.


internal/filter/lexer.go
```diff
@@ -0,0 +1,149 @@
...
+
+	defer func() {
+		// All the grammar rules panics when encountering any errors while reducing the filter rules, so try
+		// to recover from it and return an error instead. Since we're used a named return values, we can set
+		// the err value even in deferred function. See https://go.dev/blog/defer-panic-and-recover
+		if r := recover(); r != nil {
+			lex.err = errors.New(fmt.Sprint(r))
+		}
+
+		err = lex.err
```
oxzi, line 39: In case of an error, the `lex.rule` should be cleared as otherwise an incomplete result will be returned.
For example, see this testing output of an invalid expression of mine:

```
    parser_test.go:188: 
        	Error Trace:	/home/apenning/git/github.com/Icinga/icinga-notifications/internal/filter/parser_test.go:188
        	Error:      	Expected nil, but got: &errors.errorString{s:"1:120 (119): syntax error: unexpected T_EQUAL, expecting \")\""}
        	Test:       	TestParser/ParseChain
        	Messages:   	There should be no errors but got: 1:120 (119): syntax error: unexpected T_EQUAL, expecting ")"
    parser_test.go:190: 
        	Error Trace:	/home/apenning/git/github.com/Icinga/icinga-notifications/internal/filter/parser_test.go:190
        	Error:      	Not equal: 
        	            	expected: *filter.Condition(&filter.Condition{op:"=", column:"foo", value:"bar"})
        	            	actual  : *filter.Exists(&filter.Exists{column:"foo"})
        	Test:       	TestParser/ParseChain
```

internal/filter/parser.y
```diff
@@ -0,0 +1,186 @@
...
+		if $1 != "" {
+			// NewChain is only going to return an error if an invalid operator is specified, and since
+			// we explicitly provide the NONE operator, we don't expect an error to be returned.
+			$$, _ = NewChain(NONE, $2)
+		} else {
+			$$ = $2
+		}
+
+		yylex.(*Lexer).rule = $$
+	}
```
oxzi, line 132: Semicolon missing?

internal/filter/parser.y
```diff
@@ -0,0 +1,186 @@
...
+
+condition_expr: "(" filter_chain ")"
```
oxzi, line 134: By resolving brackets at this point, brackets around `identifier comparison_op identifier` (i.e. `(foo)=(bar)`) are not possible.
However, I don't know if this input expression should even be accepted.

oxzi (commented):
I just went a bit over the code and needed to educate myself about yacc again. Feel free to argue on each of my comments.

internal/filter/parser.y
yhabteab: Our IPL web parser does not accept such filter strings, so it's up to @julianbrost whether to accept such here or not:
```php
ipl\Web\Filter\ParseException : Invalid filter "(foo)=(bar)", unexpected = at pos 6: Expected logical operator
 /usr/local/src/ipl-web/src/Filter/Parser.php:561
```

internal/filter/lexer.go
```diff
@@ -0,0 +1,149 @@
...
+	lex := new(Lexer)
+	lex.IsIdentRune = isIdentRune
+	lex.Init(strings.NewReader(expr))
+
+	// scanner.Init sets the error function to nil, therefore, we have to register
+	// our error function after the scanner initialization.
+	lex.Scanner.Error = lex.ScanError
+
+	// Enable parsers error verbose to get more context of the parsing failures
+	yyErrorVerbose = true
```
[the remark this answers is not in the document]
yhabteab: No idea 🤷 how to set this with `goyacc`, but for now I will just use `init()` as discussed offline.

internal/filter/parser.y
```diff
@@ -0,0 +1,183 @@
...
+		yylex.(*Lexer).rule = $$
+	}
+	;
+
+filter_chain: conditions_expr logical_op maybe_negated_condition_expr
+	{
+		$$ = reduceFilter($2, $1, $3)
+	}
+	| conditions_expr %prec PREFER_SHIFTING_LOGICAL_OP
+	;
```
julianbrost, line 113: Why is there an asymmetry in the operands regarding possible negation?

internal/filter/parser.y
```diff
@@ -0,0 +1,183 @@
...
+
+conditions_expr: maybe_negated_condition_expr logical_op maybe_negated_condition_expr
+	{
+		$$ = reduceFilter($2, $1, $3)
+	}
+	| maybe_negated_condition_expr %prec PREFER_SHIFTING_LOGICAL_OP
+	;
```
julianbrost, line 120: Am I missing something or does this limit a `filter_chain` to at most 3 `condition_expr`?

julianbrost: Yes, it does, which limits `filter_rule` to 6 `condition_expr`, so parsing `a & b & c & d & e & f & g` fails (spaces for readability, parsing fails with and without them).

internal/filter/parser.y
julianbrost: Ah, I got confused by the very similar names `conditions_expr` and `condition_expr`.

internal/filter/types.go
```diff
@@ -1,188 +1,236 @@
...
 
-func (less *LessThan) Eval(filterable Filterable) (bool, error) {
-	match, err := filterable.EvalLess(less.column, less.value)
-	if err != nil {
-		return false, err
+		match = filterable.EvalExists(c.column) && !match
 	}
 
 	return match, nil
 }
```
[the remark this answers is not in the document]
yhabteab: I don't understand why this is a problem now, I haven't changed anything since the previous version. I have just have reduce all the types into a single method.

internal/filter/lexer.go
```diff
@@ -0,0 +1,158 @@
...
+	"errors"
+	"fmt"
+	"regexp"
+	"strings"
+	"text/scanner"
+)
+
+// identifiersMatcher contains a compiled regexp and is used by the Lexer to match filter identifiers.
+// Currently, it allows to match any character except a LogicalOp and CompOperator.
+var identifiersMatcher = regexp.MustCompile("[^!&|~<>=()]")
```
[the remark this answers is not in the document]
yhabteab: I don't really understand on why don't doing it the way it is, and I don't want to list all the valid chars because I don't know them all.

internal/filter/parser.y
yhabteab: Yes, I noticed that last time too, but didn't have time to push all the fixes.

internal/filter/lexer.go
```diff
@@ -0,0 +1,158 @@
...
+
+	yyParse(lex)
+
+	return lex.rule, lex.err
+}
+
+// Lexer is used to tokenize the filter input into a set literals.
+// This is just a wrapper around the Scanner type and implements the yyLexer interface used by the parser.
+type Lexer struct {
+	scanner.Scanner
```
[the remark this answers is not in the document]
yhabteab: Go white spaces are just like in PHP and the IPL parser also skips all these characters, just like this lexer does, so I won't do anything in this regard. For other customizations, see the inline comments in the `Parse()` function.

internal/filter/parser.y
yhabteab: Should be fixed with https://github.com/Icinga/icinga-notifications/pull/41/commits/5484f912d2498beaee1182f23d6dbcdceab82573

internal/filter/types.go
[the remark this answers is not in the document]
julianbrost: It's not like this has magically become a problem in the meantime. It's just that the old code was merged in an early phase and while in a hurry, so it was better to have some working code than fully cleaned up code. But now, since you're reworking large parts of the code, I think it makes sense to clean up these parts as well.

internal/filter/parser.y
yhabteab: If you have a better name that fits this use case, I will accept it.

yhabteab:
Unfortunately :), this won't won't be needed anymore as it was decided to use another filter syntax instead, so closing.

Qoute from https://github.com/Icinga/icinga-notifications/issues/406#issuecomment-4083644053

> You're right, it's not gone completely yet, but escalation conditions are very limited in comparison right now. It's also still the first implementation in use, which we wanted to redo (https://github.com/Icinga/icinga-notifications/pull/41), but that went dormant after these expression were removed for object filters. But I would rather prefer abandoning that PR entirely and use something trivial to parse instead. (Or in other words: putting effort into implementing a parser for is a waste of time to me.)

PR #64, opened 2023-05-10
Write `schedule/group_id` for contact notified histories

yhabteab:
When a contact is notified because of being member of a `Schedule` or a `ContactGroup`, it may not be clear why a particular contact was notified as it has no direct reference to the current incident. With this PR, noma-web can easily identify that the contact was notified due to a matching `ContactGroup` or `Schedule` and rendere a proper message accordingly.

### DIFF
--- a/internal/listener/listener.go
+++ b/internal/listener/listener.go
@@ -399,6 +399,9 @@ func (l *Listener) ProcessEvent(w http.ResponseWriter, req *http.Request) {
 	contactChannels := make(map[*recipient.Contact]map[string]struct{})
 
 	escalationRecipients := make(map[recipient.Key]bool)
+	groupsOrSchedules := make(map[*recipient.Contact]recipient.Key)
+	groupsOrSchedulesWithoutMembers := make(map[recipient.Recipient]bool)
+
 	for escalationID, state := range currentIncident.EscalationState {
 		escalation := l.runtimeConfig.GetRuleEscalation(escalationID)
 		if state.TriggeredAt.Time().IsZero() {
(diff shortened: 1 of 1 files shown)


internal/recipient/recipient.go
```diff
@@ -33,6 +33,24 @@ func (r Key) MarshalText() (text []byte, err error) {
 	}
 }
 
+// CopyNonNil copies non-nil fields from the provided recipient.Key to the current one and
+// returns the modified recipient key.
+func (r Key) CopyNonNil(recipientKey Key) Key {
```
julianbrost, line 38: This function and how you use `recipient.Key` in `incident.HistoryRow` contradicts the current assumption in the code that there's a 1:1 relation between `recipient.Recipient` and `recipient.Key`, as it can be seen in multiple places:

https://github.com/Icinga/icinga-notifications/blob/002d14dc19e6a68bb894d0c9ad8829463c68cc9b/internal/recipient/recipient.go#L16-L21
https://github.com/Icinga/icinga-notifications/blob/002d14dc19e6a68bb894d0c9ad8829463c68cc9b/internal/recipient/recipient.go#L23-L34
https://github.com/Icinga/icinga-notifications/blob/002d14dc19e6a68bb894d0c9ad8829463c68cc9b/internal/config/runtime.go#L92-L112

I'd want to keep that assumption in place as `recipient.Key` is used as a key in maps for example and that would result in nasty bugs if `recipient.Key` values with more than one member set would ever be used in a map key.

So I think it would be better to add the three columns as explicit members in `incident.HistoryRow` if it's now intended that more than one is set to a non-null value.

internal/listener/listener.go
```diff
@@ -447,7 +450,14 @@ func (l *Listener) ProcessEvent(w http.ResponseWriter, req *http.Request) {
...
 
 			if !managed || state.Role > incident.RoleRecipient {
-				for _, c := range escalationRecipient.Recipient.GetContactsAt(ev.Time) {
+				r := escalationRecipient.Recipient
+				_, isContact := r.(*recipient.Contact)
+				contacts := r.GetContactsAt(ev.Time)
+				for _, c := range contacts {
+					if !isContact {
+						groupsOrSchedules[c] = recipient.ToKey(r)
+					}
```
julianbrost, line 459: What's supposed to happen if the same contact if references multiple times, i.e. directly, via a group and via a schedule? This would just arbitrarily pick a group or schedule. We could write all of that information into the database, but that would probably mean writing multiple rows for the same notification, i.e. the web interface would have to handle and display this probably, so I don't want to decide this on my own.

No matter how we do it, a comment in the schema file how these rows are to be interpreted would be great.

julianbrost (changes requested):

Issue #63, opened 2023-05-10
Trigger time-based escalations without external events


julianbrost:
At the moment, time-based escalations (something like `incident_age>=1h`) will only trigger if a new event is processed (which is not guaranteed to happen). Therefore, we have to trigger some escalations internally.

The simple fix for this would be to check all not yet escalations against open incidents periodically, but that's inefficient and introduces delays.

The nicer, but also more complex to implement solution for this would be to extract information about relevant times where the result of evaluating the escalation condition might change. Take the escalation condition `(incident_severity>=critical&incident_age>=1h)|incident_age>=3h`. Extracting each value `incident_age` is compared against gives `1h` and `3h`, therefore, the only times when it's interesting to reevaluate this condition would be `incident.started_at + 1h` and `incident.started_at + 3h` (and if the severity changes, but that is already checked when processing that event that changes this severity).

While I expect `>=` to be the most useful operator for comparing `incident_age`, we also have to consider what happens when using `=` or `>`. If we evaluate `incident_age=1h` at exactly `incident.started_at + 1h`, this will trigger (which we can't reliably achieve with periodic evaluation). Similarly, if we evaluate `incident_age>1h` at that very same time, it would not yet trigger, even though it would a tiny moment later, so we might want to evaluate it bit before and after as well (a bit = 1ns should do the trick).

PR #67, opened 2023-05-11
Restructure codes

yhabteab:
Moves all `IncidentHistory` related codes from `incident.go` to a new `history.go` file and the `ContactRole` type to the `contact_role.go` file in the same package. The `Incident#GetCurrent()` and all `CurrentIncidents` related are also moved to the `incidents.go` file. And last but not least, all current incident event processing has been extracted from `Listener#ProcessEvent()` into the `incident` package, which should now make it easier to trigger an escalation and notify contacts without getting an external event.

**This PR does not change anything about the previous functionality, it just restructures the code and moves it back and forth.**

refs #63


internal/incident/incident.go
```diff
@@ -29,17 +31,30 @@ type Incident struct {
 	Rules           map[ruleID]struct{}
 	Recipients      map[recipient.Key]*RecipientState
 
-	incidentRowID int64
+	Icingaweb2Url string
```
julianbrost, line 34: That `Icingaweb2Url` doesn't make much sense here. It's passed as an argument to `(*Incident).ProcessEvent()` which then sets this member, then calls `(*Incident).notifyContacts()` which then accesses the member to pass it as an argument to other functions.

Either continue the URL as a parameter or add a member for the full config (and initialize that one in a proper place rather than as a side effect of `ProcessEvent()`).

julianbrost (changes requested):
According to the commit message, 6bddc8534a78d435069493b7433bf2930d11f935 is supposed to move code, but that commit itself mostly adds code, so I think something went wrong when splitting this into individual commits. 3e46a97fe0ec280f826a9bccb0b283dde53197ab is also quite large (I think it also accidentally contains some changes that should be in the aforementioned commit) and hard to follow/review as it moves and restructures code at the same time. So I'd highly appreciate it if you'd break these commits down into smaller chunks and separate moving code from changes to the code.

internal/incident/incident.go
```diff
@@ -60,351 +74,469 @@ func (i *Incident) HasManager() bool {
...
+		EventID:          utils.ToDBInt(ev.ID),
+		Type:             RecipientRoleChanged,
+		Time:             types.UnixMilli(time.Now()),
+		NewRecipientRole: newRole,
+		OldRecipientRole: oldRole,
+		Message:          utils.ToDBString(ev.Message),
 	}
 
-	incidentRow := &IncidentRow{ID: currentIncident.incidentRowID, RecoveredAt: types.UnixMilli(currentIncident.RecoveredAt)}
-	_, err := currentIncident.db.NamedExec(`UPDATE "incident" SET "recovered_at" = :recovered_at WHERE id = :id`, incidentRow)
```
julianbrost, line 385: Looks like this update got lost while refactoring.

When updating this PR, please keep the commits up to (including) c21dac071f257ae79dacba0a656671a95a129a59 as is as I've already looked over those.

julianbrost (changes requested):
This time, please keep the commits up to and including 33b9a6cea2ecfde4524766e2cbc972e0b8a58eca as is.

internal/incident/incident.go
```diff
@@ -60,351 +73,475 @@ func (i *Incident) HasManager() bool {
...
-	stmt := utils.BuildInsertStmtWithout(i.db, historyRow, "id")
-	if fetchId {
-		historyId, err := utils.InsertAndFetchId(i.db, stmt, historyRow)
-		if err != nil {
-			return types.Int{}, err
-		}
+	defer func() {
+		i.Unlock()
+		i.runtimeConfig.RUnlock()
+	}()
```
julianbrost, line 84: Just use two separate `defer` statements, you don't gain much from putting this in a function. Also, I'd put them right after the corresponding lock statement, because that gives you the teardown order you typically want (though that doesn't really matter here).

yhabteab:
I will update the previous commits.

julianbrost (changes requested):
Some things I noticed so far.

yhabteab:
```diff
diff --git a/internal/incident/incident.go b/internal/incident/incident.go
index 7ab9d65..c4a73ca 100644
--- a/internal/incident/incident.go
+++ b/internal/incident/incident.go
@@ -76,12 +76,10 @@ func (i *Incident) HasManager() bool {
 // ProcessEvent processes the given event for the current incident.
 func (i *Incident) ProcessEvent(ev event.Event, created bool) error {
 	i.Lock()
-	i.runtimeConfig.RLock()
+	defer i.Unlock()
 
-	defer func() {
-		i.Unlock()
-		i.runtimeConfig.RUnlock()
-	}()
+	i.runtimeConfig.RLock()
+	defer i.runtimeConfig.RUnlock()
 
 	causedBy, err := i.processIncidentAndSourceSeverity(ev, created)
 	if err != nil {
@@ -101,9 +99,7 @@ func (i *Incident) ProcessEvent(ev event.Event, created bool) error {
 	}
 
 	// Re-evaluate escalations based on the newly evaluated rules.
-	if err := i.evaluateEscalations(); err != nil {
-		return err
-	}
+	i.evaluateEscalations()
 
 	return i.notifyContacts(&ev, causedBy)
 }
@@ -280,7 +276,7 @@ func (i *Incident) evaluateRules(eventID int64, causedBy types.Int) (types.Int,
 			if err != nil {
 				i.logger.Errorln(err)
 
-				return causedBy, err
+				return types.Int{}, err
 			}
 
 			if insertedID.Valid && !causedBy.Valid {
@@ -294,7 +290,7 @@ func (i *Incident) evaluateRules(eventID int64, causedBy types.Int) (types.Int,
 
 // evaluateEscalations evaluates this incidents rule escalations if they aren't already.
 // Returns error on database failure.
-func (i *Incident) evaluateEscalations() error {
+func (i *Incident) evaluateEscalations() {
 	if i.EscalationState == nil {
 		i.EscalationState = make(map[int64]*EscalationState)
 	}
@@ -337,8 +333,6 @@ func (i *Incident) evaluateEscalations() error {
 			}
 		}
 	}
-
-	return nil
 }
 
 // notifyContacts evaluates the incident.EscalationState and checks if escalations need to be triggered
@@ -402,7 +396,7 @@ func (i *Incident) notifyContacts(ev *event.Event, causedBy types.Int) error {
 			escalationRecipients[escalationRecipient.Key] = true
 
 			if !managed || state.Role > RoleRecipient {
-				for _, c := range escalationRecipient.Recipient.GetContactsAt(time.Now()) {
+				for _, c := range escalationRecipient.Recipient.GetContactsAt(ev.Time) {
 					if contactChannels[c] == nil {
 						contactChannels[c] = make(map[string]struct{})
 					}
@@ -424,7 +418,7 @@ func (i *Incident) notifyContacts(ev *event.Event, causedBy types.Int) error {
 
 		isEscalationRecipient := escalationRecipients[recipientKey]
 		if !isEscalationRecipient && (!managed || state.Role > RoleRecipient) {
-			for _, contact := range r.GetContactsAt(time.Now()) {
+			for _, contact := range r.GetContactsAt(ev.Time) {
 				if contactChannels[contact] == nil {
 					contactChannels[contact] = make(map[string]struct{})
 				}
```

internal/incident/history.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/icinga/icinga-notifications/internal/rule"
+	"github.com/icinga/icinga-notifications/internal/utils"
+	"github.com/icinga/icingadb/pkg/types"
+	"time"
+)
+
+// Sync initiates an *incident.IncidentRow from the current incident state and syncs it with the database.
+// Before syncing any incident related database entries, this method should be called at least once.
+// Returns an error on db failure.
+func (i *Incident) Sync() error {
```
julianbrost, line 16: Not sure why this function is in this file now. Doesn't really have anything to do with history.

internal/incident/history.go
```diff
@@ -0,0 +1,159 @@
...
+	_, err := i.db.NamedExec(stmt, state)
+	if err != nil {
+		return types.Int{}, fmt.Errorf("failed to insert incident rule escalation state: %w", err)
+	}
+
+	return i.AddHistory(hr, true)
+}
+
+// AddEvent Inserts incident history record to the database and returns an error on db failure.
+func (i *Incident) AddEvent(ev *event.Event) error {
```
julianbrost, line 69: Also not really inserting history, this is just the incident <-> event mapping.

internal/incident/history.go
julianbrost, line 1: The `history.go` file name is somewhat misleading as many functions do much more than writing history. So for example, the main purpose of the `AddRecipient()` function is to add some recipient to the current incident state. It also happens to write a history entry for that action, but that's not the main purpose of that function.

julianbrost (changes requested):
Unless you say there's something fundamentally broken in the existing commit history, I'd suggest doing future changes in this PR in new commits.

internal/listener/listener.go
```diff
@@ -144,378 +143,13 @@ func (l *Listener) ProcessEvent(w http.ResponseWriter, req *http.Request) {
...
-
-	if oldSourceSeverity == ev.Severity {
-		msg := fmt.Sprintf("%s: ignoring superfluous %q state event from source %d", obj.DisplayName(), ev.Severity.String(), ev.SourceId)
-		_, _ = fmt.Fprintln(w, msg)
-
-		l.logger.Warnln(msg)
+	if err := currentIncident.ProcessEvent(ev, created); err != nil {
+		_, _ = fmt.Fprintln(w, err)
 		return
 	}
```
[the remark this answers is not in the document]
yhabteab: > Are there any errors where you wouldn't want to log it?

No! But usually the logs contain more information about the errors than a simple http response. That's why I chose this way, because the individual functions that generate these errors can provide more details than here in listener and decide themselves which error they pass to the caller. The logging generally gets more naturell and nice with this [PR](https://github.com/Icinga/icinga-notifications/pull/76).

internal/incident/history.go
yhabteab: Well, which name do you suggest?

internal/listener/listener.go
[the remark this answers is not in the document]
julianbrost: My thinking was that here, the primary error reporting mechanism should be the log, not the HTTP response. If there's a problem with the database, this should always show up in our logs, the user shouldn't have to search logs of all sources for errors (if they'd even log them in a useful way).

Primarily, I noticed this because it is inconsistent with other error handling in the same function:
https://github.com/Icinga/icinga-notifications/blob/3a6912e66b108288387dabbe022f402fd417381d/internal/listener/listener.go#L116-L123

In general, I just want to ensure that every error (except those resulting in 400 Bad Request due to bad input) to show up in the log. If it was done here, it would be easier to verify (for me as part of the review), but sure, logging where the error first appears will likely result in more useful messages.

internal/incident/history.go
julianbrost: I don't really see what the functions in here have in common, which would be the starting point for finding such a name.

I think I'd just keep all of this inside `incident.go` for now and maybe try (in a later PR) to split this by phase of the event handling, maybe like one file for all code used for ack handling, one for rule/escalation processing, one for notifications (but I haven't put much thought into this, so don't treat this as the perfect suggestion, it's just an initial thought).

julianbrost: Fine for me, very generic, but that way it isn't misleading.

internal/contracts/cotracts.go
julianbrost, line 1: There's a typo in the file name.

julianbrost (changes requested):
That should be the last iteration.

PR #68, opened 2023-05-12
Trigger time based escalations

### LINKED ISSUE #63: Trigger time-based escalations without external events

yhabteab:
fixes #63

### DIFF
--- a/internal/filter/contracts.go
+++ b/internal/filter/contracts.go
@@ -12,4 +12,5 @@ type Filterable interface {
 // Filter is implemented by every filter chains and filter conditions.
 type Filter interface {
 	Eval(filterable Filterable) (bool, error)
+	ExtractConditions() []*Condition
 }
--- a/internal/filter/parser_test.go
+++ b/internal/filter/parser_test.go
@@ -94,34 +94,33 @@ func TestFilter(t *testing.T) {
 
 		rule, err = Parse("foo=bar&bar=foo")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.IsType(t, &All{}, rule)
+		assert.IsType(t, &Chain{}, rule)
 
 		rule, err = Parse("foo=bar|bar=foo")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.IsType(t, &Any{}, rule)
+		assert.IsType(t, &Chain{}, rule)
 
 		rule, err = Parse("!(foo=bar|bar=foo)")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.IsType(t, &None{}, rule)
+		assert.IsType(t, &Chain{}, rule)
 
 		rule, err = Parse("!foo")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-
-		assert.Equal(t, &None{rules: []Filter{NewExists("foo")}}, rule)
+		assert.Equal(t, &Chain{op: NONE, rules: []Filter{NewExists("foo")}}, rule)
 
 		rule, err = Parse("foo")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.Equal(t, NewExists("foo"), rule)
+		assert.Equal(t, &Exists{column: "foo"}, rule)
 
 		rule, err = Parse("!(foo=bar|bar=foo)&(foo=bar|bar=foo)")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
 
-		expected := &All{rules: []Filter{
-			&None{rules: []Filter{
+		expected := &Chain{op: ALL, rules: []Filter{
+			&Chain{op: NONE, rules: []Filter{
 				&Equal{column: "foo", value: "bar"},
 				&Equal{column: "bar", value: "foo"},
 			}},
-			&Any{rules: []Filter{
+			&Chain{op: ANY, rules: []Filter{
 				&Equal{column: "foo", value: "bar"},
 				&Equal{column: "bar", value: "foo"},
 			}},
--- a/internal/filter/parser_test.go
+++ b/internal/filter/parser_test.go
@@ -62,35 +62,43 @@ func TestFilter(t *testing.T) {
 	t.Run("ParserIdentifiesAllKindOfFilters", func(t *testing.T) {
 		rule, err := Parse("foo=bar")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.IsType(t, &Equal{}, rule)
+		expected := &Condition{op: Equal, column: "foo", value: "bar"}
+		assert.Equal(t, expected, rule)
 
 		rule, err = Parse("foo!=bar")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.IsType(t, &UnEqual{}, rule)
+		expected = &Condition{op: UnEqual, column: "foo", value: "bar"}
+		assert.Equal(t, expected, rule)
 
 		rule, err = Parse("foo=bar*")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.IsType(t, &Like{}, rule)
+		expected = &Condition{op: Like, column: "foo", value: "bar*"}
+		assert.Equal(t, expected, rule)
 
 		rule, err = Parse("foo!=bar*")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.IsType(t, &Unlike{}, rule)
+		expected = &Condition{op: UnLike, column: "foo", value: "bar*"}
+		assert.Equal(t, expected, rule)
 
 		rule, err = Parse("foo<bar")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.IsType(t, &LessThan{}, rule)
+		expected = &Condition{op: LessThan, column: "foo", value: "bar"}
+		assert.Equal(t, expected, rule)
 
 		rule, err = Parse("foo<=bar")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.IsType(t, &LessThanOrEqual{}, rule)
+		expected = &Condition{op: LessThanEqual, column: "foo", value: "bar"}
+		assert.Equal(t, expected, rule)
 
 		rule, err = Parse("foo>bar")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.IsType(t, &GreaterThan{}, rule)
+		expected = &Condition{op: GreaterThan, column: "foo", value: "bar"}
+		assert.Equal(t, expected, rule)
 
 		rule, err = Parse("foo>=bar")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.IsType(t, &GreaterThanOrEqual{}, rule)
+		expected = &Condition{op: GreaterThanEqual, column: "foo", value: "bar"}
+		assert.Equal(t, expected, rule)
 
 		rule, err = Parse("foo=bar&bar=foo")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
@@ -115,47 +123,52 @@ func TestFilter(t *testing.T) {
 		rule, err = Parse("!(foo=bar|bar=foo)&(foo=bar|bar=foo)")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
 
-		expected := &Chain{op: ALL, rules: []Filter{
+		expectedChain := &Chain{op: ALL, rules: []Filter{
 			&Chain{op: NONE, rules: []Filter{
-				&Equal{column: "foo", value: "bar"},
-				&Equal{column: "bar", value: "foo"},
+				&Condition{op: Equal, column: "foo", value: "bar"},
+				&Condition{op: Equal, column: "bar", value: "foo"},
 			}},
 			&Chain{op: ANY, rules: []Filter{
-				&Equal{column: "foo", value: "bar"},
-				&Equal{column: "bar", value: "foo"},
+				&Condition{op: Equal, column: "foo", value: "bar"},
+				&Condition{op: Equal, column: "bar", value: "foo"},
 			}},
 		}}
-		assert.Equal(t, expected, rule)
+		assert.Equal(t, expectedChain, rule)
 	})
 
 	t.Run("ParserIdentifiesSingleCondition", func(t *testing.T) {
 		rule, err := Parse("foo=bar")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
 
-		expected := &Equal{column: "foo", value: "bar"}
+		expected := &Condition{op: Equal, column: "foo", value: "bar"}
 		assert.Equal(t, expected, rule, "Parser doesn't parse single condition correctly")
 	})
 
 	t.Run("UrlEncodedFilterExpression", func(t *testing.T) {
 		rule, err := Parse("col%3Cumn<val%3Cue")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.Equal(t, &LessThan{column: "col<umn", value: "val<ue"}, rule)
+		expected := &Condition{op: LessThan, column: "col<umn", value: "val<ue"}
+		assert.Equal(t, expected, rule)
 
 		rule, err = Parse("col%7Cumn=val%7Cue")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.Equal(t, &Equal{column: "col|umn", value: "val|ue"}, rule)
+		expected = &Condition{op: Equal, column: "col|umn", value: "val|ue"}
+		assert.Equal(t, expected, rule)
 
 		rule, err = Parse("col%26umn<=val%26ue")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.Equal(t, &LessThanOrEqual{column: "col&umn", value: "val&ue"}, rule)
+		expected = &Condition{op: LessThanEqual, column: "col&umn", value: "val&ue"}
+		assert.Equal(t, expected, rule)
 
 		rule, err = Parse("col%28umn>val%28ue")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.Equal(t, &GreaterThan{column: "col(umn", value: "val(ue"}, rule)
+		expected = &Condition{op: GreaterThan, column: "col(umn", value: "val(ue"}
+		assert.Equal(t, expected, rule)
 
 		rule, err = Parse("col%29umn>=val%29ue")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.Equal(t, &GreaterThanOrEqual{column: "col)umn", value: "val)ue"}, rule)
+		expected = &Condition{op: GreaterThanEqual, column: "col)umn", value: "val)ue"}
+		assert.Equal(t, expected, rule)
 	})
 }
 
--- a/internal/filter/parser_test.go
+++ b/internal/filter/parser_test.go
@@ -114,7 +114,7 @@ func TestFilter(t *testing.T) {
 
 		rule, err = Parse("!foo")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.Equal(t, &Chain{op: NONE, rules: []Filter{NewExists("foo")}}, rule)
+		assert.Equal(t, &Chain{op: None, rules: []Filter{NewExists("foo")}}, rule)
 
 		rule, err = Parse("foo")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
@@ -123,12 +123,12 @@ func TestFilter(t *testing.T) {
 		rule, err = Parse("!(foo=bar|bar=foo)&(foo=bar|bar=foo)")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
 
-		expectedChain := &Chain{op: ALL, rules: []Filter{
-			&Chain{op: NONE, rules: []Filter{
+		expectedChain := &Chain{op: All, rules: []Filter{
+			&Chain{op: None, rules: []Filter{
 				&Condition{op: Equal, column: "foo", value: "bar"},
 				&Condition{op: Equal, column: "bar", value: "foo"},
 			}},
-			&Chain{op: ANY, rules: []Filter{
+			&Chain{op: Any, rules: []Filter{
 				&Condition{op: Equal, column: "foo", value: "bar"},
 				&Condition{op: Equal, column: "bar", value: "foo"},
 			}},
--- a/internal/filter/types.go
+++ b/internal/filter/types.go
@@ -71,6 +71,19 @@ func (c *Chain) Eval(filterable Filterable) (bool, error) {
 	}
 }
 
+func (c *Chain) ExtractConditions() []*Condition {
+	var conditions []*Condition
+	for _, rule := range c.rules {
+		if _, ok := rule.(*Exists); ok {
+			continue
+		}
+
+		conditions = append(conditions, rule.ExtractConditions()...)
+	}
+
+	return conditions
+}
+
 // CompOperator is a type used for grouping the individual comparison operators of a filter string.
 type CompOperator string
 
@@ -165,6 +178,10 @@ func (c *Condition) Eval(filterable Filterable) (bool, error) {
 	}
 }
 
+func (c *Condition) ExtractConditions() []*Condition {
+	return []*Condition{c}
+}
+
 // Column returns the column of this Condition.
 func (c *Condition) Column() string {
 	return c.column
@@ -177,6 +194,10 @@ func (c *Condition) Value() string {
 
 type Exists struct {
 	column string
+}
+
+func (e *Exists) ExtractConditions() []*Condition {
+	panic("filter exists doesn't support extract conditions")
 }
 
 func NewExists(column string) *Exists {
--- a/internal/filter/types.go
+++ b/internal/filter/types.go
@@ -74,10 +74,6 @@ func (c *Chain) Eval(filterable Filterable) (bool, error) {
 func (c *Chain) ExtractConditions() []*Condition {
 	var conditions []*Condition
 	for _, rule := range c.rules {
-		if _, ok := rule.(*Exists); ok {
-			continue
-		}
-
 		conditions = append(conditions, rule.ExtractConditions()...)
 	}
 
@@ -197,7 +193,7 @@ type Exists struct {
 }
 
 func (e *Exists) ExtractConditions() []*Condition {
-	panic("filter exists doesn't support extract conditions")
+	return nil
 }
 
 func NewExists(column string) *Exists {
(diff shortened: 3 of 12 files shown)


internal/rule/condition.go
```diff
@@ -11,74 +12,92 @@ type EscalationFilter struct {
...
+			}
+
+			if incidentAge < age && (retryAfter <= 0 || retryAfter > age) {
+				retryAfter = age
+			}
+		}
+	}
+
+	return retryAfter
+}
```
julianbrost, line 31: This function is a really good candidate for a doc comment explaining what it does.

julianbrost (commented):
First few comments only, i.e. those until I noticed that I should probably look at #67 first.

internal/incident/incident.go
```diff
@@ -595,6 +643,117 @@ func (i *Incident) processAcknowledgementEvent(ctx context.Context, tx *sqlx.Tx,
...
+
+	if err := i.ReloadRecipients(ctx); err != nil {
+		// The error is already logged within `i.ReloadRecipients()`
+		return
+	}
+
+	if !i.RecoveredAt.IsZero() {
+		// Incident is recovered in the meantime.
+		return
+	}
```
julianbrost, line 675: This check should also be fine before performing any database query, the incident is locked so it shouldn't change.

internal/incident/incident.go
```diff
@@ -354,20 +368,50 @@ func (i *Incident) evaluateEscalations() {
...
 						matched = false
+					} else if !matched {
+						age := cond.ExtractRetryAfter(escalation.Condition.ExtractConditions(), cond.IncidentAge)
+						if age > 0 && (retryAfter <= 0 || retryAfter > age) {
+							retryAfter = time.Until(i.StartedAt.Add(age))
+							i.scheduler.condition = cond
+							i.scheduler.condition.IncidentAge = age
+						} else {
+							i.scheduler.condition = nil
+						}
```
julianbrost, line 414: Please explain what the value in `i.scheduler.condition` is supposed to express. I find it hard to follow what's going on here, `condition` stores an `EscalationFilter` which actually is more like a filter context (i.e. stores the context in which a filter is evaluated), but now it stores the time after which the filter is supposed to be reevaluated?

internal/incident/incidents.go
```diff
@@ -20,6 +21,86 @@ var (
...
+
+			_, err = incident.evaluateRules(childCtx, tx, 0, types.Int{})
+			if err != nil {
+				return err
+			}
+
+			incident.evaluateEscalations()
+			if err = tx.Commit(); err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
yhabteab: No! This just loads all open incidents from the database and evaluates all their rules and escalation without waiting for external events of the incident object.

internal/incident/incident.go
yhabteab: `i.scheduler.condition` just stores the current escalation filter condition from which the `retryAfter` is extracted and modifies it's `IncidentAge` threshold and never stores `the time after which the filter is supposed to be reevaluated?`. This is needed because when you construct the escalation filter each time evaluating the escalations like:
https://github.com/Icinga/icinga-notifications/blob/14f00144b0fa82d5e40e70d3dbd70bb4588ff017/internal/incident/incident.go#L348

The condition won't match even it's retried after the correct **extracted** retry after interval.

julianbrost:
Has a merge conflict now because I've merged #76 recently.

yhabteab:
Used filter: `incident.age >= 2m`

```bash
2023-09-20T11:41:07.272+0200    INFO    listener        Processing event
2023-09-20T11:41:07.274+0200    INFO    incident        Opened incident {"object": "icinga2"}
2023-09-20T11:41:07.277+0200    INFO    incident        Source 1 severity changed from ok to crit       {"object": "icinga2", "incident": "#2"}
2023-09-20T11:41:07.280+0200    INFO    incident        Incident severity changed from ok to crit       {"object": "icinga2", "incident": "#2"}
2023-09-20T11:41:07.283+0200    INFO    incident        Rule "default" matches  {"object": "icinga2", "incident": "#2"}
```

**2 Minutes later:**
```bash
2023-09-20T11:44:07.253+0200    INFO    incident        Rule "default" reached escalation "[S] Office Hours"    {"object": "icinga2", "incident": "#2"}
2023-09-20T11:44:07.264+0200    INFO    incident        Notify contact "Yonas Habteab" via "rocketchat" {"object": "icinga2", "incident": "#2"}
2023-09-20T11:44:07.266+0200    ERROR   incident        Could not find config for channel       {"object": "icinga2", "incident": "#2", "type": "rocketchat"}
```

internal/incident/incidents.go
```diff
@@ -19,6 +20,89 @@ var (
...
+			db:               db,
+			runtimeConfig:    runtimeConfig,
+			StartedAt:        incidentRow.StartedAt.Time(),
+			incidentRowID:    incidentRow.ID,
+			configFile:       configFile,
+			SeverityBySource: map[int64]event.Severity{},
+			EscalationState:  map[escalationID]*EscalationState{},
+			Rules:            map[ruleID]struct{}{},
+			Recipients:       map[recipient.Key]*RecipientState{},
+		}
```
julianbrost, line 48: You're duplicating the struct initialization here and it contains lots of "set to default values" (that primarily is initializing maps), so I think a `NewIncident()` function would help, otherwise changing anything becomes annoying (and if another required field is added, there's a good chance it might be forgotten in one place, which wouldn't happen if it was an argument of that `NewIncident()` function).

internal/incident/incidents.go
```diff
@@ -19,6 +20,89 @@ var (
 	currentIncidentsMu sync.Mutex
 )
 
+// LoadOpenIncidents loads all active (not yet closed) incidents from the database and restores all their states.
+// Returns error ony database failure.
+func LoadOpenIncidents(
```
julianbrost, line 25: But in general, this function looks pretty similar to `GetCurrent()` and duplicates it quite a bit.

Shouldn't it be possible to simplify this function to something like this pseudocode:
```
for id in `select object_id from incident where recovered_at is null` {
  obj = LoadObjFromDB(id)
  incident = GetCurrentIncident(obj)
  incident.RestoreTimers()
}
```

Yes, that queries the `incident` table twice, but I'd rather keep the code simpler for now, as there'll be some quite intensive schema changes soon and we can optimize something like this after that was done.

internal/incident/incident.go
julianbrost: Please try to explain what these values are supposed to represent, without just explaining how they're set/accessed. So for example for the `Rules` member of the struct, that explanation would be something like "a set of all rules that are active on this incident and influence how notifications are sent" (instead of something like "rules are added when they match on this incident").

julianbrost (changes requested):

internal/object/db_types.go
```diff
@@ -27,3 +30,55 @@ type ObjectRow struct {
...
+	}
+
+	obj := &Object{db: db, ID: id, Tags: tags, Metadata: metadata}
+
+	cacheMu.Lock()
+	defer cacheMu.Unlock()
+
+	cache[id.String()] = obj
+
+	return obj, nil
```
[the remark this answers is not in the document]
yhabteab: We don't need such checks here, I don't even think that mutex lock is required, since this method is ever called when starting the daemon and before the listener becomes ready. And as discussed yesterday, an object can only have a single active incident at a time, i.e this method will never load the same object multiple times from the database.

internal/incident/incident.go
julianbrost: After some more looking at this, I think you're trying to store the `EscalationFilter` context for the time in the future when the escalation conditions are due to be evaluated the next time. However, and that's probably what has irritated me, you don't seem to consider that other external events can happen in that time. So if you have an escalation for `incident_age>=2h` and after 1h, the incident goes from warning to crit, that has to affect the context that's used for evaluating the condition, however you seem to always use the already stored one.

internal/incident/incidents.go
```diff
@@ -31,17 +94,9 @@ func GetCurrent(
...
-			runtimeConfig:    runtimeConfig,
-			configFile:       configFile,
-			Recipients:       map[recipient.Key]*RecipientState{},
-			EscalationState:  map[escalationID]*EscalationState{},
-			SeverityBySource: map[int64]event.Severity{},
-			Rules:            map[ruleID]struct{}{},
-		}
+		incident := NewIncident(db, runtimeConfig, configFile)
+		incident.Object = obj
+		incident.logger = logger.With(zap.String("object", obj.DisplayName()))
```
julianbrost, line 99: `object` and `logger` should be arguments of `NewIncident()` as well. After all, if they aren't set, that object won't function properly.

internal/incident/incident.go
```diff
@@ -575,6 +640,115 @@ func (i *Incident) RestoreEscalationStateRules(states []*EscalationState) {
...
+		return
+	}
+
+	ctx := context.TODO()
+	if i.RestoreRecipients(ctx) != nil {
+		// The error is already logged within `i.RestoreRecipients()`
+		return
+	}
+
+	i.evaluateEscalations(i.StartedAt.Add(i.timer.incidentAgeFilter))
```
julianbrost, line 663: Make that function return whether any new escalations were triggered. If there were none, you can return here and not send any notifications.

internal/incident/incident.go
```diff
@@ -575,6 +640,115 @@ func (i *Incident) RestoreEscalationStateRules(states []*EscalationState) {
...
+		i.logger.Errorw("Failed to start time based incident escalations database transaction", zap.Error(err))
+
+		return
+	}
+	defer func() { _ = tx.Rollback() }()
+
+	if i.notifyContacts(ctx, tx, &event.Event{Time: time.Now(), Type: event.TypeInternal}, types.Int{}) != nil {
+		// The error is already logged within `i.notifyContacts()`
+		return
+	}
```
julianbrost, line 676: Change the event type to `escalation`. Then it also somewhat makes sense to send it to all contacts: for the new ones as a "hey, this incident was escalated to you" and for the old ones "hey, this incident was escalated to more people".

julianbrost (changes requested):

julianbrost:
For completeness also as a comment here: I've pushed a [branch](https://github.com/Icinga/icinga-notifications/tree/jbrost/time-based-escalation), that for one, rebases it onto the current `main` branch and also contains some draft commits that simplify the timer logic (there may be a timer for a reevaluation in the future, any evaluation of the escalations (including from new external events) just always cancel the timer and schedule a new one as needed).

julianbrost:
I've just pushed an updated version with a bit cleaned up commits and rebased onto the current `main` to the aforementioned branch.

@yhabteab Can you please have a look if you can continue from here on again and make use of that branch for this PR?

Main open points should be at the moment:
- Who get's notified. I think you said that the functions you added in #99 will be helpful here to only contacts/channels that were newly escalated to.
- `evaluateEscalations()` needs a running database transaction now, but for the reevaluation, in case nothing new matches, we shouldn't need a transaction, because in that case, the transaction should be ended without executing a single query. So maybe we want to split that function, see also the inline TODO comment on my branch.

yhabteab:
> evaluateEscalations() needs a running database transaction now, but for the reevaluation, in case nothing new matches, we shouldn't need a transaction, because in that case, the transaction should be ended without executing a single query. 

Would it be that expensive if we just keep it the way it is? We would have to sync the dummy event beforehand anyway, which seems not to be the case at the moment (in your branch). At worst, that tx would perform a single insert query and rollback it afterwards.

julianbrost:
> Would it be that expensive if we just keep it the way it is?

I mean it's not prohibitively expensive, but it's also obvious room for improvement when we have all the information we need before starting the transaction.

> We would have to sync the dummy event beforehand anyway,

Do we have to? My idea idea would be to be to split `evaluateEscalations()` into two functions: one that doesn't need the database and simply evaluates the filters and returns which evaluations are to be triggered and another one to actually trigger escalations (without considering conditions, force trigger so to say). A bit similar to the recent rework in #99.

> which seems not to be the case at the moment (in your branch). 

Quite possible, I only looked at `type = 'notified'` events, those have the correct `event_id` set, others seem to have it missing.

yhabteab:
I have reset this PR to your branch and based on that I did also implemented the missing parts you mentioned. From my tests everything seems to work, I just don't like what https://github.com/Icinga/icinga-notifications/pull/68/commits/60ba92edd6bb4305353ea0c32ee3e93ce7c761f5 is doing at the moment. These two commits have already cleaned it all up https://github.com/Icinga/icinga-notifications/pull/41/commits/57cd57981d196ac99dfde08b3260563ea2992eea, https://github.com/Icinga/icinga-notifications/pull/41/commits/76837ad789893b35f9a871e2b753df1bcbb78b3c, but you need to decide if you want to review  that PR or just merge this one and rework it once again in the other PR.

julianbrost:
I don't want #41 to become a dependency for this PR. So if those two commits would work here on their own, we can consider it, otherwise this PR should probably better stay as it is.

internal/incident/incident.go
```diff
@@ -320,61 +422,119 @@ func (i *Incident) evaluateEscalations(ctx context.Context, tx *sqlx.Tx, ev *eve
...
+
+						if !matched {
+							// Evaluate whether the escalation filter evaluates to true 1 nanosecond
+							// after the configured incident age condition.
+							return escalation.Condition.Eval(
+								&rule.EscalationFilter{IncidentAge: cond.IncidentAge + time.Nanosecond, IncidentSeverity: cond.IncidentSeverity},
+							)
+						}
+
+						return matched, nil
```
julianbrost, line 456: I know that I suggested evaluating the rule at incident_age±1ns to capture strange uses of the `=` and `>` operator there, but now it feels like we're just making the code here more complex when we should probably tell users, that they should just go with `incident_age>=x` and maybe some others like `incident_age<x`. But for sure a restricted set of operators, because for `incident_age>=1h` it's obvious what should happen. Do we really want to make sure that `incident_age=1h` and `incident_age>1h` behave the same, or just offer `>=` in the web module for `incident_age`?

internal/incident/incident.go
```diff
@@ -298,13 +387,26 @@ func (i *Incident) evaluateRules(ctx context.Context, tx *sqlx.Tx, eventID int64
...
 
+	// Escalations are reevaluated now, reset any existing timer, if there might be future time-based escalations,
+	// this function will start a new timer.
+	if i.timer != nil {
+		i.logger.Info("Stopping reevaluate timer due to escalation evaluation")
+		i.timer.Stop()
+		i.timer = nil
+	}
+
+	cond := &rule.EscalationFilter{IncidentAge: eventTime.Sub(i.StartedAt), IncidentSeverity: i.Severity}
```
julianbrost, line 405: `cond` makes no sense for me here as a variable name. Should probably be something like `filterVars` (values for the variables that can be used in the escalation condition expression) or `filterContext` (context the escalation condition is evaluated in).

(I'm not a fan of that type being named `EscalationFilter` as it's not really a filter, but I don't want to start renaming it in this PR.)

internal/filter/parser_test.go
```diff
@@ -62,101 +62,98 @@ func TestFilter(t *testing.T) {
...
 
 		rule, err = Parse("foo=bar|bar=foo")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.IsType(t, &Any{}, rule)
+		assert.IsType(t, &Chain{}, rule)
 
 		rule, err = Parse("!(foo=bar|bar=foo)")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-		assert.IsType(t, &None{}, rule)
+		assert.IsType(t, &Chain{}, rule)
```
julianbrost, line 113: These should now also check `rule.op`.

internal/filter/parser_test.go
```diff
@@ -62,101 +62,98 @@ func TestFilter(t *testing.T) {
...
 
 		rule, err = Parse("!foo")
 		assert.Nil(t, err, "There should be no errors but got: %s", err)
-
-		assert.Equal(t, &None{rules: []Filter{NewExists("foo")}}, rule)
+		assert.IsType(t, &Chain{}, rule)
```
julianbrost, line 117: Why no more `assert.Equal()` here?

internal/filter/parser_test.go
```diff
@@ -62,101 +62,98 @@ func TestFilter(t *testing.T) {
...
-			&None{rules: []Filter{
-				&Equal{column: "foo", value: "bar"},
-				&Equal{column: "bar", value: "foo"},
-			}},
-			&Any{rules: []Filter{
-				&Equal{column: "foo", value: "bar"},
-				&Equal{column: "bar", value: "foo"},
-			}},
-		}}
-		assert.Equal(t, expected, rule)
```
julianbrost, line 129: Why was this test removed altogether?

internal/filter/types.go
```diff
@@ -1,67 +1,106 @@
...
 
-	return true, nil
-}
+const (
+	// NONE represents a filter chain type that matches when none of its ruleset matches.
+	NONE LogicalOp = "!"
+	// ALL represents a filter chain type that matches when all of its ruleset matches.
+	ALL LogicalOp = "&"
+	// ANY represents a filter chain type that matches when at least one of its ruleset matches.
+	ANY LogicalOp = "|"
```
julianbrost, line 16: Why are these ALL CAPS in contrast to the `CompOperator` constants?

internal/filter/types.go
julianbrost, line 1: `NewCondition` seems unused now and also seems incomplete now and should therefore better be removed (or updated and used everywhere a `Condition` is created).

internal/filter/types.go
```diff
@@ -73,116 +112,108 @@ func NewCondition(column string, value string) *Condition {
...
+	case UnEqual:
+		match, err = filterable.EvalEqual(c.column, c.value)
+		if err != nil {
+			return false, err
+		}
 
-func NewExists(column string) *Exists {
-	return &Exists{column: column}
-}
+		match = filterable.EvalExists(c.column) && !match
```
julianbrost, line 132: I find it surprising that `EvalEqual` is done before `EvalExists`. But at the very least, it should be `!match && EvalExists()`, no need to run that function if it can't influence the result anyways.

julianbrost (changes requested):

internal/filter/parser_test.go
yhabteab: Just ignore theses! There's an extended test case in the other PR.

internal/rule/condition.go
```diff
@@ -3,82 +3,111 @@ package rule
...
 }
 
-func (c *EscalationFilter) EvalEqual(key string, value string) (bool, error) {
+// ExtractRetryAfter extracts a time.Duration from the given filter conditions.
+// The retry after duration is extracted from the specified filter conditions that evaluate the "incident_age"
+// filter column. This function returns rule.RetryNever when all the incident age filter values are less than
+// the actual incident age or none of the filter conditions evaluates the "incident_age" column. Otherwise,
+// a time.Duration < rule.RetryNever and > the current incident age at which this escalation condition should
+// be re-evaluated is returned.
+func (e *EscalationFilter) ExtractRetryAfter(conditions []filter.Conditioner, incidentAge time.Duration) time.Duration {
```
julianbrost, line 25: This should either not be implemented on `*EscalationFilter` or not have the `incidentAge` parameter.

julianbrost (changes requested):

internal/filter/contracts.go
```diff
@@ -12,4 +12,12 @@ type Filterable interface {
...
 	Eval(filterable Filterable) (bool, error)
+	ExtractConditions() []Conditioner
+}
+
+// Conditioner provides a way to access a filter condition column and value.
+// It is implemented by the Condition and Exists filter types.
+type Conditioner interface {
+	Column() string
+	Value() string
 }
```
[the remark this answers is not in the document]
yhabteab: > That interface feels like it's not necessary at all. After all, it's only used by `ExtractConditions()` which in turn is used to get all values `incident_age` is compared with.

How do you think the fields from the `EscalationFilter#ExtractRetryAfter()` method should be accessed without making them exportable then? Yes, of course the filter `Exists` does not have to implement this interface, but I don't see any disadvantages in doing so either.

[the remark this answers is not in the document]
julianbrost: > How do you think the fields from the `EscalationFilter#ExtractRetryAfter()` method should be accessed without making them exportable then?

Would there be anything from with just exporting them?

> Yes, of course the filter `Exists` does not have to implement this interface, but I don't see any disadvantages in doing so either.

There is no need for it and the the current implementation claims that the column/variable `incident_age` is compared with the value `incident_age` which just isn't the case.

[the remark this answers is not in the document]
yhabteab: Just removed the interface but still didn't make the fields exportable.

julianbrost (changes requested):

internal/rule/condition.go
```diff
@@ -3,82 +3,112 @@ package rule
...
 
-func (c *EscalationFilter) EvalEqual(key string, value string) (bool, error) {
+// ExtractRetryAfter extracts a time.Duration from the given filter conditions relative to the actual incident age.
+//
+// The retry after duration is extracted from the specified escalation condition that evaluate the "incident_age"
+// filter column. This function returns rule.RetryNever when all the incident age filter values are less than
+// the actual incident age or none of the filter conditions evaluates the "incident_age" column. Otherwise,
+// a time.Duration < rule.RetryNever and > the current incident age at which this escalation condition should
+// be re-evaluated is returned.
+func (e *EscalationFilter) ExtractRetryAfter(escalationCond filter.Filter) time.Duration {
```
julianbrost, line 26: My suggestion for the overall function documentation (including a renaming of the function itself) would be the following:

```suggestion
// ReevaluateAfter returns the duration after which escalationCond should be reevaluated the 
// next time on the incident represented by e.
//
// escalationCond must correspond to an escalation that did not trigger on the incident
// represented by e before. If nothing in the incident changes apart from time passing by,
// the escalation is guaranteed to not trigger within the returned duration. After that
// duration, the escalation should be reevaluated and it may or may not trigger. If anything
// else changes, for example due to an external event, the escalation must be reevaluated as
// well.
func (e *EscalationFilter) ReevaluateAfter(escalationCond filter.Filter) time.Duration {
```

Parts of the current comment could be moved to inline comments within the implementation to explain it if you deem that necessary. But my impression is that it just rephrases the code in human language without giving much extra explanation.

internal/rule/condition.go
```diff
@@ -3,82 +3,114 @@ package rule
...
+	for _, condition := range escalationCond.ExtractConditions() {
+		if condition.Column() == "incident_age" {
+			incidentAgeFilterValue, err := time.ParseDuration(condition.Value())
+			if err != nil {
+				continue
+			}
+
+			if incidentAgeFilterValue > e.IncidentAge {
+				retryAfter = min(retryAfter, incidentAgeFilterValue)
+			}
```
julianbrost, line 39: D'oh, in my head, we already changed the return value to be relative to `e` (~now) instead of the incident start. Thus, I've written the comment with that in mind, but now it's actually inconsistent with the implementation.

Anyways, I'd change the implementation to match the comment, because I think this makes sense that way around. It can also be written a bit more compact without becoming harder to understand IMHO, maybe even a little improvement in that regard with no more `continue`.

```suggestion
			v, err := time.ParseDuration(condition.Value())
			if err == nil && v > e.IncidentAge {
				// The incident age is compared with a value in the future. Once that age is
				// reached, the escalation could trigger, so consider that time for reevaluation.
				retryAfter = min(retryAfter, v - e.IncidentAge)
			}
```

Of course, also requires a change where it's called to use the return value in the appropriate way.

julianbrost (changes requested):

internal/rule/condition.go
yhabteab: I somehow can't quite grasp why you keep bringing up how complex this is. It's only 2-3 lines of code that you don't need any comments to be able to understand it. But hey, that shouldn't matter, even if I didn't understand your comments from yesterday enough to understand what the function does compared to the previous ones. The comments just describe what might or might not happen when using this function somewhere in incident package.

julianbrost: The main change request here is to do this, i.e. change the reference point for the return value to match the updated documentation:
```suggestion
			incidentAgeFilterValue, err := time.ParseDuration(condition.Value())
			if err != nil {
				continue
			}

			if incidentAgeFilterValue > e.IncidentAge {
				retryAfter = min(retryAfter, incidentAgeFilterValue - e.IncidentAge)
			}
```

But then I thought this can be simplified a bit and proposed a larger change.

Regarding the function documentation and inline comments: I think that those should give context to the code, they should say why and what for things are done. The function we're talking here has a very specific purpose: if I have an incident and an escalation that hasn't triggered yet, tell me for how long is it safe to ignore that escalation, because the result of evaluating the condition won't change until then. From the view of someone calling the function, that should be the relevant information, not what values are compared with the incident age in which way. I was trying to push you incrementally towards a direction where I think that is easier to understand from the interface and documentation of the function. I thought just requesting a big "just do it completely differently" change would have been frustrating for you, but my approach is already taking more iterations and time than I had expected, so it wasn't very successful either.

internal/incident/incident.go
```diff
@@ -320,61 +426,89 @@ func (i *Incident) evaluateEscalations(ctx context.Context, tx *sqlx.Tx, ev *eve
...
+		i.logger.Infow("Scheduling escalation reevaluation", zap.Duration("after", retryAfter), zap.Time("at", nextEvalAt))
+		i.timer = time.AfterFunc(retryAfter, func() {
+			i.logger.Info("Reevaluating escalations")
 
-						return errors.New("failed to upsert escalation state")
-					}
+			i.RetriggerEscalations(&event.Event{
+				Type:    event.TypeInternal,
+				Time:    nextEvalAt,
+				Message: fmt.Sprintf("Incident reached age %v", retryAfter),
```
julianbrost, line 465: This message is now incorrect with the changed value: If you have escalations `incident_age>=1m` and `incident_age>=2m`, the second escalation will also use "age 1m" because that's the difference to the last evaluation.

julianbrost (changes requested):
One message in the event is incorrect. Apart from that, this PR should be good to go.

julianbrost (approved):

PR #76, opened 2023-05-17
Use `sqlx.Tx` for event processing

yhabteab:
To keep the database consistent.


internal/object/object.go
```diff
@@ -117,46 +126,30 @@ func (o *Object) UpdateMetadata(source int64, name string, url types.String, ext
...
 	}
 
-	if m := o.Metadata[source]; m != nil {
-		m.Name = name
-		m.URL = url
-		m.ExtraTags = extraTags
-	} else {
-		o.Metadata[source] = sourceMetadata
-	}
+	o.Metadata[source] = sourceMetadata
```
julianbrost, line 152: Note to my self: double-check that there are no assumptions on constant addresses in the code. (This clearly looks like code I wrote to keep the pointer in `o.Metadata[source]` constant, at least for the config, there were some assumptions on that.)

internal/listener/listener.go
```diff
@@ -86,16 +85,18 @@ func (l *Listener) ProcessEvent(w http.ResponseWriter, req *http.Request) {
 		}
 	}
 
-	obj, err := object.FromTags(l.db, ev.Tags)
+	ctx := req.Context()
```
julianbrost, line 88: Please explain the idea behind using `req.Context()` here.

I'm not sure if we want this in the end, but that will probably depend on how we handle HA in the end.

Typically, that context is used to stop generating a response soon after a connection has died because this means that the response won't be received anyways. But in our use-case, the main effect of the request is tracking the incident state and sending notifications, so if we've received the request, it can make sense to continue processing it even after the connection that sent the event died.

julianbrost (changes requested):

julianbrost:
I have just created #81, you might want to have a look at this. I'd keep this PR in the current state with only one transaction to keep it smaller. Splitting the transaction and marking the notifications as pending will require some extra work including schema changes so I feel like that can be done separately. Though, if you have the feeling for future changes to this PR, you'd have to do things that will be changed with #81 again, we can also consider doing this in here as well.

internal/incident/incident.go
```diff
@@ -73,51 +75,70 @@ func (i *Incident) HasManager() bool {
...
+		eventRow := event.NewEventRow(ev, i.Object.ID)
+		eventID, err := utils.InsertAndFetchId(ctx, tx, utils.BuildInsertStmtWithout(i.db, eventRow, "id"), eventRow)
+		if err != nil {
+			i.logger.Errorw("Failed to insert event and fetch its ID", zap.Error(err))
+
+			return errors.New("can't insert event and fetch its ID")
+		}
+
+		ev.ID = eventID
+	}
```
julianbrost, line 102: Is there any benefit in doing this here compared to where it was done before (i.e. in listener)? At least `UpdateMetadata()` hasn't much to do with the current incident, so I don't think this makes much sense in a receiver function on the incident type.

Similarly for writing the event to the database. At the moment it's tied to an incident, but in the future, also events like downtime starts will be handled, and these can occur without an active incident and should also be written to the database, so I wouldn't move this here either.

julianbrost (changes requested):

PR #77, opened 2023-05-25
Code quality

yhabteab:
Just some code enhancements (better logging, better request validation).


internal/listener/listener.go
```diff
@@ -158,14 +163,18 @@ func (l *Listener) ProcessEvent(w http.ResponseWriter, req *http.Request) {
 func (l *Listener) checkDebugPassword(w http.ResponseWriter, r *http.Request) bool {
 	expectedPassword := l.configFile.DebugPassword
 	if expectedPassword == "" {
+		l.logger.Warnw("Config dump disables, no debug-password set in config", zap.String("url", r.RequestURI), zap.String("agent", r.UserAgent()))
+
```
julianbrost, line 167: I wouldn't log this here as there's little point seeing this in the logs. After all, these endpoints are intended for manual consumption, so you'll see that message (not like in some source doing whatever it does with the response body).

internal/listener/listener.go
```diff
@@ -158,14 +163,18 @@ func (l *Listener) ProcessEvent(w http.ResponseWriter, req *http.Request) {
...
 		w.WriteHeader(http.StatusForbidden)
 		_, _ = fmt.Fprintln(w, "config dump disables, no debug-password set in config")
 
 		return false
 	}
 
-	_, providedPassword, _ := r.BasicAuth()
+	username, providedPassword, _ := r.BasicAuth()
 	if subtle.ConstantTimeCompare([]byte(expectedPassword), []byte(providedPassword)) != 1 {
+		l.logger.Warnw("Unauthorized request", zap.String("url", r.RequestURI), zap.String("user", username), zap.String("agent", r.UserAgent()))
```
julianbrost, line 176: I don't see a value in logging user name and user agent here. These are just user provided values that the rest of the code doesn't care about at all.

internal/listener/listener.go
```diff
@@ -176,6 +185,12 @@ func (l *Listener) checkDebugPassword(w http.ResponseWriter, r *http.Request) bo
 }
 
 func (l *Listener) DumpConfig(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodGet {
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		_, _ = fmt.Fprintln(w, "Bad request method")
+		return
+	}
```
julianbrost, line 192: I'd say something like "GET required" as it not only says you're doing something wrong but also what you should do instead.

internal/incident/incident.go
```diff
@@ -461,21 +503,35 @@ func (i *Incident) notifyContacts(ev *event.Event, causedBy types.Int) error {
...
+					"failed to send via channel", zap.String("type", chType), zap.String("object", i.ObjectDisplayName()),
+					zap.String("incident", i.String()), zap.Error(err),
+				)
 				continue
 			}
+
+			i.logger.Infow(
+				"Successfully sent a message vai channel", zap.String("contact", contact.String()), zap.String("type", chType),
+				zap.String("object", i.ObjectDisplayName()), zap.String("incident", i.String()),
+			)
```
julianbrost, line 534: Typo in "via".

julianbrost (changes requested):
This PR seems to introduce quite some inconsistencies in the capitalization of log messages, please try to stay consistent there.

internal/incident/incident.go
```diff
@@ -276,7 +265,7 @@ func (i *Incident) evaluateRules(eventID int64, causedBy types.Int) (types.Int,
 			}
 
 			i.Rules[r.ID] = struct{}{}
-			i.logger.Infof("[%s %s] rule %q matches", i.Object.DisplayName(), i.String(), r.Name)
+			i.logger.Infof("Rule %q matches on", r.Name)
```
julianbrost, line 268: I wouldn't add the "on" here at the end. I think this is supposed to be a reference to the JSON that's appended to the log message, but I think the log message should be self-contained with the JSON providing additional context.

internal/incident/incidents.go
```diff
@@ -42,10 +42,13 @@ func GetCurrent(
...
 		err := db.QueryRowx(db.Rebind(db.BuildSelectStmt(ir, ir)+` WHERE "object_id" = ? AND "recovered_at" IS NULL`), obj.ID).StructScan(ir)
 		if err != nil && err != sql.ErrNoRows {
-			return nil, false, fmt.Errorf("incident query failed with: %w", err)
+			logger.Errorw("Failed to load incident from database", zap.String("object", obj.DisplayName()), zap.Error(err))
+
+			return nil, false, errors.New("failed to load incident from database")
 		} else if err == nil {
 			incident.incidentRowID = ir.ID
 			incident.StartedAt = ir.StartedAt.Time()
+			incident.logger = logger.With(zap.String("object", obj.DisplayName()), zap.String("incident", incident.String()))
```
julianbrost, line 51: You're initializing `incident.logger` only here but not when creating a new incident. In that case, it will panic.

julianbrost (changes requested):

internal/incident/incident.go
```diff
@@ -86,13 +85,12 @@ func (i *Incident) ProcessEvent(ev event.Event, created bool) error {
 		if err != nil {
 			return err
 		}
+
+		i.logger = i.logger.With(zap.String("object", i.ObjectDisplayName()), zap.String("incident", i.String()))
```
julianbrost, line 89: You shouldn't use `.With()` to update existing fields, they just get accumulated (note the duplicate `object`):

```
2023-05-26T11:27:00.097+0200    INFO    incident        Incident severity changed from ok to warning    {"object": "dc1.example.com: memory", "object": "dc1.example.com: memory", "incident": "#3809"}
```

julianbrost (changes requested):

internal/incident/sync.go
```diff
@@ -46,7 +46,7 @@ func (i *Incident) AddHistory(historyRow *HistoryRow, fetchId bool) (types.Int,
 	} else {
 		_, err := i.db.NamedExec(stmt, historyRow)
 		if err != nil {
-			return types.Int{}, fmt.Errorf("failed to insert incident history: %w", err)
+			return types.Int{}, err
```
julianbrost, line 49: If you remove that context information here, please make sure that all callers add sufficient context.

For example, the following code will just log the error without anything saying that the error was related to writing the history:
https://github.com/Icinga/icinga-notifications/blob/f5fb90f73b112104b1a3c975ae8ba414d9b0004e/internal/incident/incident.go#L454-L457

(There may be more callers that need adjustment, I haven't looked through all.)

julianbrost (changes requested):

internal/incident/incident.go
```diff
@@ -385,18 +376,24 @@ func (i *Incident) notifyContacts(ev *event.Event, causedBy types.Int) error {
...
 			if err != nil {
-				i.logger.Errorln(err)
+				i.logger.Errorw(
+					"Failed to add escalation triggered history", zap.String("rule", r.Name),
+					zap.String("escalation", escalation.DisplayName()), zap.Error(err),
+				)
 
-				return err
+				return errors.New("failed to add escalation triggered history")
 			}
```
julianbrost, line 385: `AddEscalationTriggered()` actually writes to both `incident_rule_escalation_state` and `incident_history` so that log message can be misleading. Probably better move the logging inside that function with two different messages.

internal/incident/incident.go
```diff
@@ -287,9 +278,9 @@ func (i *Incident) evaluateRules(eventID int64, causedBy types.Int) (types.Int,
...
 			insertedID, err := i.AddRuleMatchedHistory(r, history)
 			if err != nil {
-				i.logger.Errorln(err)
+				i.logger.Errorw("Failed to add incident rule matched history", zap.String("rule", r.Name), zap.Error(err))
 
-				return types.Int{}, err
+				return types.Int{}, errors.New("failed to add incident rule matched history")
 			}
```
julianbrost, line 284: Same here, that function actually does more than just write history (even despite its name).

internal/incident/sync.go
julianbrost: Also affects this code (the error will be logged by the caller with a pretty generic message):
https://github.com/Icinga/icinga-notifications/blob/c4b6db47f6c598bd964f1fff0863dda9bd7faf28/internal/incident/sync.go#L112-L115

julianbrost (changes requested):

internal/incident/incident.go
```diff
@@ -285,11 +276,18 @@ func (i *Incident) evaluateRules(eventID int64, causedBy types.Int) (types.Int,
...
+			}
+
+			insertedID, err := i.AddHistory(history, true)
 			if err != nil {
-				i.logger.Errorln(err)
+				i.logger.Errorw("Failed to inserted rule matched incident history", zap.String("rule", r.Name), zap.Error(err))
 
-				return types.Int{}, err
+				return types.Int{}, errors.New("failed to insert rule matched incident history")
 			}
```
julianbrost, line 291: You mixed up some error messages here.

internal/incident/incident.go
```diff
@@ -285,11 +276,18 @@ func (i *Incident) evaluateRules(eventID int64, causedBy types.Int) (types.Int,
...
-			insertedID, err := i.AddRuleMatchedHistory(r, history)
+			err := i.AddRuleMatched(r)
+			if err != nil {
+				i.logger.Errorw("Failed to upsert incident rule", zap.String("rule", r.Name), zap.Error(err))
+
+				return types.Int{}, errors.New("failed to insert incident rule matched history")
+			}
+
+			insertedID, err := i.AddHistory(history, true)
```
julianbrost, line 286: Also, as that was now moved out of the function, better move the initialization of the `HistoryRow` right above the call to `AddHistory()`.

julianbrost (changes requested):

PR #79, opened 2023-05-26
Init incident rules on daemon startup

### LINKED ISSUE #78: Duplicate rule matched history entries after daemon restart
To reproduce:

1. Bring an object into a problem state for which a rule matches.
2. Stop and start the daemon again.
3. Bring the same object into another problem state so that a new event is sent.
4. Observe how there now are two "Rule [rule name] matched on this incident" entries displayed in the incident history.

I haven't looked into this in more detail, but if I had to guess, I suspect some state isn't restored from the database.

yhabteab:
fixes #78

### DIFF
--- a/internal/incident/incidents.go
+++ b/internal/incident/incidents.go
@@ -39,6 +39,7 @@ func GetCurrent(
 			Recipients:       map[recipient.Key]*RecipientState{},
 			EscalationState:  map[escalationID]*EscalationState{},
 			SeverityBySource: map[int64]event.Severity{},
+			Rules:            map[ruleID]struct{}{},
 		}
 
 		err := db.QueryRowx(db.Rebind(db.BuildSelectStmt(ir, ir)+` WHERE "object_id" = ? AND "recovered_at" IS NULL`), obj.ID).StructScan(ir)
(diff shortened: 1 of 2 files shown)


internal/incident/incidents.go
```diff
@@ -77,8 +77,14 @@ func GetCurrent(
...
 
+			incident.runtimeConfig.RLock()
+			defer incident.runtimeConfig.RUnlock()
+
 			for _, state := range states {
 				incident.EscalationState[state.RuleEscalationID] = state
+
+				escalation := incident.runtimeConfig.GetRuleEscalation(state.RuleEscalationID)
+				incident.Rules[escalation.RuleID] = struct{}{}
 			}
```
julianbrost, line 88: `defer` is executed only when leaving the function. Until that happens, this function does quite a few things that don't need this lock including database queries.

Now it's a bit annoying that `defer` only works on function scopes, so maybe put this block of code into it's own function, something like `func (i *Incident) RestoreState(states []*EscalationState)`.

julianbrost (changes requested):

julianbrost (changes requested):

julianbrost (approved):

PR #87, opened 2023-09-14
Add `membership_hash` column to `schedule_member` table

### LINKED ISSUE #83: `membership_hash` column required for `schedule_member` table
To efficiently update timeperiod entries, we need to assign a membership hash based on recipients for the given schedule. 

Currently a new timeperiod has been created each time a new timeperiod entry has been added instead of using the existing timeperiod for the given schedule and the combination of users and usergroups. Hence, adding the `membership_hash` to the `schedule_member` solves this issue.

raviks789:
fixes #83

### DIFF
--- a/schema/pgsql/schema.sql
+++ b/schema/pgsql/schema.sql
@@ -89,6 +89,7 @@ CREATE TABLE schedule_member (
     timeperiod_id bigint NOT NULL REFERENCES timeperiod(id),
     contact_id bigint REFERENCES contact(id),
     contactgroup_id bigint REFERENCES contactgroup(id),
+    membership_hash bytea DEFAULT NULL,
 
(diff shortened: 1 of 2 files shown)


schema/pgsql/schema.sql
```diff
@@ -89,6 +89,7 @@ CREATE TABLE schedule_member (
     timeperiod_id bigint NOT NULL REFERENCES timeperiod(id),
     contact_id bigint REFERENCES contact(id),
     contactgroup_id bigint REFERENCES contactgroup(id),
+    membership_hash bytea DEFAULT NULL
```
julianbrost, line 92: Missing `,` at the end of the line.

julianbrost (changes requested):

julianbrost (approved):

PR #98, opened 2023-09-21
Directly link objects to their sources

### LINKED ISSUE #84: Directly link objects with sources
Currently, objects have no direct relations with a source but are only indirectly linked to sources via events. That means that multiple sources can submit events for the same object at the moment, where the idea was to allow the same incident to be fed with events from multiple sources. However, that's quite a strong assumption and we'll want incidents covering multiple objects anyways, so it makes sense just doing the same for the "same" object from multiple sources (these will be considered different objects after this change, that's why same is in quotes).

### Tasks

Every step has an implicit "adapt the Go code accordingly" attached to it:

* Move the `source_id` foreign key from the `event` table to the `object` table (TBD: include in primary key or in `object.id` hash?).
* This makes the `source_object` table obsolete, move the information from it to the `object` table.
* After review for schema and Go changes: coordinate corresponding changes in PHP.

yhabteab:
resolves #84

### DIFF
--- a/internal/incident/history_event_type.go
+++ b/internal/incident/history_event_type.go
@@ -9,7 +9,6 @@ type HistoryEventType int
 
 const (
 	HistoryEventTypeNull HistoryEventType = iota
-	SourceSeverityChanged
 	SeverityChanged
 	RecipientRoleChanged
 	EscalationTriggered
@@ -20,7 +19,6 @@ const (
 )
 
 var historyTypeByName = map[string]HistoryEventType{
-	"source_severity_changed":   SourceSeverityChanged,
 	"incident_severity_changed": SeverityChanged,
 	"recipient_role_changed":    RecipientRoleChanged,
 	"escalation_triggered":      EscalationTriggered,
--- a/internal/incident/incident.go
+++ b/internal/incident/incident.go
@@ -23,10 +23,10 @@ type ruleID = int64
 type escalationID = int64
 
 type Incident struct {
-	Object           *object.Object
-	StartedAt        time.Time
-	RecoveredAt      time.Time
-	SeverityBySource map[int64]event.Severity
+	Object      *object.Object
+	StartedAt   time.Time
+	RecoveredAt time.Time
+	Severity    event.Severity
 
 	EscalationState map[escalationID]*EscalationState
 	Rules           map[ruleID]struct{}
@@ -52,16 +52,6 @@ func (i *Incident) String() string {
 
 func (i *Incident) ID() int64 {
 	return i.incidentRowID
-}
-
-func (i *Incident) Severity() event.Severity {
-	maxSeverity := event.SeverityOK
-	for _, s := range i.SeverityBySource {
-		if s > maxSeverity {
-			maxSeverity = s
-		}
-	}
-	return maxSeverity
 }
 
 func (i *Incident) HasManager() bool {
@@ -101,9 +91,13 @@ func (i *Incident) ProcessEvent(ctx context.Context, tx *sqlx.Tx, ev event.Event
 		return i.processAcknowledgementEvent(ctx, tx, ev)
 	}
 
-	causedBy, err := i.processSeverityChangedEvent(ctx, tx, ev)
-	if err != nil {
-		return err
+	var err error
+	var causedBy types.Int
+	if !created {
+		causedBy, err = i.processSeverityChangedEvent(ctx, tx, ev)
+		if err != nil {
+			return err
+		}
 	}
 
 	// Check if any (additional) rules match this object. Filters of rules that already have a state don't have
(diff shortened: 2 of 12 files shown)


internal/object/object.go
```diff
@@ -230,26 +217,29 @@ func (o *Object) EvalExists(key string) bool {
...
+	if ok {
+		return true
 	}
 
 	return false
 }
 
 // TODO: the return value of this function must be stable like forever
-func ID(tags map[string]string) types.Binary {
+func (o *Object) ID() types.Binary {
```
julianbrost, line 229: Why did you make this a method on `*Object`? The ID is supposed to be a deterministic value computed from a limited set of inputs, so I think it's a good idea to keep those as explicit parameters.

internal/object/object.go
```diff
@@ -230,26 +217,29 @@ func (o *Object) EvalExists(key string) bool {
...
+	if o.id.Valid() {
+		return o.id
+	}
+
 	h := sha256.New()
+	h.Write([]byte{byte(o.SourceId)})
```
julianbrost, line 235: This won't work properly for `o.SourceID >= 256`.

internal/incident/incident.go
```diff
@@ -119,74 +109,47 @@ func (i *Incident) ProcessEvent(ctx context.Context, tx *sqlx.Tx, ev event.Event
...
+func (i *Incident) processSeverityChangedEvent(ctx context.Context, tx *sqlx.Tx, ev event.Event, created bool) (types.Int, error) {
+	oldIncidentSeverity := i.Severity
+	if oldIncidentSeverity == event.SeverityNone {
+		oldIncidentSeverity = event.SeverityOK
 	}
 
-	if oldSourceSeverity == ev.Severity {
+	var causedByHistoryId types.Int
+	newIncidentSeverity := ev.Severity
+	if oldIncidentSeverity == newIncidentSeverity && !created {
```
julianbrost, line 120: What is the `&& !created` you added for? If the incident is created, the old severity should be ok and the new severity something else, so that should already be captured by the other part of the condition.

internal/incident/incident.go
```diff
@@ -119,74 +109,47 @@ func (i *Incident) ProcessEvent(ctx context.Context, tx *sqlx.Tx, ev event.Event
...
-	if ev.Severity == event.SeverityOK {
-		delete(i.SeverityBySource, ev.SourceId)
-	}
-
-	newIncidentSeverity := i.Severity()
-	if newIncidentSeverity != oldIncidentSeverity {
-		i.logger.Infof(
-			"Incident severity changed from %s to %s", oldIncidentSeverity.String(), newIncidentSeverity.String(),
-		)
+	if oldIncidentSeverity != newIncidentSeverity {
```
julianbrost, line 127: Then this should also no longer need this if. So if we get here, shouldn't there always be a change?

internal/object/object.go
```diff
@@ -17,39 +19,60 @@ import (
...
+	objectCache, ok := cache[id.String()]
 	if ok {
-		return object, nil
+		objectCache.ExtraTags = object.ExtraTags
+
+		return objectCache, nil
 	}
 
-	object = &Object{ID: id, Tags: tags, db: db}
 	cache[id.String()] = object
```
julianbrost, line 67: Please try to keep the structure so that the cache lookup happens first and depending on whether it's in there, use it or create it. I don't see why you'd want to do `NewObject()` and then dismiss it later (and having that just makes it easier to introduce bugs where the wrong one is returned or something like that).

schema/pgsql/upgrades/015.sql
```diff
@@ -0,0 +1,18 @@
+ALTER TABLE object
+  ADD COLUMN source_id bigint NOT NULL REFERENCES source(id) DEFAULT 1,
+  ADD COLUMN name text NOT NULL DEFAULT '',
```
julianbrost, line 3: The schema migration should not leave defaults for columns set that aren't set by the full schema file. Also, if in doubt - at this stage of the project - it's better to just truncate the tables instead of leaving rows with some fake values that don't work properly (i.e. in order to migrate the existing data for this change, you'd have to recompute all IDs and change them throughout the database, otherwise, this just leaves half-broken data behind).

schema/pgsql/schema.sql
```diff
@@ -123,51 +123,43 @@ CREATE TABLE source (
...
     object_id bytea NOT NULL REFERENCES object(id),
-    source_id bigint NOT NULL REFERENCES source(id),
-
     tag text NOT NULL,
     value text NOT NULL,
 
-    CONSTRAINT pk_object_extra_tag PRIMARY KEY (object_id, source_id, tag),
-    FOREIGN KEY (object_id, source_id) REFERENCES source_object(object_id, source_id)
+    CONSTRAINT pk_object_extra_tag PRIMARY KEY (object_id, tag),
+    FOREIGN KEY (object_id) REFERENCES object(id)
```
julianbrost, line 147: That `FOREIGN KEY` definition probably is redundant (should already be created by the `REFERENCES object(id)` in the column defintion).

schema/pgsql/schema.sql
```diff
@@ -123,51 +123,43 @@ CREATE TABLE source (
...
-    source_id bigint NOT NULL REFERENCES source(id),
     object_id bytea NOT NULL REFERENCES object(id),
     type text NOT NULL,
     severity severity,
     message text,
     username text,
 
     CONSTRAINT pk_event PRIMARY KEY (id),
-    FOREIGN KEY (object_id, source_id) REFERENCES source_object(object_id, source_id)
+    FOREIGN KEY (object_id) REFERENCES object(id)
```
julianbrost, line 162: Same redundant `FOREIGN KEY` definition here.

julianbrost (changes requested):

internal/incident/incident.go
yhabteab: If the incident has just been created, it will first enter `Incident#processIncidentOpenedEvent()` and set the `severity` to the event severity.
https://github.com/Icinga/icinga-notifications/blob/288093e269564567e3d50966310adc02be107591/internal/incident/incident.go#L186-L188
Afterwards, when it goes in here (`Incident#processSeverityChangedEvent()`), old and new severities are the same, of course.

internal/incident/incident.go
yhabteab: See the above comment!

internal/incident/incident.go
julianbrost: I see, looks like you're changing more than I expected: Previously, incidents were always created in an ok state, so the first event opening the incident wasn't treated differently from later events in this function. You're changing this to directly creating the incident at the right severity. Nothing wrong with that, but this should probably go with some other code changes then, for example, is there any case left where `processSeverityChangedEvent()` would actually do anything if called with `create=true`? So if you change it so that there is no longer an explicit severity change when opening the incident, this function probably shouldn't even be called when creating an incident.

You can probably also see this in the generated history, with the current main branch, it looks something like that:

1. Opened this incident
2. Source Icinga 2 reported a severity change from Ok to Warning: [...output/message...]
3. Incident severity changed from Ok to Warning

What I'd have expected from this PR would be a history like:

1. Opened this incident
2. Severity changed from Ok to Warning: [...output/message...]

But with the changes you started, it should probably look something like this:

1. Opened incident at severity Warning: [...output/message...]

Which probably is more on point, but as I said, required slightly more extensive changes, which might also explain why you were confused about the cause column in the history table.

internal/incident/history_event_type.go
```diff
@@ -9,7 +9,6 @@ type HistoryEventType int
 
 const (
 	HistoryEventTypeNull HistoryEventType = iota
-	SourceSeverityChanged
```
julianbrost, line 12: The now obsolete value should also be removed from the enum in the schema.

internal/incident/incident.go
```diff
@@ -217,18 +163,27 @@ func (i *Incident) processSeverityChangedEvent(ctx context.Context, tx *sqlx.Tx,
...
 
 		return errors.New("can't insert incident to the database")
 	}
 
-	i.logger.Info("Opened incident")
+	i.logger.Infow(fmt.Sprintf("Source %d opened incident at severity %q", ev.SourceId, i.Severity.String()), zap.String("message", ev.Message))
+
 	historyRow := &HistoryRow{
 		Type:    Opened,
 		Time:    types.UnixMilli(ev.Time),
```
julianbrost, line 189: If the incident is now opened with the non-ok severity directly (which arguably makes more sense), the information that was previously contained in other history events should now be available in that history event (that is at least the new/initial severity and the message).

internal/incident/incident.go
```diff
@@ -120,90 +114,42 @@ func (i *Incident) ProcessEvent(ctx context.Context, tx *sqlx.Tx, ev event.Event
...
-	)
+	i.logger.Infof("Incident severity changed from %s to %s", oldSeverity.String(), newSeverity.String())
 
 	history := &HistoryRow{
 		EventID:     utils.ToDBInt(ev.ID),
-		Type:        SourceSeverityChanged,
 		Time:        types.UnixMilli(time.Now()),
-		NewSeverity: ev.Severity,
-		OldSeverity: oldSourceSeverity,
-		Message:     utils.ToDBString(ev.Message),
```
julianbrost, line 146: Similarly, that event should still contain the message.

julianbrost (changes requested):

julianbrost (changes requested):
One more detail, but then this should be fine.

PR #99, opened 2023-09-25
Send notifications outside a database transaction

### LINKED ISSUE #81: Send notifications outside of database transaction
#76 will put the event processing inside a database transaction which is a good thing, but comes with one problem: while sending the notifications, that transaction isn't committed yet, so if for some reason, the transaction fails to commit, we'd have sent a notification for which there is no trace in the database.

Thus, that transaction should be split into two: first, one that updates the incidents and adds some kind of pending notification. This one should be committed, only then should the notification be sent and finally the notification is marked as completed in the database.

yhabteab:
fixes #81


julianbrost (changes requested):
Regarding the schema: I'm not yet entirely sure if we want to encode that information in the event type. We might also want an explicit failure state and maybe also some extra timestamps (so that you'd see if a notification was significantly delayed or something like that). But for now, probably keep the schema as-is.

internal/incident/incident.go
```diff
@@ -533,6 +481,74 @@ func (i *Incident) RestoreEscalationStateRules(states []*EscalationState) {
...
+	}
+
+	managed := i.HasManager()
+	contactChannels := make(recipientsChannelType)
+	escalationRecipients := make(map[recipient.Key]bool)
+	for escalationID := range i.EscalationState {
+		escalation := i.runtimeConfig.GetRuleEscalation(escalationID)
+		for c, channels := range i.getEscalationRecipientsChannel(escalation, t) {
+			escalationRecipients[recipient.ToKey(c)] = true
+			contactChannels[c] = channels
```
julianbrost, line 528: Wouldn't you want to merge `channels` into `contactChannels[c]` if the latter already contains something? (Maybe it would also make sense to change `getEscalationRecipientsChannel()` so that it adds everything to a map given as an input instead of returning a new map.)

internal/incident/incident.go
```diff
@@ -533,6 +481,74 @@ func (i *Incident) RestoreEscalationStateRules(states []*EscalationState) {
 	}
 }
 
+// getEscalationRecipientsChannel returns the configured channel of all the provided escalation recipients.
+func (i *Incident) getEscalationRecipientsChannel(escalation *rule.Escalation, t time.Time) recipientsChannelType {
+	if i.Recipients == nil {
+		i.Recipients = make(map[recipient.Key]*RecipientState)
+	}
```
julianbrost, line 488: I don't see why this function initializes `i.Recipients` given that it never writes to that map anyways.

internal/incident/incident.go
```diff
@@ -533,6 +481,74 @@ func (i *Incident) RestoreEscalationStateRules(states []*EscalationState) {
...
+
+	return contactChannels
+}
+
+// getRecipientsChannel returns all the configured channels of the current incident and escalation recipients.
+func (i *Incident) getRecipientsChannel(t time.Time) recipientsChannelType {
+	if i.Recipients == nil {
+		i.Recipients = make(map[recipient.Key]*RecipientState)
+	}
```
julianbrost, line 519: I don't see why this function initializes `i.Recipients` given that it never writes to that map anyways.

internal/incident/incident.go
```diff
@@ -533,6 +479,64 @@ func (i *Incident) RestoreEscalationStateRules(states []*EscalationState) {
...
+	managed := i.HasManager()
+	contactChannels := make(recipientsChannelType)
+	for escalationID := range i.EscalationState {
+		escalation := i.runtimeConfig.GetRuleEscalation(escalationID)
+		i.getEscalationRecipientsChannel(escalation, contactChannels, t)
+	}
+
+	escalationRecipients := make(map[recipient.Key]bool)
+	for contact := range contactChannels {
+		escalationRecipients[recipient.Key{ContactID: utils.ToDBInt(contact.ID)}] = true
```
julianbrost, line 517: Shouldn't just `recipient.ToKey(contact)` work here instead of initializing `recipient.Key` manually?

internal/incident/incident.go
```diff
@@ -65,7 +66,9 @@ func (i *Incident) HasManager() bool {
 }
 
 // ProcessEvent processes the given event for the current incident.
-func (i *Incident) ProcessEvent(ctx context.Context, tx *sqlx.Tx, ev event.Event, created bool) error {
+// If non-acknowledgement event is given, this returns pending incident notification histories of type *HistoryRow
+// and are used to notify the recipients based on these entries.
+func (i *Incident) ProcessEvent(ctx context.Context, tx *sqlx.Tx, ev event.Event, created bool) ([]*HistoryRow, error) {
```
julianbrost, line 71: `HistoryRow` has lots of fields and it's a bit unclear, which ones are actually used. I think a dedicated type containing the very specific information needed would be nicer (so something like contact, channel, ..., and incident history entry ID to update with the result).

internal/incident/incident.go
```diff
@@ -533,6 +479,64 @@ func (i *Incident) RestoreEscalationStateRules(states []*EscalationState) {
...
+				if contactChannels[contact] == nil {
+					contactChannels[contact] = make(map[string]struct{})
+				}
+				contactChannels[contact][contact.DefaultChannel] = struct{}{}
+			}
+		}
+	}
+
+	return contactChannels
+}
```
julianbrost, line 538: Please add a few comments within the function that say what the individual parts do.

internal/incident/incident.go
```diff
@@ -533,6 +479,64 @@ func (i *Incident) RestoreEscalationStateRules(states []*EscalationState) {
...
+	contactChannels := make(recipientsChannelType)
+	for escalationID := range i.EscalationState {
+		escalation := i.runtimeConfig.GetRuleEscalation(escalationID)
+		i.getEscalationRecipientsChannel(escalation, contactChannels, t)
+	}
+
+	escalationRecipients := make(map[recipient.Key]bool)
+	for contact := range contactChannels {
+		escalationRecipients[recipient.Key{ContactID: utils.ToDBInt(contact.ID)}] = true
+	}
```
julianbrost, line 518: Is that extra map actually needed? The code below looks somewhat like "for all recipients in `i.Recipients`, add them with their default channel to `contactChannels` if they're not already in there", which sounds perfectly doable without this extra map.

julianbrost (changes requested):

julianbrost (changes requested):

julianbrost (changes requested):

julianbrost (changes requested):

internal/incident/incident.go
```diff
@@ -535,6 +510,37 @@ func (i *Incident) RestoreEscalationStateRules(states []*EscalationState) {
 	}
 }
 
+// getRecipientsChannel returns all the configured channels of the current incident and escalation recipients.
+func (i *Incident) getRecipientsChannel(t time.Time) recipientsChannelType {
+	contactChannels := make(recipientsChannelType)
+	// Load all escalations recipients channels
+	for escalationID := range i.EscalationState {
+		escalation := i.runtimeConfig.GetRuleEscalation(escalationID)
+		contactChannels.loadEscalationRecipientsChannel(escalation, i, t)
```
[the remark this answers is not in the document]
yhabteab: Do you mean that check?
https://github.com/Icinga/icinga-notifications/blob/a94a03dfd5114420fdf128df1555df5f5f227f6e/internal/incident/incident.go#L326-L328

This part of the code has been moved to `Incident#evaluateEscalations()`, and I don't see why it should still be needed.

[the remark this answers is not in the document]
julianbrost: > Do you mean that check?

Yes. Basically, I looked over the removed lines in the diff and figured out whether they were just moved somewhere or actually removed. And that if looked to be just removed.

In general, `GetRuleEscalation()` can return `nil`. At the moment it's unlikely, but that's probably due to the fact that at the moment, foreign key constrains pretty much prevent you from removing configuration again. But something about that has to change, so I don't think it's a bad idea to prepare for the situation where an incident still has some state for an escalation stage that has been deleted in the meantime, i.e. check for `nil` and ignore any such state.

internal/incident/incident.go
```diff
@@ -550,6 +563,31 @@ type RecipientState struct {
 	Role ContactRole
 }
 
+type recipientsChannel map[*recipient.Contact]map[int64]struct{}
```
julianbrost, line 566: A comment what it's supposed to store would be helpful (contacts, each one with a set of channel IDs). And this should probably also be reflected in the name: it stores contacts, not recipients (that's the more generic term for contact or group or schedule) and also multiple channels. So maybe something like `contactsWithChannels`.

julianbrost (changes requested):

julianbrost (changes requested):

internal/incident/incident.go
```diff
@@ -300,172 +338,116 @@ func (i *Incident) evaluateEscalations() {
...
+	chType := ch.Type
 
-			_, err := i.AddHistory(ctx, tx, hr, false)
-			if err != nil {
-				i.logger.Errorw(
-					"Failed to insert contact notified incident history", zap.String("contact", contact.String()),
-					zap.Error(err),
-				)
-			}
+	i.logger.Infof("Notify contact %q via %q", contact.FullName, chType)
```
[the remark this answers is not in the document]
yhabteab: > If there was a need to shorten the message, it should probably state the name instead of the type as there can be multiple configured channels of the same type, so the name provides more information.

This is due to the previous version of the current implementation, as there was no way to access the full channel configuration, but then forgot to adjust it accordingly.

julianbrost (approved):

PR #100, opened 2023-09-25
Reference channels by their ID

### LINKED ISSUE #88: Reference channels by ID instead of type
Currently, the channels are referenced by type in other places of the config (for example event rules). The idea was, that you probably want to configure "notify by SMS" rather than "notify by this very specific channel that happens to be SMS" as that would allow swapping out the SMS provider by changing just the channel configuration and nothing else.

However, this turned out to be quite strange in practice and introduces all kind of weird situations and complexity. For example, you have to handle multiple configurations for the same channel type and also, this means channels would have to support multiple instances internally. For example, sending notifications to multiple RocketChat instances, would require the corresponding channel implementation to support multiple instances and dispatch to them, instead of simply having multiple RocketChat channels configured.

Simply referencing a specific channel (i.e. channel configuration) would make everything much simpler in my opinion, the downsides could be mitigated by a "replace all occurrences of this channel with another one" option.

yhabteab:
resolves #88

### Associated Web PR
- https://github.com/Icinga/icinga-notifications-web/pull/125

### DIFF
--- a/internal/config/rule.go
+++ b/internal/config/rule.go
@@ -110,7 +110,7 @@ func (r *RuntimeConfig) fetchRules(ctx context.Context, tx *sqlx.Tx) error {
 		recipientLogger := r.logger.With(
 			zap.Int64("id", recipient.ID),
 			zap.Int64("escalation_id", recipient.EscalationID),
-			zap.String("channel_type", recipient.ChannelType.String))
+			zap.Int64("channel_id", recipient.ChannelID.Int64))
 
 		escalation := escalationsByID[recipient.EscalationID]
 		if escalation == nil {
@@ -165,7 +165,7 @@ func (r *RuntimeConfig) applyPendingRules() {
 					recipientLogger := r.logger.With(
 						zap.Int64("id", recipient.ID),
 						zap.Int64("escalation_id", recipient.EscalationID),
-						zap.String("channel_type", recipient.ChannelType.String))
+						zap.Int64("channel_id", recipient.ChannelID.Int64))
 
 					if recipient.ContactID.Valid {
 						id := recipient.ContactID.Int64
--- a/internal/config/verify.go
+++ b/internal/config/verify.go
@@ -17,10 +17,10 @@ func (r *RuntimeConfig) debugVerify() error {
 	if r.Channels == nil {
 		return errors.New("RuntimeConfig.Channels is nil")
 	} else {
-		for typ, channel := range r.Channels {
-			err := r.debugVerifyChannel(typ, channel)
+		for id, channel := range r.Channels {
+			err := r.debugVerifyChannel(id, channel)
 			if err != nil {
-				return fmt.Errorf("RuntimeConfig.Channels[%q] is invalid: %w", typ, err)
+				return fmt.Errorf("RuntimeConfig.Channels[%d] is invalid: %w", id, err)
 			}
 		}
 	}
@@ -94,13 +94,13 @@ func (r *RuntimeConfig) debugVerify() error {
 	return nil
 }
 
-func (r *RuntimeConfig) debugVerifyChannel(typ string, channel *channel.Channel) error {
-	if channel.Type != typ {
-		return fmt.Errorf("channel %p has type %q but is referenced as %q", channel, channel.Type, typ)
+func (r *RuntimeConfig) debugVerifyChannel(id int64, channel *channel.Channel) error {
+	if channel.ID != id {
+		return fmt.Errorf("channel %p has id %d but is referenced as %d", channel, channel.ID, id)
 	}
 
-	if other := r.Channels[typ]; other != channel {
-		return fmt.Errorf("channel %p is inconsistent with RuntimeConfig.Channels[%q] = %p", channel, typ, other)
+	if other := r.Channels[id]; other != channel {
+		return fmt.Errorf("channel %p is inconsistent with RuntimeConfig.Channels[%d] = %p", channel, id, other)
 	}
 
 	return nil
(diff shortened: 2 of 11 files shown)


schema/pgsql/upgrades/017.sql
```diff
@@ -0,0 +1,18 @@
+TRUNCATE TABLE rule_escalation_recipient;
```
julianbrost, line 1: Please try to avoid truncating tables if possible, especially if it's manually entered configuration. And in the next block, you're writing statements to actually migrate from `channel_type` to `channel_id`, so if it can be done for one table, it should be easy to do for all.

internal/config/rule.go
```diff
@@ -165,7 +165,7 @@ func (r *RuntimeConfig) applyPendingRules() {
 					recipientLogger := r.logger.With(
 						zap.Int64("id", recipient.ID),
 						zap.Int64("escalation_id", recipient.EscalationID),
-						zap.String("channel_type", recipient.ChannelType.String))
+						zap.Int64("channel_type", recipient.ChannelID.Int64))
```
julianbrost, line 168: Logging key should be adapted here as well.

schema/pgsql/upgrades/017.sql
```diff
@@ -0,0 +1,18 @@
...
+
+UPDATE contact SET default_channel = 1 WHERE default_channel LIKE '%rocketchat%';
+UPDATE contact SET default_channel = 2 WHERE default_channel LIKE '%email%';
+ALTER TABLE contact RENAME COLUMN default_channel TO default_channel_id;
+ALTER TABLE contact
+  ALTER COLUMN default_channel_id TYPE bigint USING default_channel_id::bigint,
+  ADD FOREIGN KEY (default_channel_id) REFERENCES channel(id);
```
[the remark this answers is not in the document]
julianbrost: > Because, I don't want to introduce a new `==` operator just for this.

What would be wrong with the regular `=`?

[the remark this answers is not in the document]
yhabteab: 🤦‍♂️

schema/pgsql/upgrades/017.sql
```diff
@@ -0,0 +1,21 @@
+ALTER TABLE contact ADD COLUMN default_channel_id bigint REFERENCES channel(id);
+
+UPDATE contact SET default_channel_id = (SELECT id FROM channel WHERE channel.type = 'rocketchat' LIMIT 1) WHERE default_channel = 'rocketchat';
+UPDATE contact SET default_channel_id = (SELECT id FROM channel WHERE channel.type = 'email' LIMIT 1) WHERE default_channel = 'email';
```
julianbrost, line 4: You shouldn't have to even give the channel types yourself. It should be perfectly fine to reference columns from the outer query in the subquery, so it should be possible to just do:
```sql
UPDATE contact
SET default_channel_id = (SELECT id FROM channel WHERE channel.type = contact.default_channel LIMIT 1);
```

internal/config/verify.go
julianbrost, line 1: Format strings should now use `%d` instead of `%q`.

internal/incident/incident.go
```diff
@@ -462,10 +462,18 @@ func (i *Incident) notifyContacts(ctx context.Context, tx *sqlx.Tx, ev *event.Ev
...
+		for chId := range channels {
+			chConf := i.runtimeConfig.Channels[chId]
+			if chConf == nil {
+				i.logger.Errorw("Could not find config for channel", zap.Int64("channel_id", chId))
+				continue
+			}
+
+			chType := chConf.Type
+
 			i.logger.Infof("Notify contact %q via %q", contact.FullName, chType)
```
julianbrost, line 474: This should now also include the ID, or probably even better, the name from the channel config. One side effect of this change is now that you can for example configure multiple RocketChat channels that correspond to different servers. And if you do, it's good to know which one was referenced here.

schema/pgsql/upgrades/017.sql
julianbrost, line 1: This file would probably benefit from grouping the statements per table. Currently, it's a bit unclear.

julianbrost (changes requested):
What's the state of https://github.com/Icinga/icinga-notifications-web/pull/125? Ready enough for me to test everything together?

internal/config/verify.go
```diff
@@ -94,13 +94,13 @@ func (r *RuntimeConfig) debugVerify() error {
 	return nil
 }
 
-func (r *RuntimeConfig) debugVerifyChannel(typ string, channel *channel.Channel) error {
-	if channel.Type != typ {
-		return fmt.Errorf("channel %p has type %q but is referenced as %q", channel, channel.Type, typ)
+func (r *RuntimeConfig) debugVerifyChannel(id int64, channel *channel.Channel) error {
+	if channel.ID != id {
+		return fmt.Errorf("channel %q has id %d but is referenced as %d", channel, channel.ID, id)
```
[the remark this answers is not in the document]
yhabteab: Mistakingly!

yhabteab:
> What's the state of [Icinga/icinga-notifications-web#125](https://github.com/Icinga/icinga-notifications-web/pull/125)? Ready enough for me to test everything together?

Yes

schema/pgsql/upgrades/017.sql
julianbrost, line 1: From my side, this would be the next PR to be merged that involves a schema change. If https://github.com/Icinga/icinga-notifications-web/pull/125 gets a positive review, renaming this file is the only thing that's missing here.

julianbrost:
Schema rollback in case someone else needs it during testing:

```sql
ALTER TABLE contact ADD COLUMN default_channel text;
UPDATE contact SET default_channel = (SELECT type FROM channel WHERE channel.id = contact.default_channel_id);
ALTER TABLE contact
  DROP COLUMN default_channel_id,
  ALTER COLUMN default_channel SET NOT NULL;

ALTER TABLE rule_escalation_recipient ADD COLUMN channel_type text;
UPDATE rule_escalation_recipient SET channel_type = (SELECT type FROM channel WHERE channel.id = rule_escalation_recipient.channel_id);
ALTER TABLE rule_escalation_recipient DROP COLUMN channel_id;

ALTER TABLE incident_history ADD COLUMN channel_type text;
UPDATE incident_history SET channel_type = (SELECT id FROM channel WHERE channel.id = incident_history.channel_id);
ALTER TABLE incident_history DROP COLUMN channel_id;
```

schema/pgsql/upgrades/017.sql
julianbrost: Got a positive review, so this will be next. Thus, please rename to `016.sql`.

julianbrost (changes requested):

julianbrost (approved):

PR #101, opened 2023-09-27
Enhance channel

sukhwinder33445:


pkg/rpc/rpc.go
```diff
@@ -0,0 +1,135 @@
...
+		err = fmt.Errorf("failed to write request: %w", err)
+		r.setErr(err)
+
+		return nil, err
+	}
+
+	select {
+	case response := <-promise:
+		if response.Error != "" {
+			return nil, fmt.Errorf("plugin response contains error: %s", response.Error)
```
julianbrost, line 82: I'd just use `errors.New(response.Error)` here, i.e. not prefix with an extra message. Rationale: that's the closest to what the corresponding function on the remote side returned, so `Call()` returns either what that function returned, or the special `rpc.Error`.

pkg/rpc/rpc.go
```diff
@@ -0,0 +1,148 @@
...
+
+		r.requestsMu.Lock()
+		promise := r.pendingRequests[response.Id]
+		delete(r.pendingRequests, response.Id)
+		r.requestsMu.Unlock()
+
+		if promise != nil {
+			promise <- response
+		} else {
+			r.logger.Warn("Unknown response")
```
julianbrost, line 119: Log message could be a bit more informative. Something like "Ignored response for unknown ID $id/without ID".

pkg/rpc/rpc.go
```diff
@@ -0,0 +1,148 @@
...
+	"sync"
+)
+
+type Error struct {
+	cause error
+}
+
+func (err *Error) Error() string {
+	return fmt.Sprintf("RPCError: %s", err.cause.Error())
+}
```
julianbrost, line 18: References an outdated type name in the error message, I'd go with just `"RPC error: %s"`.

pkg/rpc/rpc.go
```diff
@@ -0,0 +1,148 @@
...
+
+	pendingRequests map[uint64]chan plugin.JsonRpcResponse
+	lastRequestId   uint64
+
+	errChannel chan error
+	rpcErr     *Error
+
+	requestsMu sync.Mutex
+	encoderMu  sync.Mutex
+	errMu      sync.Mutex
```
julianbrost, line 38: Please group the mutexes together with the members they actually protect.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,136 @@
...
+	Method string          `json:"method"`
+	Params json.RawMessage `json:"params"`
+	Id     uint64          `json:"id"`
+}
+
+type JsonRpcResponse struct {
+	Result json.RawMessage `json:"result"`
+	Error  string          `json:"error"`
+	Id     uint64          `json:"id"`
+}
```
julianbrost, line 60: These types probably better fit into the `rpc` package.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,136 @@
...
+
+type Plugin interface {
+	Send(req *NotificationRequest) error
+	LoadConfig(jsonStr string) error
```
julianbrost, line 64: You can just use `json.RawMessage` as type for that parameter. Otherwise, you're just casting it around anyways.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,136 @@
...
+				} else {
+					response.Result = result
+				}
+
+			case "SetConfig":
+				if err = plugin.LoadConfig(string(request.Params)); err != nil {
+					response.Error = fmt.Errorf("failed to set plugin config: %w", err).Error()
+				}
+
+			case "SendNotification":
```
julianbrost, line 96: Names here should be consistent with the functions in the `Plugin` interface. I think it would also be a good idea to move these strings to constants.

internal/config/channel.go
```diff
@@ -57,10 +58,13 @@ func (r *RuntimeConfig) applyPendingChannels() {
 		if pendingChannel == nil {
 			delete(r.Channels, typ)
 		} else if currentChannel := r.Channels[typ]; currentChannel != nil {
-			currentChannel.ID = pendingChannel.ID
-			currentChannel.Name = pendingChannel.Name
-			currentChannel.Config = pendingChannel.Config
-			currentChannel.ResetPlugin()
+			if currentChannel.ID != pendingChannel.ID || currentChannel.Name != pendingChannel.Name || currentChannel.Config != pendingChannel.Config {
```
julianbrost, line 61: Please add a comment stating what this `if` is a workaround for.

internal/channel/channel.go
```diff
@@ -13,38 +19,88 @@ type Channel struct {
 	Type   string `db:"type"`
 	Config string `db:"config" json:"-"` // excluded from JSON config dump as this may contain sensitive information
 
-	plugin Plugin
+	Logger *logging.Logger
```
julianbrost, line 22: I don't think there's a reason to export the `Logger` attribute.

internal/channel/channel.go
```diff
@@ -13,38 +19,88 @@ type Channel struct {
...
-func (c *Channel) GetPlugin() (Plugin, error) {
-	if c.plugin == nil {
-		f := Channels[c.Type]
-		if f == nil {
-			return nil, fmt.Errorf("unknown channel type %q", c.Type)
-		}
+func (c *Channel) StartPlugin(pluginDir string) error {
+	if c.cmd != nil {
+		return nil
+	}
```
julianbrost, line 31: This function must use `c.mu`.

internal/channel/channel.go
```diff
@@ -13,38 +19,88 @@ type Channel struct {
...
+	logger.Debug("Successfully set config: ", c.Config)
+
+	return nil
 }
 
 func (c *Channel) ResetPlugin() {
-	c.plugin = nil
-}
+	if c.cmd != nil {
```
julianbrost, line 76: Same here, it should also lock `c.mu` before accessing `c.cmd`. (Hint: you have to be careful with the locking when calling this function from `StartPlugin()`.)

internal/channel/channel.go
```diff
@@ -13,38 +19,88 @@ type Channel struct {
...
+	c.cmd = cmd
+	c.rpc = rpc.NewRPC(writer, reader, logger)
+	c.mu.Unlock()
+
+	if err = c.SetConfig(c.Config); err != nil {
+		c.ResetPlugin()
+
+		return fmt.Errorf("failed to set config: %w", err)
+	}
+	logger.Debug("Successfully set config: ", c.Config)
```
julianbrost, line 70: Use `Debugw()` if you want to log the config.

internal/channel/plugin.go
```diff
@@ -0,0 +1,71 @@
...
+func (c *Channel) rpcCall(method string, params json.RawMessage) (json.RawMessage, error) {
+	result, err := c.rpc.Call(method, params)
+	if err != nil {
+		var rpcErr *rpc.Error
+
+		if errors.As(err, &rpcErr) {
+			c.ResetPlugin()
+		}
+
+		return nil, fmt.Errorf("call failed: %w", err)
```
julianbrost, line 67: I don't think "call failed" adds much useful information. And not wrapping the error here allows quite some simplification of the function:

```go
result, err := c.rpc.Call(method, params)

var rpcErr *rpc.Error
if errors.As(err, &rpcErr) {
	c.ResetPlugin()
}

return result, err
```

internal/channel/plugin.go
```diff
@@ -0,0 +1,71 @@
...
+func (c *Channel) SendNotification(contact *recipient.Contact, i contracts.Incident, ev *event.Event, icingaweb2Url string) error {
+	contactStruct := &plugin.Contact{FullName: contact.FullName}
+	for _, addr := range contact.Addresses {
+		contactStruct.Addresses = append(contactStruct.Addresses, &plugin.Address{Type: addr.Type, Address: addr.Address})
+	}
+
+	req := plugin.NotificationRequest{
+		Contact:       contactStruct,
+		Incident:      plugin.Incident{Id: i.ID(), ObjectDisplayName: i.ObjectDisplayName()},
+		Event:         &plugin.Event{Time: ev.Time, URL: ev.URL, Type: ev.Type, Severity: ev.Severity.String(), Username: ev.Username, Message: ev.Message},
```
julianbrost, line 44: Please split this into one line per attribute for better readability.

internal/channel/channel.go
sukhwinder33445: It is initialised in `config/channel.go`. Must therefore be exported to have access to this property.

pkg/rpc/rpc.go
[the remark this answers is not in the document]
julianbrost: > Also, please add documentation comments, at least for all non-trivial functions. For the `RPC` type, this should also explain the error reporting and under which circumstances you have to create a new instance.

Please also do that.

pkg/rpc/rpc.go
```diff
@@ -0,0 +1,166 @@
...
+
+	decoder *json.Decoder
+	logger  *zap.SugaredLogger
+
+	pendingRequests map[uint64]chan JsonRpcResponse
+	lastRequestId   uint64
+	requestsMu      sync.Mutex
+
+	errChannel chan struct{} // never transports a value, only closed through setErr() to signal an occurred error
+	rpcErr     *Error        // only initialised via setErr(), if a rpc (non-skippable) error has occurred
```
julianbrost, line 49: What exactly do you mean by "non-skippable"? Fatal/non-recoverable?

pkg/rpc/rpc.go
```diff
@@ -0,0 +1,166 @@
...
+	errMu      sync.Mutex
+}
+
+// NewRPC creates and returns an RPC object and triggers a go routine to assign incoming responses to given requests.
+// The returned object cannot be used further if an error has occurred and must be recreated.
+func NewRPC(writer io.WriteCloser, reader io.Reader, logger *zap.SugaredLogger) *RPC {
```
julianbrost, line 55: Starting a go routine is more of an implementation detail and not really relevant for the caller. Also, I'd move the documentation for the error handling to the functions that can actually return an error.

pkg/rpc/rpc.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	go rpc.processResponses()
+
+	return rpc
+}
+
+// Call send a request with given parameters and a unique id (to assign responses to requests).
+// Returns the response result or error
+func (r *RPC) Call(method string, params json.RawMessage) (json.RawMessage, error) {
```
julianbrost, line 72: Again, ID assignment is an implementation detail, these are never visible to the caller.

The description of the returned error should be more verbose here: two different kinds of error can be returned: errors of type `*rpc.Error` mean that the communication failed and that future calls on this instance won't work and a new `*RPC` has to be created, whereas other error just mean that the remote function returned some error (that's non-fatal for the RPC object).

pkg/rpc/rpc_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"encoding/json"
+	"fmt"
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/zap/zaptest"
+	"io"
+	"sync"
+	"testing"
+)
+
+func Test_newRPC(t *testing.T) {
```
julianbrost, line 13: Just name this `TestRPC`. It tests more than just that `NewRPC` function.

julianbrost (changes requested):
I'm still missing some kind of basic documentation for the communication protocol. Even if it's just a few sentences with some example JSON messages. In general, it would be really nice to have some (almost) ready to use examples that can be used for testing the plugins individually, similar to the examples in [`internal/event/examples/`](https://github.com/Icinga/icinga-notifications/tree/5e5a117ca7e714b4b13091ea35aff968bf105dcc/internal/event/examples), which you can just POST to the daemon using curl to feed it with some simple events (almost ready to use because you don't want real hostnames (if in doubt use `example.invalid`) or credentials in there).

pkg/rpc/rpc.go
sukhwinder33445: Yes. Changed it to Fatal/non-recoverable.

doc/channel-plugin.md
julianbrost, line 1: That's not really what I had in mind, unfortunately.

The goal of this documentation should be to explain the interface without referring to the implementation too much. So if it references the implementation, it should so so after explaining the protocol, more as a hint where what was just explained is implemented.

This file should also include the following information: How is the protocol structured? Requests, responses, both sent as JSON-formatted messages, inspired by JSON-RPC, out-of-order responses with request/response IDs, and so on.

julianbrost (changes requested):

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,137 @@
...
+	Username string    `json:"username"`
+	Message  string    `json:"message"`
+}
+
+type NotificationRequest struct {
+	Contact       *Contact `json:"contact"`
+	Incident      Incident `json:"incident"`
+	Event         *Event   `json:"event"`
+	IcingaWeb2Url string   `json:"icingaweb2_url"`
+}
```
julianbrost, line 56: Part of documenting the interface between daemon and channel plugins is to check if the structure makes sense and define a clean interface. As just discussed, that could be improved a bit.

Currently, object attributes are split across `incident` and `event` keys in the JSON. It would make sense to add an extra key `object` that contains all the object related attributes (name, url, tags).

Additionally, the channel doesn't really need the Icinga Web 2 URL. It's only used to generate the incident URL, so it would make more sense to add this as `url` within `incident` instead.

internal/channel/plugin.go
```diff
@@ -0,0 +1,80 @@
...
+
+	if !strings.HasSuffix(icingaweb2Url, "/") {
+		icingaweb2Url = icingaweb2Url + "/"
+	}
+
+	incidentUrl := fmt.Sprintf("Incident: %snotifications/incident?id=%d\n", icingaweb2Url, i.ID())
+
+	req := plugin.NotificationRequest{
+		Contact:  contactStruct,
+		Object:   &plugin.Object{Name: i.ObjectDisplayName(), Url: ev.URL, Tags: ev.Tags},
```
julianbrost, line 50: You forgot to initialize `ExtraTags` here. Also, these are enough fields so that splitting the initialization across several lines is probably a good idea.

doc/channel-plugin.md
julianbrost, line 1: I'd split this file into two major sections:
1. How does the protocol work in general, irrespective of concrete message types.
2. Details about the messages in use.

julianbrost (changes requested):

internal/channel/examples/set-config.json
```diff
@@ -0,0 +1,9 @@
+{
+  "method": "SetConfig",
+  "params": {
+    "host": "example.com",
+    "port": "25",
+    "from": "notifications-daemon@test.com"
```
julianbrost, line 6: Same thing here regarding example domains.

internal/channel/plugin.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (c *Channel) SendNotification(contact *recipient.Contact, i contracts.Incident, ev *event.Event, icingaweb2Url string) error {
+	contactStruct := &plugin.Contact{FullName: contact.FullName}
+	for _, addr := range contact.Addresses {
+		contactStruct.Addresses = append(contactStruct.Addresses, &plugin.Address{Type: addr.Type, Address: addr.Address})
+	}
+
+	if !strings.HasSuffix(icingaweb2Url, "/") {
+		icingaweb2Url = icingaweb2Url + "/"
+	}
```
julianbrost, line 44: You can use `+=` here.

internal/channel/plugin.go
```diff
@@ -0,0 +1,86 @@
...
+		Contact: contactStruct,
+		Object: &plugin.Object{
+			Name:      i.ObjectDisplayName(),
+			Url:       ev.URL,
+			Tags:      ev.Tags,
+			ExtraTags: ev.ExtraTags,
+		},
+		Incident: &plugin.Incident{
+			Id:  i.ID(),
+			Url: fmt.Sprintf("%snotifications/incident?id=%d\n", icingaweb2Url, i.ID()),
```
julianbrost, line 56: The URL should not include a trailing `\n`.

internal/channel/channel.go
```diff
@@ -13,38 +19,110 @@ type Channel struct {
...
-		if err != nil {
-			return nil, fmt.Errorf("error initializing channel type %q: %w", c.Type, err)
-		}
+	if c.cmd != nil {
+		return nil
+	}
+
+	if matched, _ := regexp.MatchString("^[a-zA-Z0-9]*$", c.Type); !matched {
+		return fmt.Errorf("channel type must only contain a-zA-Z0-9, %s given", c.Type)
+	}
```
julianbrost, line 38: Prefer `%q` when arbitrary strings (as in you don't exactly know what can be in the string) as this automatically escapes most strange things.

doc/channel-plugin.md
```diff
@@ -0,0 +1,58 @@
...
+Channel plugin uses stdin and stdout as json format to communicate. The request and response structure is inspired by
+JSON-RPC.
+
+### Request
+
+The request must be in json format and should contain following keys:
+
+- `method`: The request method to call
+- `params`: The params for request method
+- `id`: Unsigned int value. Require to assign the response to it's request as responses can be out of order
```
julianbrost, line 14: **Required** to assign the response to **its** request as responses can be **sent** out of order

doc/channel-plugin.md
julianbrost, line 1: When referring to JSON in text (and not as part of a package or file name), please write it in upper-case.

doc/channel-plugin.md
```diff
@@ -0,0 +1,58 @@
+## Channel Plugin
+
+Channel plugins are processes that run independently of each other.
+
+Channel plugin uses stdin and stdout as json format to communicate. The request and response structure is inspired by
```
julianbrost, line 5: I'd write the first sentence something more like this:

> Channel plugins receive JSON-formatted requests on stdin and reply with JSON-formatted responses on stdout.

It should also be made clear that these are long-running processes and receive many requests over their lifetime.

doc/channel-plugin.md
```diff
@@ -0,0 +1,58 @@
...
+
+The request must be in json format and should contain following keys:
+
+- `method`: The request method to call
+- `params`: The params for request method
+- `id`: Unsigned int value. Require to assign the response to it's request as responses can be out of order
+
+Example:
+
+```
```
julianbrost, line 18: Use ` ```json ` to get syntax highlighting for the JSON blocks.

doc/channel-plugin.md
```diff
@@ -0,0 +1,58 @@
...
+    "method": "Add",
+    "params": {
+        "num1": 5,
+        "num2": 3
+    },
+    "id" : 2020
+}
+```
+
+The example json files can be seen under `internal/channel/examples/`.
```
julianbrost, line 29: You should get a clickable link in the rendered markdown if you use ```[`internal/channel/examples/`](../internal/channel/examples/)```.

doc/channel-plugin.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Response
+
+The response is in json format and contains following keys:
+
+- `result`: The result as json format, omitted if an error has occurred or the method does not return (e.g. setter calls)
```
julianbrost, line 35: does not return **a value**

It does return, otherwise it wouldn't send a response.

doc/channel-plugin.md
```diff
@@ -0,0 +1,58 @@
...
+    "result": 8,
+    "id" : 2020
+}
+```
+
+### Methods
+
+Currently, the channel plugin include following three methods:
+
+- `SetConfig`: Initialise the channel plugin with given config. The `params` key should contain the config. On success the response json only contains request id, ad
```
julianbrost, line 52: "ad"? That sentence looks unfinished.

Should also say that the config is specific to the plugin and that the plugin defines what it expects there.

doc/channel-plugin.md
```diff
@@ -0,0 +1,58 @@
...
+  See example [here](../internal/channel/examples/set-config.json)
+- `GetInfo`: Get the information about the channel e.g. Name. The `params` key can be omitted.
+  See example [here](../internal/channel/examples/get-info.json)
+- `SendNotification`: Send the notifications. The `params` key should contain the information about the contact to be
+  notified, corresponding object, the incident and the triggered event. See
+  example [here](../internal/channel/examples/send-notification.json)
```
julianbrost, line 58: I'd just use `([example](../internal/...))`, so that is looks like this:

> `MethodName`: Description ([example](../internal/...)).

julianbrost (changes requested):

doc/channel-plugin.md
```diff
@@ -0,0 +1,59 @@
+## Channel Plugin
+
+Channel plugins are processes that runs continuously and independently of each other. They receive many requests over
+their lifetime.
```
julianbrost, line 4: that **run** continuously

doc/channel-plugin.md
```diff
@@ -0,0 +1,59 @@
...
+They receive JSON-formatted requests on stdin and reply with JSON-formatted responses on stdout. The request and
+response structure is inspired by JSON-RPC.
```
julianbrost, line 6: The line break in the middle of the paragraph looks somewhat strange, I'd just continue the line (at least I fail to see a pattern why you'd add a line break there)

doc/channel-plugin.md
```diff
@@ -0,0 +1,59 @@
...
+  id.
+
+Example:
+
+```json
+{
+  "result": 8,
+  "id": 2020
+}
+```
```
julianbrost, line 46: A second example for an error response would be helpful as well.

julianbrost (changes requested):

julianbrost (changes requested):

julianbrost (changes requested):
I found a problem while testing: If the plugin process is terminated externally, a broken state is kept that results in subsequent attempts to send via that channel to panic:

```
2023/10/09 08:48:10 http: panic serving 192.168.0.16:47426: assignment to entry in nil map
goroutine 4869 [running]:
net/http.(*conn).serve.func1()
    /usr/local/go/src/net/http/server.go:1868 +0xb9
panic({0x7cbea0?, 0x8f5380?})
    /usr/local/go/src/runtime/panic.go:920 +0x270
github.com/icinga/icinga-notifications/pkg/rpc.(*RPC).Call(0xc000072a80, {0x853743, 0x10}, {0xc0002fc000, 0x26b, 0x280})
    /src/icinga-notifications/pkg/rpc/rpc.go:85 +0xf9
github.com/icinga/icinga-notifications/internal/channel.(*Channel).rpcCall(0xc000202120, {0x853743?, 0x6?}, {0xc0002fc000?, 0x2?, 0x10?})
    /src/icinga-notifications/internal/channel/plugin.go:78 +0x30
github.com/icinga/icinga-notifications/internal/channel.(*Channel).SendNotification(0xc000202120?, 0xc0002021e0, {0x8fa2a0, 0xc00029c510}, 0xc0001a9498, {0xc00002a180, 0x15})
    /src/icinga-notifications/internal/channel/plugin.go:72 +0x5c7
github.com/icinga/icinga-notifications/internal/incident.(*Incident).notifyContacts(0xc00029c510, {0x8fabe8, 0xba2060}, 0x1?, 0xc0001a9498, {{0x13?, 0xd0?}})
    /src/icinga-notifications/internal/incident/incident.go:448 +0x1130
github.com/icinga/icinga-notifications/internal/incident.(*Incident).ProcessEvent(0xc00029c510, {0x8fabe8, 0xba2060}, 0x1?, {{0xc1410d7292769eaa, 0x535fa23f78, 0xb739e0}, 0x1, {0xc00022eee8, 0x13}, ...}, ...)
    /src/icinga-notifications/internal/incident/incident.go:113 +0x6f2
github.com/icinga/icinga-notifications/internal/listener.(*Listener).ProcessEvent(0xc00032b3b0, {0x8fa390, 0xc0000d81c0}, 0xc0004d6900)
    /src/icinga-notifications/internal/listener/listener.go:155 +0xa25
net/http.HandlerFunc.ServeHTTP(0xc000580000?, {0x8fa390?, 0xc0000d81c0?}, 0xc00051fac8?)
    /usr/local/go/src/net/http/server.go:2136 +0x29
net/http.(*ServeMux).ServeHTTP(0x8085e0?, {0x8fa390, 0xc0000d81c0}, 0xc0004d6900)
    /usr/local/go/src/net/http/server.go:2514 +0x142
github.com/icinga/icinga-notifications/internal/listener.(*Listener).ServeHTTP(0xc00032b3b0, {0x8fa390, 0xc0000d81c0}, 0xc0000d8101?)
    /src/icinga-notifications/internal/listener/listener.go:46 +0x12c
net/http.serverHandler.ServeHTTP({0x8f9230?}, {0x8fa390?, 0xc0000d81c0?}, 0x6?) 
    /usr/local/go/src/net/http/server.go:2938 +0x8e
net/http.(*conn).serve(0xc00015c1b0, {0x8fab08, 0xc00030dc50})
    /usr/local/go/src/net/http/server.go:2009 +0x5f4
created by net/http.(*Server).Serve in goroutine 1
    /usr/local/go/src/net/http/server.go:3086 +0x5cb
```

To reproduce:
1. Start everything and make sure a notification is sent (i.e. there is a `rocketchat` child process for example).
2. Terminate that running child process using SIGTERM.
3. Trigger another notification via that channel.

julianbrost:
Please show examples of the log messages (all levels, including debug) for (1) an orderly restart (for example after a config change) and (2) an unexpected restart (for example after a plugin crash).

sukhwinder33445:
### Ordinary start:
Screenshot 2023-10-10 at 11 13 33

### Stop plugin on config change:
Screenshot 2023-10-10 at 11 12 07

### Unexpected stop:
Screenshot 2023-10-10 at 11 15 14

PR #104, opened 2023-09-29
Add git versions for builds and http responses

### LINKED ISSUE #103: Include Git version information in builds
It would be helpful to be able to determine which source a daemon binary was built for, therefore, it would be useful include the git commit (we don't have versions number yet) in the builds and show it in a few places:

1. `--version` command line switch
2. Log message on daemon startup
3. `Server:` header in HTTP responses

The Icinga DB project contains code that can be used for this.

sukhwinder33445:
resolves #103

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Include version information on `git archive'
+/internal/version.go export-subst
--- a/cmd/icinga-notifications-daemon/main.go
+++ b/cmd/icinga-notifications-daemon/main.go
@@ -4,20 +4,37 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"github.com/icinga/icinga-notifications/internal"
 	"github.com/icinga/icinga-notifications/internal/config"
 	"github.com/icinga/icinga-notifications/internal/listener"
 	"github.com/icinga/icingadb/pkg/logging"
 	"github.com/icinga/icingadb/pkg/utils"
 	"go.uber.org/zap"
 	"os"
+	"runtime"
 	"time"
 )
 
 func main() {
 	var configPath string
+	var showVersion bool
 
 	flag.StringVar(&configPath, "config", "", "path to config file")
+	flag.BoolVar(&showVersion, "version", false, "print version")
 	flag.Parse()
+
+	if showVersion {
+		// reuse internal.Version.print() once the project name is configurable
+		fmt.Println("Icinga Notifications version:", internal.Version.Version)
+		fmt.Println()
+
+		fmt.Println("Build information:")
+		fmt.Printf("  Go version: %s (%s, %s)\n", runtime.Version(), runtime.GOOS, runtime.GOARCH)
+		if internal.Version.Commit != "" {
+			fmt.Println("  Git commit:", internal.Version.Commit)
+		}
+		return
+	}
 
 	if configPath == "" {
 		_, _ = fmt.Fprintln(os.Stderr, "missing -config flag")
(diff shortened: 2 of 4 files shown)


internal/listener/listener.go
```diff
@@ -46,6 +47,8 @@ func (l *Listener) Run() error {
 }
 
 func (l *Listener) ProcessEvent(w http.ResponseWriter, req *http.Request) {
+	w.Header().Set("Server", "icinga-notifications/"+internal.Version.Version)
```
julianbrost, line 50: It would be nice if this was not just specifically set in this response handler but instead consistent for all responses.

Have a look at the [`ListenAndServe()`](https://pkg.go.dev/net/http#ListenAndServe) function as well as the [`Handler`](https://pkg.go.dev/net/http#Handler) and [`ServeMux`](https://pkg.go.dev/net/http#ServeMux) types from the `net/http` packages and think about how you can put together/wrap them to achieve this.

.gitattributes
julianbrost, line 1: Please add a newline at the end of the file.

cmd/icinga-notifications-daemon/main.go
```diff
@@ -15,10 +16,17 @@ import (
 
 func main() {
 	var configPath string
+	var version bool
```
julianbrost, line 19: I'd rename the variable (and just the variable) to `showVersion`. Better expresses what the value means.

cmd/icinga-notifications-daemon/main.go
```diff
@@ -43,6 +51,7 @@ func main() {
 	}
 
 	logger := logs.GetLogger()
+	logger.Infof("Starting daemon with version: %s", internal.Version.Version)
```
julianbrost, line 54: I think "with version" sounds a bit strange. I'd just go for "Starting Icinga Notifications daemon (v1.2.3-abcd)" (with or without parenthesis, no strong preference there)

julianbrost (changes requested):

cmd/icinga-notifications-daemon/main.go
sukhwinder33445: I prefer parenthesis.

cmd/icinga-notifications-daemon/main.go
```diff
@@ -15,10 +16,17 @@ import (
...
 func main() {
 	var configPath string
+	var showVersion bool
 
 	flag.StringVar(&configPath, "config", "", "path to config file")
+	flag.BoolVar(&showVersion, "version", false, "print version")
 	flag.Parse()
 
+	if showVersion {
+		internal.Version.Print()
```
julianbrost, line 26: Good that I built `github.com/icinga/icingadb/pkg/version` somewhat reusable. Not so good that I forget that you might want to replace the project name if you actually reuse it. :facepalm: 

So probably copy [these few lines](https://github.com/Icinga/icingadb/blob/c20f066bd87528960ec095c2e6fd88d014156b6c/pkg/version/version.go#L83-L90) here for now (I don't really care about the system information at the moment).

internal/version.go
```diff
@@ -0,0 +1,8 @@
+package internal
+
+import "github.com/icinga/icingadb/pkg/version"
+
+// Version contains version and Git commit information.
+//
+// The placeholders are replaced on `git archive` using the `export-subst` attribute.
+var Version = version.Version("0.0.0", "$Format:%(describe)$", "$Format:%H$")
```
julianbrost, line 8: Just for the record: `$Format:%(describe)$` does not work properly when there are no tags in the repository. This means that at the moment, if one actually uses `git archive` (including for example downloading the source from GitHub as a tarball), it would report an empty string as version. But that will fix itself once there's a tag in this repo.

julianbrost (changes requested):

julianbrost (approved):

PR #105, opened 2023-10-04
Flexible ID tags

sukhwinder33445:


schema/pgsql/upgrades/018.sql
julianbrost, line 1: Please add a trailing newline to the file.

schema/pgsql/upgrades/018.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE object_id_tag (
+    object_id bytea NOT NULL REFERENCES object(id),
+    tag text NOT NULL,
+    value text NOT NULL,
+
+    CONSTRAINT pk_object_id_tag PRIMARY KEY (object_id, tag)
+);
+
+INSERT INTO object_id_tag (object_id, tag, value) SELECT id, 'host' AS tag, host FROM object;
+INSERT INTO object_id_tag (object_id, tag, value) SELECT id, 'service' AS value, service FROM object WHERE service IS NOT NULL;
```
julianbrost, line 10: When giving explicit column names for the select statement, it should be `'service' AS tag` (but it should be just fine without any column aliases).

julianbrost (changes requested):
The schema upgrade script will probably have to be renamed as well before merging the PR. But as far as I can tell, the corresponding changes that will be needed in notifications-web were not started yet, were they? (So I don't want to say it will be the next one and should use `016.sql` yet.)

julianbrost:
Changes look good so far from just looking over the code. I'll continue with testing once a corresponding web PR is ready (so that I don't break my test setup while doing so).

julianbrost:
@yhabteab Can you do me a favor and test this, especially in conjunction with https://github.com/Icinga/icinga-notifications-web/pull/133?

Also, in case anyone needs it, I already have a SQL snippet lying around to roll back that schema change:
```sql
ALTER TABLE object ADD COLUMN host text, ADD COLUMN service text;
UPDATE object SET host = (SELECT value FROM object_id_tag WHERE object_id = id AND tag = 'host');
UPDATE object SET service = (SELECT value FROM object_id_tag WHERE object_id = id AND tag = 'service');
ALTER TABLE object ALTER COLUMN host SET NOT NULL;
```

yhabteab (changes requested):

yhabteab (approved):

Issue #107, opened 2023-10-06
Query channel plugin information from all available plugins and write it to the database


julianbrost:
This issue is pretty much a follow-up to #82 where the corresponding PR #101 moved the channel implementation into separate processes.

The task for this issue is to actually make use of the `GetInfo` method already added. To make use of this, on startup, the daemon should list all available plugins in the plugin directory, start them and call their `GetInfo` method and write the obtained information to a new table in the database. The purpose of this table will be so that notifications-web can use it to list available channel types as well as use it to provide a form for the channel configuration and populate the `channel.config` column with corresponding JSON.

For now, I suggest the following information should be added to the plugin info and therefore the database:
- Plugin name
- Plugin version
- Author information (name, e-mail, URL)
- Configuration configuration options: a list of form fields that notifications-web should present when configuring a channel of that particular type

For completeness, my suggestion for the latter would be something like that, but the exact details should be coordinated with notifications-web. For the daemon, this will pretty much be an opaque value (it should just write the JSON into the database without interpreting it).

```json
[
    {"name": "url", "label": "RocketChat Instance", "type": "text", "required": true, "help": "Base URL of the RocketChat installation, for example https://chat.example.com/"},
    {"name": "user_id", "label": "User ID", "type": "text", "required": true, "help": "User ID associated with the access token"},
    {"name": "token", "label": "Token", "type": "password", "required": true, "help": "Personal access token for the bot user"}
]
```

nilmerg:
I'd suggest to generalize as much as possible.

## Types

* string (single line of free text)
* text (multiple lines of free text)
* bool (yes or no)
* option (single) or options (multiple) with a property called `options` which holds a map of value/label pairs
* secret (single line of sensible information)

## Label

Should be a map of locale/message pairs. A locale should use the standard format (*language_REGION*) and it should be assumed to be encoded with UTF8. (Without the suffix in the locale)

## Help

Should be a map of locale/message pairs. A locale should use the standard format (*language_REGION*) and it should be assumed to be encoded with UTF8. (Without the suffix in the locale)

## Default

A default value should be optional. If given, the UI displays it, but doesn't write it back to the configuration.

## Required

Should be a boolean, defining whether the option is mandatory or not.

## Structure

```
{
    "name": string,
    "type": string,
    "label": object,
    ["options": object,]
    ["help": object,]
    ["default": mixed,]
    ["required": bool]
}
```

PR #108, opened 2023-10-09
Introduce plugin table

sukhwinder33445:
resolves https://github.com/Icinga/icinga-notifications/issues/107

- [x] change upgrade schema name -- no change needed

### DIFF
--- a/cmd/channel/email/main.go
+++ b/cmd/channel/email/main.go
@@ -111,30 +111,6 @@ func (ch *Email) GetInfo() *plugin.Info {
 			},
 			Placeholder: "icinga@example.com",
 		},
-		{
-			Name: "password",
-			Type: "secret",
-			Label: map[string]string{
-				"en_US": "Password",
-				"de_DE": "Passwort",
-			},
-		},
-		{
-			Name: "tls",
-			Type: "bool",
-			Label: map[string]string{
-				"en_US": "TLS / SSL",
-				"de_DE": "TLS / SSL",
-			},
-		},
-		{
-			Name: "tls_certcheck",
-			Type: "bool",
-			Label: map[string]string{
-				"en_US": "Certificate Check",
-				"de_DE": "Zertifikat prüfen",
-			},
-		},
 	}
 
 	configAttrs, err := json.Marshal(elements)
--- a/cmd/channel/email/main.go
+++ b/cmd/channel/email/main.go
@@ -2,9 +2,12 @@ package main
 
 import (
 	"bytes"
+	"database/sql"
 	"encoding/json"
 	"fmt"
+	"github.com/icinga/icinga-notifications/internal"
 	"github.com/icinga/icinga-notifications/pkg/plugin"
+	"github.com/icinga/icingadb/pkg/types"
 	"net"
 	"net/smtp"
 	"os"
@@ -80,7 +83,72 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
 }
 
 func (ch *Email) GetInfo() *plugin.Info {
-	return &plugin.Info{Name: "Email"}
+	elements := []*plugin.ConfigOption{
+		{
+			Name: "host",
+			Type: "string",
+			Label: map[string]string{
+				"en_US": "SMTP Host",
+				"de_DE": "SMTP Host",
+			},
+		},
+		{
+			Name: "port",
+			Type: "number",
+			Label: map[string]string{
+				"en_US": "SMTP Port",
+				"de_DE": "SMTP Port",
+			},
+			Min: types.Int{NullInt64: sql.NullInt64{Int64: 0, Valid: true}},
+			Max: types.Int{NullInt64: sql.NullInt64{Int64: 65535, Valid: true}},
+		},
+		{
+			Name: "from",
+			Type: "string",
+			Label: map[string]string{
+				"en_US": "From",
+				"de_DE": "Von",
+			},
+			Placeholder: "icinga@example.com",
+		},
+		{
+			Name: "password",
+			Type: "secret",
+			Label: map[string]string{
+				"en_US": "Password",
+				"de_DE": "Passwort",
+			},
+		},
+		{
+			Name: "tls",
+			Type: "bool",
+			Label: map[string]string{
+				"en_US": "TLS / SSL",
+				"de_DE": "TLS / SSL",
+			},
+		},
+		{
+			Name: "tls_certcheck",
+			Type: "bool",
+			Label: map[string]string{
+				"en_US": "Certificate Check",
+				"de_DE": "Zertifikat prüfen",
+			},
+		},
+	}
+
+	configAttrs, err := json.Marshal(elements)
+	if err != nil {
+		panic(err)
+	}
+
+	return &plugin.Info{
+		Type:             "email",
+		Name:             "Email",
+		Version:          internal.Version.Version,
+		Author:           "Icinga GmbH",
+		ConfigAttributes: configAttrs,
+	}
 }
 
 func (ch *Email) GetServer() string {
--- a/cmd/channel/email/main.go
+++ b/cmd/channel/email/main.go
@@ -119,7 +119,6 @@ func (ch *Email) GetInfo() *plugin.Info {
 	}
 
 	return &plugin.Info{
-		Type:             "email",
 		Name:             "Email",
 		Version:          internal.Version.Version,
 		Author:           "Icinga GmbH",
--- a/internal/channel/channel.go
+++ b/internal/channel/channel.go
@@ -20,34 +20,40 @@ type Channel struct {
 
 	Logger *zap.SugaredLogger
 
-	newConfigCh chan struct{}
-	pluginCh    chan *Plugin
+	restartCh chan newConfig
+	pluginCh  chan *Plugin
 
 	pluginCtx       context.Context
 	pluginCtxCancel func()
 }
 
+// newConfig helps to store the channel's updated properties
+type newConfig struct {
+	ctype  string
+	config string
+}
+
 // Start initializes the channel and starts the plugin in the background
 func (c *Channel) Start(ctx context.Context, logger *zap.SugaredLogger) {
 	c.Logger = logger
-	c.newConfigCh = make(chan struct{})
+	c.restartCh = make(chan newConfig)
 	c.pluginCh = make(chan *Plugin)
 	c.pluginCtx, c.pluginCtxCancel = context.WithCancel(ctx)
 
-	go c.runPlugin()
+	go c.runPlugin(c.Type, c.Config)
 }
 
 // initPlugin returns a new Plugin or nil if an error occurred during initialization
-func (c *Channel) initPlugin() *Plugin {
+func (c *Channel) initPlugin(cType string, config string) *Plugin {
 	c.Logger.Debug("Initializing channel plugin")
 
-	p, err := NewPlugin(c.Type, c.Logger)
+	p, err := NewPlugin(cType, c.Logger)
 	if err != nil {
 		c.Logger.Errorw("Failed to initialize channel plugin", zap.Error(err))
 		return nil
 	}
 
-	if err := p.SetConfig(c.Config); err != nil {
+	if err := p.SetConfig(config); err != nil {
 		c.Logger.Errorw("Failed to set channel plugin config, terminating the plugin", zap.Error(err))
 		p.Stop()
 		return nil
@@ -57,9 +63,9 @@ func (c *Channel) initPlugin() *Plugin {
 }
 
 // runPlugin is called as go routine to initialize and maintain the plugin by receiving signals on given chan(s)
-func (c *Channel) runPlugin() {
+func (c *Channel) runPlugin(initType string, initConfig string) {
 	var currentlyRunningPlugin *Plugin
-
+	cType, config := initType, initConfig
 	// Helper function for the following loop to stop a running plugin. Does nothing if no plugin is running.
 	stopIfRunning := func() {
 		if currentlyRunningPlugin != nil {
@@ -79,7 +85,7 @@ func (c *Channel) runPlugin() {
 
 	for {
 		if currentlyRunningPlugin == nil {
-			currentlyRunningPlugin = c.initPlugin()
+			currentlyRunningPlugin = c.initPlugin(cType, config)
 		}
 
 		select {
@@ -88,8 +94,9 @@ func (c *Channel) runPlugin() {
 			stopIfRunning()
 
 			continue
-		case <-c.newConfigCh:
-			c.Logger.Debug("Receive reload plugin config signal")
+		case reload := <-c.restartCh:
+			c.Logger.Debug("Received new plugin config")
+			cType, config = reload.ctype, reload.config
 			stopIfRunning()
 
 			continue
@@ -123,10 +130,10 @@ func (c *Channel) Stop() {
 	c.pluginCtxCancel()
 }
 
-// ReloadConfig sends a signal to reload the channel plugin config
-func (c *Channel) ReloadConfig() {
-	c.Logger.Info("Reloading the channel plugin config")
-	c.newConfigCh <- struct{}{}
+// Restart signals to restart the channel plugin with the updated channel config
+func (c *Channel) Restart() {
+	c.Logger.Info("Restarting the channel plugin with the updated channel config")
+	c.restartCh <- newConfig{c.Type, c.Config}
 }
 
 // Notify prepares and sends the notification request, returns a non-error on fails, nil on success
--- a/internal/channel/channel.go
+++ b/internal/channel/channel.go
@@ -59,6 +59,8 @@ func (c *Channel) initPlugin(cType string, config string) *Plugin {
 		return nil
 	}
 
+	p.logger.Info("Successfully started channel plugin")
+
 	return p
 }
 
@@ -67,11 +69,15 @@ func (c *Channel) runPlugin(initType string, initConfig string) {
 	var currentlyRunningPlugin *Plugin
 	cType, config := initType, initConfig
 	// Helper function for the following loop to stop a running plugin. Does nothing if no plugin is running.
-	stopIfRunning := func() {
+	stopIfRunning := func() (int, bool) {
 		if currentlyRunningPlugin != nil {
+			pid := currentlyRunningPlugin.Pid()
 			currentlyRunningPlugin.Stop()
 			currentlyRunningPlugin = nil
+			return pid, true
 		}
+
+		return 0, false
 	}
 
 	// Helper function for the following loop to receive from rpc.Done
@@ -90,19 +96,20 @@ func (c *Channel) runPlugin(initType string, initConfig string) {
 
 		select {
 		case <-rpcDone():
-			c.Logger.Debug("Receive plugin error signal")
-			stopIfRunning()
+			if pid, stopped := stopIfRunning(); stopped {
+				c.Logger.Warnw("Channel plugin crashed", zap.Int("pid", pid))
+			}
 
 			continue
 		case reload := <-c.restartCh:
-			c.Logger.Debug("Received new plugin config")
 			cType, config = reload.ctype, reload.config
 			stopIfRunning()
 
 			continue
 		case <-c.pluginCtx.Done():
-			c.Logger.Debug("Receive plugin stop signal")
-			stopIfRunning()
+			if pid, stopped := stopIfRunning(); stopped {
+				c.Logger.Infow("Successfully stopped channel plugin", zap.Int("pid", pid))
+			}
 
 			return
 		case c.pluginCh <- currentlyRunningPlugin:
@@ -126,13 +133,12 @@ func (c *Channel) getPlugin() *Plugin {
 // Stop ends the lifecycle of its plugin.
 // This should only be called when the channel is not more required.
 func (c *Channel) Stop() {
-	c.Logger.Info("Stopping channel plugin")
 	c.pluginCtxCancel()
 }
 
 // Restart signals to restart the channel plugin with the updated channel config
 func (c *Channel) Restart() {
-	c.Logger.Info("Restarting the channel plugin with the updated channel config")
+	c.Logger.Info("Restarting the channel plugin due to a config change")
 	c.restartCh <- newConfig{c.Type, c.Config}
 }
 
@@ -140,7 +146,6 @@ func (c *Channel) Restart() {
 func (c *Channel) Notify(contact *recipient.Contact, i contracts.Incident, ev *event.Event, icingaweb2Url string) error {
 	p := c.getPlugin()
 	if p == nil {
-		c.Logger.Debug("Cannot send notification, plugin could not be started")
 		return errors.New("plugin could not be started")
 	}
 
--- a/internal/channel/channel.go
+++ b/internal/channel/channel.go
@@ -53,7 +53,7 @@ func (c *Channel) initPlugin() *Plugin {
 	}
 
 	if err := p.SetConfig(c.Config); err != nil {
-		c.Logger.Errorw("Failed to set channel plugin config", zap.Error(err))
+		c.Logger.Errorw("Failed to set channel plugin config, terminating the plugin", zap.Error(err))
 		p.Stop()
 		return nil
 	}
@@ -89,17 +89,17 @@ func (c *Channel) runPlugin() {
 
 		select {
 		case <-rpcDone():
-			c.Logger.Debug("rpc.Done(): Restarting plugin")
+			c.Logger.Debug("Receive plugin error signal")
 			stopIfRunning()
 
 			continue
 		case <-c.newConfigCh:
-			c.Logger.Debug("Plugin.ReloadConfig() triggered")
+			c.Logger.Debug("Receive reload plugin config signal")
 			stopIfRunning()
 
 			continue
 		case <-c.stopPluginCh:
-			c.Logger.Debug("Stopping the channel plugin")
+			c.Logger.Debug("Receive plugin stop signal")
 			stopIfRunning()
 
 			return
@@ -125,11 +125,13 @@ func (c *Channel) runPlugin() {
 // This should only be called when the channel is not more required.
 // Multiple calls on same channel cause panic
 func (c *Channel) Stop() {
+	c.Logger.Info("Stopping channel plugin")
 	close(c.stopPluginCh)
 }
 
 // ReloadConfig sends a signal to reload the channel plugin config
 func (c *Channel) ReloadConfig() {
+	c.Logger.Info("Reloading the channel plugin config")
 	c.newConfigCh <- struct{}{}
 }
 
--- a/internal/channel/channel.go
+++ b/internal/channel/channel.go
@@ -1,17 +1,14 @@
 package channel
 
 import (
-	"bufio"
+	"errors"
 	"fmt"
-	"github.com/icinga/icinga-notifications/internal/daemon"
-	"github.com/icinga/icinga-notifications/pkg/rpc"
+	"github.com/icinga/icinga-notifications/internal/contracts"
+	"github.com/icinga/icinga-notifications/internal/event"
+	"github.com/icinga/icinga-notifications/internal/recipient"
+	"github.com/icinga/icinga-notifications/pkg/plugin"
 	"go.uber.org/zap"
-	"io"
-	"os/exec"
-	"path/filepath"
-	"regexp"
-	"sync"
-	"time"
+	"net/url"
 )
 
 type Channel struct {
@@ -21,109 +18,152 @@ type Channel struct {
 	Config string `db:"config" json:"-"` // excluded from JSON config dump as this may contain sensitive information
 
 	Logger *zap.SugaredLogger
-	cmd    *exec.Cmd
-	rpc    *rpc.RPC
-	mu     sync.Mutex
+
+	newConfigCh    chan struct{}
+	stopPluginCh   chan struct{}
+	notificationCh chan Req
 }
 
-func (c *Channel) Start() error {
-	c.mu.Lock()
-	defer c.mu.Unlock()
+// Req is a wrapper for plugin.NotificationRequest. It provides an errCh channel for Req state.
+// Any error that occurs while processing plugin.NotificationRequest should be sent to this channel
+// to mark Req as failed.If errCh receives nil, Req was successful.
+type Req struct {
+	req   *plugin.NotificationRequest
+	errCh chan<- error
+}
 
-	if c.cmd != nil && c.rpc.Err() == nil {
+// Start initializes the channel and starts the plugin in the background
+func (c *Channel) Start(logger *zap.SugaredLogger) {
+	c.Logger = logger
+	c.newConfigCh = make(chan struct{})
+	c.stopPluginCh = make(chan struct{})
+	c.notificationCh = make(chan Req, 1)
+
+	go c.runPlugin()
+}
+
+// initPlugin returns a new Plugin or nil if an error occurred during initialization
+func (c *Channel) initPlugin() *Plugin {
+	c.Logger.Debug("Initializing channel plugin")
+
+	p, err := NewPlugin(c.Type, c.Logger)
+	if err != nil {
+		c.Logger.Errorw("Failed to initialize channel plugin", zap.Error(err))
 		return nil
 	}
 
-	if matched, _ := regexp.MatchString("^[a-zA-Z0-9]*$", c.Type); !matched {
-		return fmt.Errorf("channel type must only contain a-zA-Z0-9, %q given", c.Type)
+	if err := p.SetConfig(c.Config); err != nil {
+		c.Logger.Errorw("Failed to set channel plugin config", zap.Error(err))
+		p.Stop()
+		return nil
 	}
 
-	path := filepath.Join(daemon.Config().ChannelPluginDir, c.Type)
-
-	cmd := exec.Command(path)
-
-	writer, err := cmd.StdinPipe()
-	if err != nil {
-		return fmt.Errorf("failed to create stdin pipe: %w", err)
-	}
-
-	reader, err := cmd.StdoutPipe()
-	if err != nil {
-		return fmt.Errorf("failed to create stdout pipe: %w", err)
-	}
-
-	errPipe, err := cmd.StderrPipe()
-	if err != nil {
-		return fmt.Errorf("failed to create stderr pipe: %w", err)
-	}
-
-	go forwardLogs(errPipe, c.Logger)
-
-	if err = cmd.Start(); err != nil {
-		return fmt.Errorf("failed to start cmd: %w", err)
-	}
-	c.Logger.Debug("Cmd started successfully")
-
-	c.cmd = cmd
-	c.rpc = rpc.NewRPC(writer, reader, c.Logger)
-
-	if err = c.SetConfig(c.Config); err != nil {
-		go c.terminate(c.cmd, c.rpc)
-
-		c.cmd = nil
-		c.rpc = nil
-
-		return fmt.Errorf("failed to set config: %w", err)
-	}
-	c.Logger.Debugw("Successfully set config", zap.String("config", c.Config))
-
-	return nil
+	return p
 }
 
-func (c *Channel) Stop() {
-	c.mu.Lock()
-	defer c.mu.Unlock()
+// runPlugin is called as go routine to initialize and maintain the plugin by receiving signals on given chan(s)
+func (c *Channel) runPlugin() {
+	var currentlyRunningPlugin *Plugin
 
-	if c.cmd == nil {
-		c.Logger.Debug("channel plugin has already been stopped")
-		return
+	// Helper function for the following loop to stop a running plugin. Does nothing if no plugin is running.
+	stopIfRunning := func() {
+		if currentlyRunningPlugin != nil {
+			currentlyRunningPlugin.Stop()
+			currentlyRunningPlugin = nil
+		}
 	}
 
-	go c.terminate(c.cmd, c.rpc)
-
-	c.cmd = nil
-	c.rpc = nil
-
-	c.Logger.Debug("Stopped channel plugin successfully")
-}
-
-func forwardLogs(errPipe io.Reader, logger *zap.SugaredLogger) {
-	reader := bufio.NewReader(errPipe)
-	for {
-		line, err := reader.ReadString('\n')
-		if err != nil {
-			if err != io.EOF {
-				logger.Errorw("Failed to read stderr line", zap.Error(err))
-			}
-
-			return
+	// Helper function for the following loop to receive from rpc.Done
+	rpcDone := func() <-chan struct{} {
+		if currentlyRunningPlugin != nil {
+			return currentlyRunningPlugin.rpc.Done()
 		}
 
-		logger.Info(line)
+		return nil
+	}
+
+	for {
+		if currentlyRunningPlugin == nil {
+			currentlyRunningPlugin = c.initPlugin()
+		}
+
+		select {
+		case <-rpcDone():
+			c.Logger.Debug("rpc.Done(): Restarting plugin")
+			stopIfRunning()
+
+			continue
+		case <-c.newConfigCh:
+			c.Logger.Debug("Plugin.ReloadConfig() triggered")
+			stopIfRunning()
+
+			continue
+		case <-c.stopPluginCh:
+			c.Logger.Debug("Stopping the channel plugin")
+			stopIfRunning()
+
+			return
+		case req := <-c.notificationCh:
+			if currentlyRunningPlugin == nil {
+				currentlyRunningPlugin = c.initPlugin()
+			}
+
+			if currentlyRunningPlugin == nil {
+				c.Logger.Debug("Cannot send notification, plugin could not be started")
+				req.errCh <- errors.New("plugin could not be started")
+			} else {
+				go func(p *Plugin) {
+					// the return can take time
+					req.errCh <- p.SendNotification(req.req)
+				}(currentlyRunningPlugin)
+			}
+		}
 	}
 }
 
-// run as go routine to terminate given channel
-func (c *Channel) terminate(cmd *exec.Cmd, rpc *rpc.RPC) {
-	c.Logger.Debug("terminating channel plugin")
-	_ = rpc.Close()
+// Stop ends the lifecycle of its plugin.
+// This should only be called when the channel is not more required.
+// Multiple calls on same channel cause panic
+func (c *Channel) Stop() {
+	close(c.stopPluginCh)
+}
 
-	timer := time.AfterFunc(5*time.Second, func() {
-		c.Logger.Debug("killing the channel plugin")
-		_ = cmd.Process.Kill()
-	})
+// ReloadConfig sends a signal to reload the channel plugin config
+func (c *Channel) ReloadConfig() {
+	c.newConfigCh <- struct{}{}
+}
 
-	_ = cmd.Wait()
-	timer.Stop()
-	c.Logger.Debug("Channel plugin terminated successfully")
+// Notify prepares and sends the notification request, returns a non-error on fails, nil on success
+func (c *Channel) Notify(contact *recipient.Contact, i contracts.Incident, ev *event.Event, icingaweb2Url string) error {
+	contactStruct := &plugin.Contact{FullName: contact.FullName}
+	for _, addr := range contact.Addresses {
+		contactStruct.Addresses = append(contactStruct.Addresses, &plugin.Address{Type: addr.Type, Address: addr.Address})
+	}
+
+	icingaweb2Url, _ = url.JoinPath(icingaweb2Url, "/")
+	req := &plugin.NotificationRequest{
+		Contact: contactStruct,
+		Object: &plugin.Object{
+			Name:      i.ObjectDisplayName(),
+			Url:       ev.URL,
+			Tags:      ev.Tags,
+			ExtraTags: ev.ExtraTags,
+		},
+		Incident: &plugin.Incident{
+			Id:  i.ID(),
+			Url: fmt.Sprintf("%snotifications/incident?id=%d", icingaweb2Url, i.ID()),
+		},
+		Event: &plugin.Event{
+			Time:     ev.Time,
+			Type:     ev.Type,
+			Severity: ev.Severity.String(),
+			Username: ev.Username,
+			Message:  ev.Message,
+		},
+	}
+
+	errCh := make(chan error, 1)
+	c.notificationCh <- Req{req: req, errCh: errCh}
+
+	return <-errCh
 }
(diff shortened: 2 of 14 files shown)


julianbrost (changes requested):

internal/channel/channel.go
```diff
@@ -20,109 +18,140 @@ type Channel struct {
...
-	writer, err := cmd.StdinPipe()
+	p, err := NewPlugin(c.Type, c.Logger)
 	if err != nil {
-		return fmt.Errorf("failed to create stdin pipe: %w", err)
+		c.Logger.Errorw("Failed to initialize channel plugin", zap.Error(err))
+	} else if err := p.SetConfig(c.Config); err != nil {
+		c.Logger.Errorw("Failed to set channel plugin config", err)
+		go terminate(p)
+		p = nil
 	}
```
julianbrost, line 51: Just `return nil` from the individual error cases instead of reassigning the variable just to return it (or relying on `NewPlugin()` t return nil in this case).

internal/channel/plugin.go
```diff
@@ -1,19 +1,86 @@
...
+	cmd    *exec.Cmd
+	rpc    *rpc.RPC
+	mu     sync.Mutex
+	logger *zap.SugaredLogger
+}
+
+// NewPlugin starts and returns a new plugin instance. If the start of the plugin fails, an error is returned
+func NewPlugin(pluginName string, logger *zap.SugaredLogger) (*Plugin, error) {
+	p := &Plugin{logger: logger}
+	err := p.Start(pluginName)
```
julianbrost, line 31: There is no reason to call `Start()` anywhere else, is it? Then it should not be exported, or maybe even just merged in this function as this doesn't do much more as calling `Start()` anyways.

julianbrost (changes requested):
(So far, just the first round of findings, there'll probably be more comments in the future.)

internal/incident/incident.go
```diff
@@ -419,28 +419,17 @@ func (i *Incident) notifyContact(contact *recipient.Contact, ev *event.Event, ch
...
+	} else {
+		i.logger.Infow(
+			"Successfully sent a notification via channel plugin", zap.String("type", ch.Type), zap.String("contact", contact.FullName),
+		)
 	}
 
-	i.logger.Infow(
-		"Successfully sent a message via channel", zap.String("type", chType), zap.String("contact", contact.String()),
-	)
-
```
yhabteab, line 443: Why did remove the `return err` part?

pkg/plugin/plugin.go
```diff
@@ -18,9 +18,36 @@ const (
...
+	Required bool              `json:"required,omitempty"`
+}
+
 type Info struct {
-	Name             string          `json:"display_name"`
-	ConfigAttributes json.RawMessage `json:"config_attrs"`
+	Type             string          `db:"type" json:"type"`
+	Name             string          `db:"name" json:"name"`
+	Version          string          `db:"version" json:"version"`
+	AuthorName       string          `db:"author_name" json:"author_name"`
```
yhabteab, line 35: It's sufficient to just use `Author`.

schema/pgsql/schema.sql
yhabteab, line 1: Since you reverted the changes that channels reference plugin IDs instead of `type`, this now wouldn't work when using multiple configurations for the same plugin/channel type.

yhabteab (dismissed):
I know I'm not even good at this either, but please document at least the exported methods.

schema/pgsql/schema.sql
yhabteab: Nevermind! `plugin` is used just as a template.

internal/channel/plugin.go
```diff
@@ -27,60 +85,94 @@ func (c *Channel) GetInfo() (*plugin.Info, error) {
...
+		line, err := reader.ReadString('\n')
+		if err != nil {
+			if err != io.EOF {
+				logger.Errorw("Failed to read stderr line", zap.Error(err))
+			}
+
+			return
+		}
+
+		logger.Info(line)
```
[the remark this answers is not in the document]
sukhwinder33445: This is just a forwarding of the errorPipe logs, so no context is required here.

cmd/channel/email/main.go
```diff
@@ -80,7 +86,77 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
...
+
+	configAttrs, err := json.Marshal(elements)
+	if err != nil {
+		panic(err)
+	}
+
+	return &plugin.Info{
+		Type:             "email",
+		Name:             "Email",
+		Version:          internal.Version.Version,
```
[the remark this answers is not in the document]
sukhwinder33445: This is a built-in channel plugin. External plugins should define their own version.

pkg/plugin/plugin.go
```diff
@@ -61,11 +99,17 @@ type NotificationRequest struct {
 }
 
 type Plugin interface {
+	// GetInfo returns the corresponding plugin *Info, panics on failure
 	GetInfo() *Info
```
julianbrost, line 103: > panics on failure

I wouldn't specify that here as this function should never fail. If it does, it's a bug but stating it here like this makes it sound like this was expected behavior (if it was, `GetInfo()` should also return an `error`).

schema/pgsql/schema.sql
```diff
@@ -17,10 +17,20 @@ CREATE OR REPLACE FUNCTION anynonarrayliketext(anynonarray, text)
     $$;
 CREATE OPERATOR ~~ (LEFTARG=anynonarray, RIGHTARG=text, PROCEDURE=anynonarrayliketext);
 
+CREATE TABLE plugin (
```
julianbrost, line 20: I'd rather name this `channel_type` or maybe even `available_channel_type`.

schema/pgsql/upgrades/018.sql
```diff
@@ -0,0 +1,15 @@
...
+      name text NOT NULL,
+      version text NOT NULL,
+      author text NOT NULL,
+      config_attrs text NOT NULL,
+
+      CONSTRAINT pk_plugin PRIMARY KEY (type)
+);
+
+INSERT INTO plugin (type, name, version, author, config_attrs)
+    SELECT type, type, '0.0.0', 'Icinga GmbH', '' FROM channel;
```
julianbrost, line 12: If this stays a `select`, it should at least be a `select distinct`. But I think I'd just hard code the two currently existing types here in the schema file.

Random thought: maybe we don't even want the foreign key constraint, but that question will probably only be relevant once we want to also clean up old rows from the table. But I guess we probably want to keep the rows and only flag them as inactive, it's not like this table should get huge and keeping the information sure doesn't hurt for displaying the configuration of left-over configured channels where the channel plugin is no longer present.

internal/config/channel.go
```diff
@@ -27,8 +28,9 @@ func (r *RuntimeConfig) fetchChannels(ctx context.Context, tx *sqlx.Tx) error {
 		)
 		if channelsById[c.ID] != nil {
 			channelLogger.Warnw("ignoring duplicate config for channel type")
+		} else if matched, _ := regexp.MatchString("^[a-zA-Z0-9]*$", c.Type); !matched {
```
julianbrost, line 31: A trivial change to the regex would make it a bit more useful: `+` instead of `*` as an empty type and thereby empty file name doesn't make much sense (but wouldn't cause a big problem either).

julianbrost (changes requested):

pkg/plugin/plugin.go
```diff
@@ -18,9 +18,47 @@ const (
...
 
+// FormElement describe a form element required for the channel config
+type FormElement struct {
+	// Element name
+	Name string `json:"name"`
+	// Element type: string = text, bool = checkbox, text = textarea, option(s) = select(multi), secret = password
+	Type string `json:"type"`
+	// Element label map. Locale in the standard format (language_REGION) as key and corresponding label as value.
+	//
+	// e.g. ["en_US": "Label"]
```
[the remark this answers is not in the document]
sukhwinder33445: > Should probably be square brackets.

You mean curly brackets, right?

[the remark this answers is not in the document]
julianbrost: D'oh. Yes.

schema/pgsql/upgrades/018.sql
sukhwinder33445: Yes, the foreign key is not very helpful yet, but I had just implemented it for later use. Should I remove it for now?

pkg/plugin/plugin.go
```diff
@@ -18,9 +18,49 @@ const (
...
+	// Element type: string = text, bool = checkbox, text = textarea, option = select, options = select[multiple], secret = password
+	Type string `json:"type"`
+	// Element label map. Locale in the standard format (language_REGION) as key and corresponding label as value.
+	//
+	// e.g. {"en_US": "Save", "de_DE": "Speichern"}
+	Label map[string]string `json:"label"`
+	// Element placeholder
+	Placeholder string `json:"placeholder,omitempty"`
+	// Element title
+	Help map[string]string `json:"help,omitempty"`
```
julianbrost, line 34: Why is `Help` documented as title?

pkg/plugin/plugin.go
```diff
@@ -18,9 +18,49 @@ const (
 	MethodSendNotification = "SendNotification"
 )
 
+// FormElement describe a form element required for the channel config
+type FormElement struct {
```
julianbrost, line 22: I'd rename this type to something like `ConfigOption`. Yes, the web module will render it as a form element, but that's not all this struct represents. Take the `Name` attribute for example: that's pretty much irrelevant for the form, the only place where it matters is that it will be the key in the JSON object that's stored in the `channel.config` column.

pkg/plugin/plugin.go
```diff
@@ -18,9 +18,49 @@ const (
...
+	// Element placeholder
+	Placeholder string `json:"placeholder,omitempty"`
+	// Element title
+	Help map[string]string `json:"help,omitempty"`
+	// Element default value: bool for checkbox, string for other elements
+	Default any `json:"default,omitempty"`
+	// Set true if element is required, omit otherwise
+	Required bool `json:"required,omitempty"`
+	// Select element options map:  key => value. Only required for type option and options
+	Options map[string]string `json:"options,omitempty"`
```
julianbrost, line 40: In general, these comments should be more detailed than just repeating the type and field names.

Also, I think this would would look nicer and make the connection between field and comment more clear, if these were separated by blank lines.

pkg/plugin/plugin.go
```diff
@@ -18,9 +18,49 @@ const (
...
+
+// Info contains plugin information.
 type Info struct {
-	Name             string          `json:"display_name"`
-	ConfigAttributes json.RawMessage `json:"config_attrs"`
+	Type             string          `db:"type" json:"type"`
+	Name             string          `db:"name" json:"name"`
+	Version          string          `db:"version" json:"version"`
+	Author           string          `db:"author" json:"author"`
+	ConfigAttributes json.RawMessage `db:"config_attrs" json:"config_attrs"` // FormElement as json
```
julianbrost, line 49: The comment should at least make clear that's a JSON-encoded list/slice of `FormElement`.

internal/incident/incident.go
```diff
@@ -381,33 +382,35 @@ func (i *Incident) evaluateEscalations(ctx context.Context, tx *sqlx.Tx, ev *eve
 	return nil
 }
 
-// notifyContacts executes all the given pending notifications of the current incident.
+// notifyContacts executes all the given pending notifications of the current incident in a go routine.
```
julianbrost, line 385: With that documentation, it reads like it would be done in the background, but this function waits for it. I'd just leave the documentation here as-is as the change is not relevant for the caller.

internal/incident/incident.go
```diff
@@ -381,33 +382,35 @@ func (i *Incident) evaluateEscalations(ctx context.Context, tx *sqlx.Tx, ev *eve
...
-		if err := ctx.Err(); err != nil {
-			return err
-		}
+			return childCtx.Err()
+		})
 	}
 
-	return nil
+	return g.Wait()
 }
```
julianbrost, line 414: But is this change in any way relevant to this PR?

internal/channel/channel.go
```diff
@@ -20,109 +18,145 @@ type Channel struct {
...
+		}
 	}
-	c.Logger.Debugw("Successfully set config", zap.String("config", c.Config))
-
-	return nil
 }
 
+// Stop ends the lifecycle of its plugin.
+// This should only be called when the channel is not more required.
 func (c *Channel) Stop() {
```
julianbrost, line 116: Still doesn't say if it's safe/intended to be called multiple times.

internal/channel/plugin.go
```diff
@@ -27,60 +83,102 @@ func (c *Channel) GetInfo() (*plugin.Info, error) {
...
 
 	return err
 }
 
-func (c *Channel) SendNotification(contact *recipient.Contact, i contracts.Incident, ev *event.Event, icingaweb2Url string) error {
-	contactStruct := &plugin.Contact{FullName: contact.FullName}
-	for _, addr := range contact.Addresses {
-		contactStruct.Addresses = append(contactStruct.Addresses, &plugin.Address{Type: addr.Type, Address: addr.Address})
+// SendNotification sends the NotificationRequest as a go routine
+func (p *Plugin) SendNotification(req *plugin.NotificationRequest) {
```
julianbrost, line 94: > as a go routine

This one doesn't even start one, but it sounds like it did, while it's waiting for the response.

julianbrost (changes requested):

pkg/plugin/plugin.go
sukhwinder33445: `Name` is actually used as the name attribute of form's input elements.

pkg/plugin/plugin.go
sukhwinder33445: This is based on Johannes' suggestions. Help is used as the title attribute of the element. When the user moves the mouse pointer over an element, a tooltip is displayed with given message.

internal/channel/channel.go
sukhwinder33445: Yeah, it's not safe. And it should never call multiple times.

pkg/plugin/plugin.go
julianbrost: That implies that someone who reads "Element title" will think of the HTML title attribute. So that information would be much more helpful in the comment:

> When the user moves the mouse pointer over an element, a tooltip is displayed with given message.

cmd/channel/email/main.go
```diff
@@ -80,7 +81,76 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
...
+			},
+		},
+		{
+			Name: "tls_certcheck",
+			Type: "bool",
+			Label: map[string]string{
+				"en_US": "Certificate Check",
+				"de_DE": "Zertifikat prüfen",
+			},
+		},
```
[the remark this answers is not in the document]
sukhwinder33445: Yes, removed.

internal/incident/incident.go
julianbrost: The diff in this PR is already big enough, so I'd focus on the relevant parts here, populating the table in the database and reworking the plugin execution so that this is possible.

This looks like a change for its own PR, would make sense with or without the rest of this PR and doesn't seem needed for the rest of this PR.

julianbrost:
FYI: with #121 merged, the Go check should work properly again.

pkg/plugin/plugin.go
sukhwinder33445: > And isn't this the same form like where you select name and type

No. You are mixing up something. This information is stored directly as a property in the type plugin.info. A plugin can only have one type and one name. 

But this `Name` here is used as the name attribute for the input element of the config form. There can be more than one form elements. For example: `address`, `port`, `from` for email.

julianbrost: I don't think I'm mixing up something. With form I meant HTML form in the web module. What I wanted to say is the following: currently, this looks something like this (heavily simplified):

```html
<form class="icinga-form icinga-controls" method="POST">
  <input class="primary-submit-btn-duplicate" value="Save Changes" name="submit" type="submit">
  <input value="[...]" name="CSRFToken" required="" type="hidden">
  <input autocomplete="off" aria-required="true" value="[...]" name="name" required="" type="text">
  <select class="autosubmit" disable="" aria-required="true" name="type" required="">[...]</select>
  <input aria-required="true" value="[...]" name="url" required="" type="text">
  <input autocomplete="off" aria-required="true" value="[...]" name="user_id" required="" type="text">
  <input autocomplete="off" aria-required="true" value="[...]" name="token" required="" type="password">
  <input class="btn-remove" formnovalidate="" value="Delete" name="delete" type="submit">
  <input class="btn-primary" value="Save Changes" name="submit" type="submit">
</form>
```

Where the `<input>`/`<select>` elements with `name` being one of `submit`/`CSRFToken`/`name`/`type`/`delete` are generated by Icinga (Notifications) Web. If that was using the `FieldElement.Name` defined here unchanged in the HTML elements, these 5 names (and all others that might be added in the future) would result in a name conflict, so it would be wise for the web to prevent this from the beginning and always use something like `name="config_$name"`.

sukhwinder33445: I think I understand your point of view now. You mean, `FormElement.Name` can conflict if the web itself also provide the same name element.
I already handled this problem by using `FieldElement`. 
Web form store these values as following: `values['config'][$name]`

internal/channel/plugin.go
```diff
@@ -27,60 +83,96 @@ func (c *Channel) GetInfo() (*plugin.Info, error) {
...
-		},
-		Event: &plugin.Event{
-			Time:     ev.Time,
-			Type:     ev.Type,
-			Severity: ev.Severity.String(),
-			Username: ev.Username,
-			Message:  ev.Message,
-		},
+func forwardLogs(errPipe io.Reader, logger *zap.SugaredLogger) {
+	reader := bufio.NewReader(errPipe)
```
oxzi, line 124: A `bufio.Scanner` seems to be more fitting here. Its default behavior even covers reading lines.

internal/config/channel.go
```diff
@@ -19,6 +20,7 @@ func (r *RuntimeConfig) fetchChannels(ctx context.Context, tx *sqlx.Tx) error {
 	}
 
 	channelsById := make(map[int64]*channel.Channel)
+	regEx, _ := regexp.Compile("^[a-zA-Z0-9]+$")
```
oxzi, line 23: If you don't care about the error and expect that the static regular expression MUST compile, use `regexp.MustCompile`. By doing so, you will realize that something went wrong and don't have to discard the error.

oxzi: Also, you might wanna give the variable a more meaningful name, e.g., `channelTypeNameRe` or the like.

internal/daemonConfig/config.go
```diff
@@ -0,0 +1,77 @@
+package daemonConfig
```
oxzi, line 1: Please use a more Go-styled package name; consider <https://go.dev/blog/package-names>.

internal/daemonConfig/config.go
```diff
@@ -0,0 +1,77 @@
...
+type ConfigFile struct {
+	Listen           string                  `yaml:"listen" default:"localhost:5680"`
+	DebugPassword    string                  `yaml:"debug-password"`
+	ChannelPluginDir string                  `yaml:"channel-plugin-dir" default:"/usr/libexec/icinga-notifications/channel"`
+	Icingaweb2URL    string                  `yaml:"icingaweb2-url"`
+	Database         icingadbConfig.Database `yaml:"database"`
+	Logging          icingadbConfig.Logging  `yaml:"logging"`
+}
+
+var config *ConfigFile
```
oxzi, line 20: If this variable is used like a singleton, maybe add a short comment saying so. Currently I'm seeing access from both `Load` and `Config`.

cmd/channel/email/main.go
```diff
@@ -80,7 +81,60 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
...
+			Name: "host",
+			Type: "string",
+			Label: map[string]string{
+				"en_US": "SMTP Host",
+				"de_DE": "SMTP Host",
+			},
+		},
+		{
+			Name: "port",
+			Type: "option",
```
oxzi, line 95: Wouldn't it make more sense to have a freely selectable port, because otherwise someone will definitely be unhappy? What about the selection of the transport encryption? This cannot be read from the port per se.

.github/workflows/go.yml
```diff
@@ -28,3 +30,4 @@ jobs:
         # Enable the gosec linter w/o having to create a .golangci.yml config
         args: -E gosec
         only-new-issues: true
+        skip-cache: true
```
oxzi, line 33: Why are the already merged changes from #121 part of this PR?

internal/channel/channel.go
```diff
@@ -20,109 +18,155 @@ type Channel struct {
...
+				currentlyRunningPlugin.Stop()
+				currentlyRunningPlugin = nil
+			}
 
-	go c.terminate(c.cmd, c.rpc)
+			return
+		case req := <-c.notificationCh:
+			if currentlyRunningPlugin == nil {
+				currentlyRunningPlugin = c.initPlugin()
+			}
```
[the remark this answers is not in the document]
sukhwinder33445: This is intended to try to start the plugin again when a notification is to be sent.

internal/channel/plugin.go
```diff
@@ -27,60 +83,96 @@ func (c *Channel) GetInfo() (*plugin.Info, error) {
...
+		}
+
+		logger.Info(line)
 	}
+}
 
-	params, err := json.Marshal(req)
+// SyncPlugins upsert the plugin table with working plugins
+func SyncPlugins(channelPluginDir string, logger *logging.Logger, db *icingadb.DB) {
+	files, err := os.ReadDir(channelPluginDir)
```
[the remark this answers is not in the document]
julianbrost: Yes and no. I agree that this can be made more robust to avoid unintended behavior (exclude hidden/backup files (`.*`, `*~`) or maybe require more specific file names, check permissions, ...), but is the overall idea worse than pretty much any `*.d` config directory?

I mean we could consider something like store what's returned from `GetInfo()` currently in a separate file, so that for example you'd have the `email` binary and a `email.info` file simply containing JSON, that would save us from executing the plugins until they are actually configured, but it wouldn't change much about "if it's in this directory, it's trusted".

internal/config/channel.go
julianbrost: And move it outside of the function so that if someone would break the regex, they'd notice immediately when starting the program, not only when this code path is reached.

internal/channel/channel.go
```diff
@@ -20,109 +18,155 @@ type Channel struct {
...
 
-	c.cmd = nil
-	c.rpc = nil
+			if currentlyRunningPlugin == nil {
+				c.Logger.Debug("Cannot send notification, plugin could not be started")
+				req.errCh <- errors.New("plugin could not be started")
+			} else {
+				go func(p *Plugin) {
+					req.errCh <- p.SendNotification(req.req)
+				}(currentlyRunningPlugin)
```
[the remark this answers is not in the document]
sukhwinder33445: Yes, the channel must be buffered.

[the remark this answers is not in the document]
julianbrost: Currently, `p.SendNotification()` does all the RPC stuff and waits for the result. So it can basically take as long as a plugin might take (at the moment, there's also no timeout).

internal/channel/plugin.go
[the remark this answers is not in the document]
sukhwinder33445: Should i leave it as is for now?

internal/channel/channel.go
```diff
@@ -20,109 +18,153 @@ type Channel struct {
...
 
-	c.Logger.Debug("Stopped channel plugin successfully")
+			if currentlyRunningPlugin == nil {
+				c.Logger.Debug("Cannot send notification, plugin could not be started")
+				req.errCh <- errors.New("plugin could not be started")
+			} else {
+				go func(p *Plugin) {
+					// the return can take time
+					req.errCh <- p.SendNotification(req.req)
+				}(currentlyRunningPlugin)
```
julianbrost, line 117: After looking on this piece of code for a bit, we can also do this slightly different:
1. Remove `Req`, `notificationCh` and this case.
2. Add a `chan *Plugin` to the `Channel` struct.
3. Add a `case c.pluginCh <- currentlyRunningPlugin:` here (with no code in that case).

Now this gives a channel that you can receive from to get a `*Plugin` that can be used to send a notification. Slight caveat: the first `if currentlyRunningPlugin == nil` that attempts to start the plugin once again after a long wait is lost. But that can be addressed easily with a simple function:

```go
// getPlugin returns a fully initialized plugin that can be used for sending notifications. If there
// currently is no such plugin, for example because starting it failed, nil is returned instead.
func (c *Channel) getPlugin() *Plugin {
    p := <-c.pluginCh
    
    if p == nil {
        // The above receive might have woken runPlugin after the select was blocked for a long time.
        // In that case, a second receive gives it another chance to successfully start the plugin.
        p = <-c.pluginCh
    }

    return p
}
```

And then, `Notify()` could just do the following:

```go
p := c.getPlugin()
if p == nil {
    // TODO: error handling
}
p.SendNotification(req.req)
```

Overall, has the same effect as the current code: the value of `currentlyRunningPlugin` is passed to a different goroutine which calls `SendNotification()` on it. But this no longer needs the `Req` type, no passing around of channels for return values, and no extra goroutines need to be started.

julianbrost (changes requested):

internal/channel/channel.go
```diff
@@ -20,109 +19,149 @@ type Channel struct {
 	Config string `db:"config" json:"-"` // excluded from JSON config dump as this may contain sensitive information
 
 	Logger *zap.SugaredLogger
```
julianbrost, line 21: `Logger` is no longer accessed externally, so it can be changed to `logger` now.

internal/channel/channel.go
```diff
@@ -20,109 +19,149 @@ type Channel struct {
...
+		}
 
-	if err = cmd.Start(); err != nil {
-		return fmt.Errorf("failed to start cmd: %w", err)
-	}
-	c.Logger.Debug("Cmd started successfully")
+		var rpcDone <-chan struct{}
+		if currentlyRunningPlugin != nil {
+			rpcDone = currentlyRunningPlugin.rpc.Done()
+		}
```
julianbrost, line 70: Moving this info a function should make this loop cleaner (that function can then just be called in the select below, no manually define a variable then). Maybe also just as a closure like the helper suggested in my next comment.

internal/channel/channel.go
```diff
@@ -20,109 +19,149 @@ type Channel struct {
...
+		select {
+		case <-rpcDone:
+			c.Logger.Debug("Receive plugin error signal")
 
-	if err = c.SetConfig(c.Config); err != nil {
-		go c.terminate(c.cmd, c.rpc)
+			if currentlyRunningPlugin != nil {
+				currentlyRunningPlugin.Stop()
+				currentlyRunningPlugin = nil
+			}
```
julianbrost, line 79: This is repeated multiple times. Maybe add something like this before the loop:

```go
// Helper function for the following loop to stop a running plugin. Does nothing if no plugin is running.
stopIfRunning := func() {
	if currentlyRunningPlugin != nil {
		currentlyRunningPlugin.Stop()
		currentlyRunningPlugin = nil
	}
}

File: internal/channel/channel.go [hunk 34]
```diff
@@ -20,109 +19,149 @@ type Channel struct {
...
 	}
 
-	reader, err := cmd.StdoutPipe()
-	if err != nil {
-		return fmt.Errorf("failed to create stdout pipe: %w", err)
+	if err := p.SetConfig(c.Config); err != nil {
+		c.Logger.Errorw("Failed to set channel plugin config, terminating the plugin", zap.Error(err))
+		p.Stop()
+		return nil
 	}
```
julianbrost, line 54: This read of `c.Config` now has a race-condition with the write in `internal/config/channel.go`. Previously, starting the plugin happened while an event was processed while a read-lock on the `RuntimeConfig` was held which protected `c.Config` from concurrent modification. Now this is done in a completely independent goroutine, so that lock is no longer held. (Hint: that's the reason why I suggested `newConfigCh` to be a `chan string` instead of `chan struct{}`)

julianbrost (changes requested):

internal/channel/plugin.go
```diff
@@ -1,19 +1,90 @@
...
+	p.cmd = cmd
+	p.rpc = rpc.NewRPC(writer, reader, p.logger)
+
+	return p, nil
+}
+
+// Stop stops the plugin. Multiple calls are safe because sync.Once is used internally
+func (p *Plugin) Stop() {
+	go func() {
+		p.stopOnce.Do(func() {
```
julianbrost, line 69: You can actually swap the order in which `go` and `Do()` are nested here and only start the goroutine once. As there's nothing happening after `Do()` returns, this is safe.

internal/channel/plugin.go
```diff
@@ -1,19 +1,90 @@
...
+	}
+
+	p := &Plugin{logger: logger.With(zap.Int("pid", cmd.Process.Pid))}
+
+	p.logger.Info("Successfully started channel plugin")
+
+	go forwardLogs(errPipe, p.logger)
+
+	p.cmd = cmd
+	p.rpc = rpc.NewRPC(writer, reader, p.logger)
```
julianbrost, line 61: Now you can actually initialize all the fields right within the initializer. (`forwardLogs()` shouldn't mind if it's started after `NewRPC`, otherwise you could just do `logger = logger.With(...)` before instead of only setting it within the plugin).

internal/channel/channel.go
```diff
@@ -20,109 +19,149 @@ type Channel struct {
...
+
+		select {
+		case <-rpcDone():
+			c.Logger.Debug("Receive plugin error signal")
+			stopIfRunning()
 
-	if err = c.SetConfig(c.Config); err != nil {
-		go c.terminate(c.cmd, c.rpc)
+			continue
+		case initConfig = <-c.newConfigCh:
```
julianbrost, line 90: If you set it to a new value, it should just be called `config`.

internal/channel/plugin.go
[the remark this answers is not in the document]
oxzi: Fair enough.

However, I would still check within the `for` loop over those `files` if each file is really a file. Otherwise, when accidentally creating a subdirectory within the hopefully trusted plugin directory, creating the `NewPlugin` will fail. Furthermore, I would also deny symlinks (_or anything not being a file_) as those would dangerously widen the trusted plugin source.

internal/channel/channel.go
```diff
@@ -20,109 +19,149 @@ type Channel struct {
...
+	go c.runPlugin(c.Config)
+}
 
-	cmd := exec.Command(path)
+// initPlugin returns a new Plugin or nil if an error occurred during initialization
+func (c *Channel) initPlugin(config string) *Plugin {
+	c.Logger.Debug("Initializing channel plugin")
 
-	writer, err := cmd.StdinPipe()
+	p, err := NewPlugin(c.Type, c.Logger)
```
julianbrost, line 44: Actually, `c.Type` has the same issue [that I mentioned earlier for `c.Config`](https://github.com/Icinga/icinga-notifications/pull/108#discussion_r1380093296).

julianbrost (changes requested):

pkg/plugin/plugin.go
```diff
@@ -18,9 +18,65 @@ const (
...
+	//
+	//  string = text, int = number, bool = checkbox, text = textarea, option = select, options = select[multiple], secret = password
+	Type string `json:"type"`
+
+	// Element label map. Locale in the standard format (language_REGION) as key and corresponding label as value.
+	// Locale is assumed to be UTF-8 encoded (Without the suffix in the locale)
+	//
+	//  e.g. {"en_US": "Save", "de_DE": "Speichern"}
+	//  the first entry is used as fallback, if Icinga Web's current locale is not given in this map
+	Label map[string]string `json:"label"`
```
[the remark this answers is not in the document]
sukhwinder33445: Ohh.. i have changed it as follows:
`If the current locale of Icinga Web is not present in this map, Icinga Web selects a random one.`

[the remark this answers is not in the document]
julianbrost: Random one doesn't sound very helpful though. Maybe we should say that it should always contain an English version as a fallback (and just use the name as fallback if that doesn't exist either).

pkg/rpc/rpc.go
```diff
@@ -126,21 +135,22 @@ func (r *RPC) Close() error {
...
 		if r.requestedShutdown {
-			r.logger.Infof("Plugin shutdown triggered: %s", pendingReqMsg)
+			r.logger.Info("Plugin shutdown triggered")
 		} else {
-			r.logger.Warnf("Plugin terminated unexpectedly: %s", pendingReqMsg)
+			r.logger.Warn("Plugin terminated unexpectedly")
+		}
+
+		if reqCount := len(r.pendingRequests); reqCount > 0 {
+			r.logger.Infof("Cancelling %d pending request(s)", reqCount)
```
oxzi, line 148: How are those pending requests being canceled? I'm only seeing channel operations on `r.pendigRequests` in the `*RPC.processResponses` method, which, however, exits after an error is set,.

internal/channel/plugin.go
```diff
@@ -27,60 +98,92 @@ func (c *Channel) GetInfo() (*plugin.Info, error) {
...
+
+		info, err := p.GetInfo()
+		if err != nil {
+			p.logger.Error(err)
+			p.Stop()
+			continue
+		}
+		p.Stop()
+
+		pluginInfos = append(pluginInfos, info)
```
oxzi, line 162: This might sound a bit trivial, but you might wanna log which plugins are going to be loaded. Especially for debugging the application, it would be useful to have a list of loaded plugins in the output log.

internal/channel/plugin.go
```diff
@@ -27,60 +85,94 @@ func (c *Channel) GetInfo() (*plugin.Info, error) {
...
+		pluginLogger := logger.With(zap.String("name", file.Name()))
+		p, err := NewPlugin(file.Name(), pluginLogger)
+		if err != nil {
+			pluginLogger.Error(err)
+			continue
+		}
+
+		info, err := p.GetInfo()
+		if err != nil {
+			p.logger.Error(err)
```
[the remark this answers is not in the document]
oxzi: I'm not quite sure what this comments refers to, but `s/logger/pluginLogger/`.

pkg/rpc/rpc.go
```diff
@@ -126,21 +135,22 @@ func (r *RPC) Close() error {
...
 
+// setErr sets the err encoder/decoder fails and closes errChannel
 func (r *RPC) setErr(err error) {
-	r.errMu.Lock()
-	defer r.errMu.Unlock()
-
-	if r.err == nil {
-		pendingReqMsg := fmt.Sprintf("cancelling %d pending request(s)", len(r.pendingRequests))
+	r.errOnce.Do(func() {
 		if r.requestedShutdown {
```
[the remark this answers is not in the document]
julianbrost: Aren't all changes to this file completely independent from the rest of this PR? If so, please make a separate one for these changes where we can then do other cleanup like this as well.

pkg/rpc/rpc.go
sukhwinder33445: The pending requests are actually dropped because the rpc instance can no longer be used. The channel plugin must be reinitialized.

pkg/rpc/rpc.go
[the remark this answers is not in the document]
sukhwinder33445: Most of them yes, I have also made minor changes in this PR. I'll leave this open, until there's a new PR.

internal/channel/plugin.go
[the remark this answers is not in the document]
sukhwinder33445: As discussed offline, calling `NewPlugin` will fail in all the above cases, so let's leave it as is for now.

internal/config/channel.go
julianbrost, line 1: Similarly, quite a lot messages when updating a channel config:

```
2023-11-03T15:00:49.656Z    INFO    channel New changes detected    {"id": 2, "name": "RocketChat"} 
2023-11-03T15:00:49.656Z    INFO    channel Reloading the channel plugin with the updated channel   {"id": 2, "name": "RocketChat"} 
2023-11-03T15:00:49.657Z    INFO    channel Successfully started channel plugin {"id": 2, "name": "RocketChat", "pid": 182}
2023-11-03T15:00:49.659Z    INFO    channel Plugin shutdown triggered   {"id": 2, "name": "RocketChat", "pid": 175}
2023-11-03T15:00:49.659Z    WARN    channel Stopped channel plugin  {"id": 2, "name": "RocketChat", "pid": 175}
```

"New changes detected" is really more like a debug message. And the way it's written, it's hard to tell what it's supposed to convey without looking at the source code.

That should probably only be two messages max on info level: one for restarting the channel due to a config change an maybe another one saying that the old one terminated (though not sure if that would really be relevant at info).

cmd/channel/email/main.go
```diff
@@ -80,7 +81,54 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
...
+			Placeholder: "icinga@example.com",
+		},
+		{
+			Name: "password",
+			Type: "secret",
+			Label: map[string]string{
+				"en_US": "Password",
+				"de_DE": "Passwort",
+			},
+		},
```
julianbrost, line 117: That option isn't implemented either.

julianbrost (changes requested):

pkg/plugin/plugin.go
```diff
@@ -18,9 +18,65 @@ const (
...
 
+// ConfigOption describes a config element of the channel form
+type ConfigOption struct {
+	// Element name
+	Name string `json:"name"`
+
+	// Element type:
+	//
+	//  string = text, port = number, bool = checkbox, text = textarea, option = select, options = select[multiple], secret = password
+	Type string `json:"type"`
```
julianbrost, line 29: Please explain why you changed "int" to "port" as that looks quite inconsistent compared to the others.

internal/channel/plugin.go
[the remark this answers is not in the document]
julianbrost: I've pushed some commits that show how I think the logging can be a bit more useful in this file: https://github.com/Icinga/icinga-notifications/compare/313b2d1915a1f94f4d553bf2c4eb4e5d2cd51ca8...03b53f1dde8a878b414f44f67f86fed47a63fb05

With that, it looks something like that:

Sync available plugins:
```
DEBUG   channel Updating available channel types
DEBUG   channel Starting new channel plugin process {"type": "email", "path": "/usr/libexec/icinga-notifications/channel/email"}
DEBUG   channel Successfully started channel plugin process {"type": "email", "pid": 11} 
DEBUG   channel Starting new channel plugin process {"type": "rocketchat", "path": "/usr/libexec/icinga-notifications/channel/rocketchat"}
DEBUG   channel Requesting channel plugin stop  {"type": "email", "pid": 11}
INFO    channel deliberately delaying termination by 10 seconds {"type": "email", "pid": 11}
DEBUG   channel Successfully started channel plugin process {"type": "rocketchat", "pid": 16}
DEBUG   channel Requesting channel plugin stop  {"type": "rocketchat", "pid": 16}
DEBUG   channel Channel plugin terminated   {"type": "rocketchat", "pid": 16}
INFO    channel Successfully updated 2 available channel types: email, rocketchat
```

Configured plugin startup:
```
DEBUG   channel Initializing channel plugin {"id": 1, "name": "E-Mail"}
DEBUG   channel Starting new channel plugin process {"id": 1, "name": "E-Mail", "path": "/usr/libexec/icinga-notifications/channel/email"}
DEBUG   channel Initializing channel plugin {"id": 2, "name": "RocketChat"}
DEBUG   channel Starting new channel plugin process {"id": 2, "name": "RocketChat", "path": "/usr/libexec/icinga-notifications/channel/rocketchat"}
DEBUG   channel Successfully started channel plugin process {"id": 2, "name": "RocketChat", "pid": 25}
INFO    listener    Starting listener on http://:5680
DEBUG   channel Successfully started channel plugin process {"id": 1, "name": "E-Mail", "pid": 24}
INFO    channel Successfully started channel plugin {"id": 2, "name": "RocketChat", "pid": 25}
INFO    channel Successfully started channel plugin {"id": 1, "name": "E-Mail", "pid": 24}
```

Afterwards (as the termination of the plugins from the available plugin sync happens in the background), there's now something that shows that something strange is going on (due to the code I added for testing this):
```
WARN    channel Channel plugin did not terminate after timeout, killing it  {"type": "email", "pid": 11, "timeout": "5s"}
DEBUG   channel Channel plugin terminated   {"type": "email", "pid": 11}
```

(Please don't attempt to fix the TODO in `internal/channel/plugin.go` within this PR. This will be something for a separate PR.)

pkg/plugin/plugin.go
sukhwinder33445: Johannes suggested me to do it this way. This creates an input element on the `notifications-web` that only accepts `0 - 65535`.
He said that these options should not define the form elements exactly. They are just options from which we can (somehow) generate form elements. The other idea would be to use the type `number = number` instead and two extra optional options `min` and `max`.

julianbrost: "number" with min/max sounds more flexible to me. If there was a type "port", it should probably at most be an addition to a generic number/integer field, not replace it.

Sure, there can be reasons for more specific types, the question is always where it starts becoming too much complexity for the benefits. Sure, we could define types like hostname/address, url, e-mail, ..., but do we want to specify and maintain that list?

pkg/plugin/plugin.go
```diff
@@ -18,9 +18,72 @@ const (
...
+	//			"1":   "January",
+	//			"2":  "February",
+	//		}
+	Options map[string]string `json:"options,omitempty"`
+
+	// Element's min option defines the minimum allowed number value. It can only be used for the type number.
+	Min int `json:"min,omitempty"`
+
+	// Element's max option defines the maximum allowed number value. It can only be used for the type number.
+	Max int `json:"max,omitempty"`
```
julianbrost, line 63: `omitempty` probably doesn't do what you'd expect here. For example, look at the `port` from `email`:

```json
  {
    "name": "port",
    "type": "number",
    "label": {
      "de_DE": "SMTP Port",
      "en_US": "SMTP Port"
    },
    "max": 65535
  }
```

Note that `ConfigOption{}` and `ConfigOption{Min: 0}` are the same thing. In order to distinguish `0` from no value, you'd need a type that explicitly allows this, something like `*int` (pretty annoying to initialize, as you need multiple statements or a function for that) or something like `sql.NullInt64`, but for JSON (I don't know of such a thing in the standard library).

Also, better use `int64` instead of `int` as the latter has an architecture-dependent size.

julianbrost (changes requested):

cmd/channel/email/main.go
```diff
@@ -80,7 +83,54 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
 }
 
 func (ch *Email) GetInfo() *plugin.Info {
-	return &plugin.Info{Name: "Email"}
+	minValue := types.Int{NullInt64: sql.NullInt64{Int64: 0}}
+	minValue.Valid = true
+
+	maxValue := types.Int{NullInt64: sql.NullInt64{Int64: 65535}}
+	maxValue.Valid = true
```
julianbrost, line 90: You can just set `Valid` as well within the struct initializer (and then also move it into the initialization below so you don't need the extra variables).

pkg/plugin/plugin.go
```diff
@@ -18,9 +19,72 @@ const (
...
+}
+
+// Upsert implements the contracts.Upserter interface.
+func (c *Info) Upsert() interface{} {
+	return struct {
+		Version          string          `db:"version"`
+		Author           string          `db:"author"`
+		ConfigAttributes json.RawMessage `db:"config_attrs"`
+	}{}
 }
```
julianbrost, line 88: This is missing the name.

internal/channel/channel.go
```diff
@@ -20,109 +19,165 @@ type Channel struct {
...
-	path := filepath.Join(pluginDir, c.Type)
+// newConfig helps to store the channel's updated properties
+type newConfig struct {
+	ctype  string
+	config string
+}
 
-	cmd := exec.Command(path)
+// Start initializes the channel and starts the plugin in the background
+func (c *Channel) Start(logger *zap.SugaredLogger, ctx context.Context) {
```
julianbrost, line 37: `context.Context` is typically the first function parameter.

internal/channel/plugin.go
```diff
@@ -27,60 +110,98 @@ func (c *Channel) GetInfo() (*plugin.Info, error) {
...
+		pluginInfos = append(pluginInfos, info)
+	}
+
+	if len(pluginInfos) == 0 {
+		logger.Info("No working plugin found")
+		return
+	}
+
+	stmt, _ := db.BuildUpsertStmt(&plugin.Info{})
+	_, err = db.NamedExec(stmt, pluginInfos)
```
julianbrost, line 186: Currently, the `type` column is set from the JSON reported by the plugin, but that should always be set to the file name and not even be part of the JSON. (Simply renaming the plugin executable should not introduce any inconsistencies.)

julianbrost (changes requested):

cmd/channel/email/main.go
sukhwinder33445: Note: `types.Int` is buggy and ignores the `omitempty` flag.

pkg/plugin/plugin.go
```diff
@@ -18,9 +19,68 @@ const (
...
+
+	// Element's max option defines the maximum allowed number value. It can only be used for the type number.
+	Max types.Int `json:"max,omitempty"`
+}
+
+// Info contains plugin information.
 type Info struct {
-	Name             string          `json:"display_name"`
-	ConfigAttributes json.RawMessage `json:"config_attrs"`
+	Type             string          `db:"type"`
```
julianbrost, line 69: That field should be explicitly tagged with `json:"-"` so that it isn't exposed there at all.

julianbrost (changes requested):

cmd/channel/email/main.go
julianbrost: Not sure if that's a bug of `types.Int` or just a limitation of `encoding/json`. Anyways, looks like it just writes `null` to the JSON then. Would that be a problem for the web module?

pkg/plugin/plugin.go
```diff
@@ -18,9 +19,68 @@ const (
...
+
+// TableName implements the contracts.TableNamer interface.
+func (c *Info) TableName() string {
+	return "available_channel_type"
+}
+
+// Upsert implements the contracts.Upserter interface.
+func (c *Info) Upsert() interface{} {
+	return Info{}
 }
```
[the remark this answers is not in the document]
sukhwinder33445: Oh yes, it's no longer necessary.

cmd/channel/email/main.go
sukhwinder33445: No, Web can handle it.

julianbrost (approved):

PR #109, opened 2023-10-12
icinga2.conf: urlencode service and host names

oxzi:
Similar to the mail notifications in Icinga 2 itself[^0], both the service name and the host name should be urlencoded. Otherwise, as some checks in my demo setup has shown me, the generated URL are invalid.

- Before: …/icingaweb2/icingadb/service?name=random fortune&host.name=foo
- Afterwards: …/icingaweb2/icingadb/service?name=random%20fortune&host.name=foo

Due to limitations of the Icinga 2 language and its libraries, no generic conversation of a char/rune to its char point value was possible. For this reason a (limited) lookup table based on RFC 3986 was chosen. This works until someone starts using Unicode for their names.

[^0]: https://github.com/Icinga/icinga2/blob/v2.14.0/etc/icinga2/scripts/mail-service-notification.sh#L165

### DIFF
--- a/icinga2.conf
+++ b/icinga2.conf
@@ -9,6 +9,35 @@ if (!globals.contains("IcingaNotificationsEventSourceId")) {
 	const IcingaNotificationsEventSourceId = 1
 }
 
+// urlencode a string loosely based on RFC 3986.
+//
+// Char replacement will be performed through a simple lookup table based on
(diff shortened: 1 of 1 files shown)


julianbrost (approved):
The important stuff!

PR #112, opened 2023-10-19
Icinga 2 API as Event Source

### LINKED ISSUE #89: Rework Icinga 2 event source
Currently, the Icinga 2 source is really more of a proof of concept based on notification and event commands (implemented in [`icinga2.conf`](https://github.com/Icinga/icinga-notifications/blob/main/icinga2.conf)). However, one notable limitation here is that a checkable can only have one event command, so when using it in the current state, a user can't use custom event commands in Icinga 2 anymore.

The in the long term, there should probably be a proper integration within Icinga 2, but before going version 1.0 here, we probably don't want to be tied to any Icinga 2 release cycles. The Icinga DB feature in Icinga 2 received quite some major changes within the 2.13 branch while finalizing Icinga DB 1.0 and that something I'd rather try to avoid with this project.

So it might make sense to put some work into another interim solution that's hopefully a bit nicer and less limited than the current Icinga 2 DSL snippet that generates curl command lines. The main issue is getting information about state changes when notifications are suppressed in Icinga 2, for example when the checkable is acknowledged. One idea would be to connect to the Icinga 2 API and use its [event streams](https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#event-streams), but that would probably require some extra polling of data after Icinga 2 reloads to get everything.

oxzi:
Fetch events from the Icinga 2 API. For details, consider the commit messages.

Closes #89.
refs https://github.com/Icinga/icinga-notifications-web/pull/156 (corresponding configuration interface)


internal/eventstream/api_responses.go
```diff
@@ -0,0 +1,287 @@
...
+	switch objQueriesRes.Type {
+	case "Comment":
+		objQueriesRes.Attrs = new(Comment)
+	case "Downtime":
+		objQueriesRes.Attrs = new(Downtime)
+	case "Host", "Service":
+		objQueriesRes.Attrs = new(HostServiceRuntimeAttributes)
+	default:
+		return fmt.Errorf("unsupported type %q", objQueriesRes.Type)
+	}
```
julianbrost, line 125: For the `/v1/objects` API queries you know (in contrast to the event stream) what type you're expecting in the response. So it should be possible two avoid the two pass unmarshal here as well as some type assertions elsewhere. The most invasive change this requires is that functions like `queryObjectsApi()` have to know the type, maybe a case where generics could come in handy.

julianbrost (commented):
Comments from the first pass over the code. In general, I will need to have a closer look over that event deduplication and whether we want change that in general so that it would be able to generate events for while the notifications daemon wasn't running (but those already get lost with the current hacky approach, so no downgrade in that respect).

internal/eventstream/client_es.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		client.handleEvent(ev, "Event Stream")
+	}
+	err = lineScanner.Err()
+	if err != nil {
+		return err
+	}
+
+	return nil
```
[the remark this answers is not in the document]
oxzi: golang-keyboard

julianbrost:
From what I can tell, you only try to remove duplicates but there's nothing that would prevent reordering?

Also, looks like you're using the timestamps from Icinga 2 for the events. So far, all events receive their timestamp from the notifications daemon:
https://github.com/Icinga/icinga-notifications/blob/4044b188a9f9423be532d11fc0c70b313fc0ca39/internal/listener/listener.go#L68
I think I'd keep it that way for now, after all, we can't send notifications in the past. (Maybe, as a future extension, we might want to track both source and processing timestamps and show hints that events were delayed.)

Anyways, I have a suggestion that shouldn't be too difficult to implement and would also handle events that were missed due to a notifications daemon restart:
1. Reconnect the event stream, once established start buffering events (no processing yet)
2. Query all state from the objects API and process events accordingly (as a first step, processing an event for every object, but there's the obvious improvement of using the current incident state (which will be loaded on daemon startup as part of #68) and only emitting events where there was a change), storing the last timestamp per object.
3. Start processing the buffered events from step 1 as well as new ones, ignoring those older than the timestamps from step 2.

oxzi:
@julianbrost: I did an implementation with some necessary refactoring in the three just pushed commits. Please feel free to take a look.

internal/eventstream/client.go
```diff
@@ -0,0 +1,319 @@
...
+
+		case ev := <-client.eventDispatch:
+			if client.replayPhase.Load() {
+				reconnectionBuffer = append(reconnectionBuffer, ev)
+			} else {
+				client.CallbackFn(ev)
+			}
+		}
+	}
+}
```
julianbrost, line 254: If I didn't overlook something, you seem to be buffering the events from both steps (1) and (2) mentioned in https://github.com/Icinga/icinga-notifications/pull/112#issuecomment-1780814041 here.

My idea behind the buffering was to ensure a consistent order of the events we are processing. So a simplified version of what I would want to do would be:
1. Query all state we need to know form the objects API and establish this a the current state of all objects/incidents here by emitting corresponding events.
2. After that's done, process further updates from the event stream.

Now if that request for (2) was started only after (1) finished, events may be lost in between, therefore, it has to be started before (1), so that we know we receive all events and buffer them for future use (but only those from the event stream). However, now we may also receive some events that are older than the data read in (1), but these can be removed based on the timestamps on a per-object basis.

julianbrost (changes requested):

internal/eventstream/client.go
```diff
@@ -0,0 +1,326 @@
...
+	for {
+		select {
+		case <-client.Ctx.Done():
+			client.Logger.Warnw("Closing event dispatcher as context is done", zap.Error(client.Ctx.Err()))
+			return
+
+		case <-client.replayTrigger:
+			for _, ev := range replayBuffer {
+				client.CallbackFn(ev)
+			}
```
julianbrost, line 257: As just discussed, these should not just all be replayed, as there could be events that are duplicates, or worse, states that are older than what was read from the objects API. So there should probably be some `map[object]timestamp` to remove them.

julianbrost (changes requested):

internal/eventstream/api_responses.go
```diff
@@ -0,0 +1,308 @@
...
+
+	case string:
+		checkResult.Command = []string{command}
+
+	default:
+		return fmt.Errorf("unexpected command type %T", command)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
oxzi: I should have done the right thing from the beginning and just have deleted the unnecessary code. I have just felt a bit like a cheat doing so. Now the diff is shorter :)

internal/eventstream/client.go
```diff
@@ -0,0 +1,336 @@
...
+			go func(fn func(string), objType string) {
+				fn(objType)
+				replayWg.Done()
+			}(fn, objType)
+		}
+	}
+
+	go func() {
+		startTime := time.Now()
+		replayWg.Wait()
```
julianbrost, line 304: You could also use [`errgroup`](https://pkg.go.dev/golang.org/x/sync/errgroup) here. With that, you don't have to do that WaitGroup tracking manually and as it's used in Icinga DB (and Icinga Go Library in the future), it won't even be a new dependency.

internal/eventstream/client.go
```diff
@@ -0,0 +1,326 @@
...
+				// Limit the initial Dial timeout to enable a fast retry when the Icinga 2 API is offline. Due to the
+				// need for very long-lived connections against the Event Stream API afterward, Client.Timeout would
+				// limit the whole connection, which would be fatal.
+				//
+				// Check the "Client Timeout" section of the following (slightly outdated) blog post:
+				// https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/
+				DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
+					dialer := &net.Dialer{Timeout: 3 * time.Second}
+					return dialer.DialContext(ctx, network, addr)
+				},
```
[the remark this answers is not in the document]
oxzi: Logic was changed with a concurrent timer in the `listenEventStream` method and a timeout for the Icinga 2 Objects API requests.

internal/eventstream/client.go
```diff
@@ -0,0 +1,391 @@
...
+	var (
+		eventName      string
+		eventUrlSuffix string
+		eventTags      map[string]string
+		eventExtraTags = make(map[string]string)
+	)
+
+	if service != "" {
+		eventName = host + "!" + service
+		eventUrlSuffix = "/icingadb/service?name=" + url.PathEscape(service) + "&host.name=" + url.PathEscape(host)
```
julianbrost: Shouldn't this use `url.QueryEscape()`?

internal/eventstream/client.go
```diff
@@ -0,0 +1,391 @@
...
+		serviceGroups, err := client.fetchServiceGroups(host, service)
+		if err != nil {
+			return nil, err
+		}
+		for _, serviceGroup := range serviceGroups {
+			eventExtraTags["servicegroup/"+serviceGroup] = ""
+		}
+	} else {
+		eventName = host
+		eventUrlSuffix = "/icingadb/host?name=" + url.PathEscape(host)
```
julianbrost: Shouldn't this use `url.QueryEscape()`?

julianbrost (changes requested):
See inline comments for some details. On a higher level, I'd consider changing two things:

1. Currently, states and acknowledgements are fetched completely independently. This means those could be out of sync (you might acknowledge too early or too late in relation to the state events). So this should be more consistent if there was one request to `/v1/objects/hosts` and one to `/v1/objects/services` which is used as the baseline for everything (severity, acknowledgement, whatever we might add in the future). You emit a state event for that state and if it is flagged as acknowledged, you fetch the necessary extra information and emit another event for that as well (after the state event).
2. In `eventDispatcher()`, I find the use of `replayTrigger` and `replayPhase` somewhat confusing. Maybe that would be simpler if there was one call to `eventDispatcher()` per event stream request, i.e. it would only handle one replay phase overall (but you'd probably have to be careful to only start the next one after the previous is gone for sure and won't submit further events for processing). Other ideas I'd consider: use only channels for signalling, i.e. replace `replayPhase` with one and maybe close `eventDispatcherReplay` to signal that the replay is done.

internal/eventstream/client.go
oxzi: Unfortunately `url.QueryEscape(" "))` results in `+` instead of `%20` which breaks the URL for Icinga Web.

Initially I tried using `url.QueryEscape` but had to realize that it doesn't work for this use case.

internal/eventstream/client.go
oxzi: Same as <https://github.com/Icinga/icinga-notifications/pull/112#discussion_r1381580893>.

internal/eventstream/client.go
julianbrost: `url.PathEscape()` misses other characters like `&`: https://go.dev/play/p/z4lO-ZYlSwp

From my quick look through the `net/url` code, looks like there's no function that does exactly what we would need here then. Soo... `QueryEscape()` and replace `+` with `%20`? :see_no_evil: (And maybe open an issue in Icinga Web for that.)

oxzi:
I addressed the annotated comments within the code. However, the two main bullet points are still open.
I'll request another review when I worked on those.

internal/eventstream/client.go
oxzi: Needs more SSE2: https://github.com/php/php-src/blob/php-8.2.12/ext/standard/url.c#L472

internal/eventstream/client.go
```diff
@@ -0,0 +1,415 @@
...
+
+		case ev := <-client.eventDispatcherReplay:
+			if replayFinCh == nil {
+				client.Logger.Errorw("Dispatcher received replay event outside of the replay phase", zap.Stringer("event", ev.event))
+				break
+			}
+
+			client.CallbackFn(ev.event)
+			replayCacheUpdate(ev)
+		}
```
julianbrost, line 385: Did you consider the case where `case <-replayFinCh:` and `case ev := <-client.eventDispatcherReplay:` might become ready at the same time? I think currently they can't, but only because `client.eventDispatcherReplay` is unbuffered.

`startReplayWorker()` is already creating `replayFinCh` now, I think this would be more robust if instead of a `chan struct{}`, this was also the output stream, i.e. `chan *eventMsg` which completely replaces `client.eventDispatcherReplay`. Then with a single case for both cases that uses the `ev, ok := <- ...` receive to check whether the channel was closed.

internal/eventstream/client.go
```diff
@@ -0,0 +1,415 @@
...
+				ts, ok := replayCache[eventCache{ev.SourceId, ev.Name, ev.Type}]
+				if ok && ev.Time.Before(ts) {
+					client.Logger.Debugw("Skip replaying outdated Event Stream event", zap.Stringer("event", ev),
+						zap.Time("event timestamp", ev.Time), zap.Time("cache timestamp", ts))
+					skipCounter++
+					continue
+				}
+
+				client.CallbackFn(ev)
+			}
```
julianbrost, line 360: While this loop is running (which may take a bit given that it performs database operations), no more receives from `client.eventDispatcherEventStream` take place, which could block further reads from the event stream response. So maybe move this into a goroutine or, as that would probably get annoying given that you'd then have to synchronize access on `replayCache`, restructure the case so that it processes the first event, and loops until the buffer is empty.

internal/eventstream/client.go
```diff
@@ -0,0 +1,415 @@
...
+		case <-client.replayPhaseRequest:
+			if replayFinCh != nil {
+				// There shouldn't be multiple concurrent startReplayWorker calls. However, technically this is possible,
+				// i.e. when the Icinga 2 API connection is flapping.
+				client.Logger.Error("Another replay phase request was sent while the Client is already replaying")
+				break
+			}
+
+			client.Logger.Debug("Dispatcher enters replay phase, starting caching Event Stream events")
+			replayFinCh = client.startReplayWorker()
```
julianbrost, line 346: The idea of the replay (or maybe better named catch up phase?) is to gather events than have been lost during an event stream disconnect. Currently, it looks like this is implemented so that this catching up only happens on connection loss, if the previous catch up was completed.

To get the events as reliably as we can, the requests to the objects API must be performed after the event stream API responded (i.e. started streaming any new events that occur from then on). Otherwise something like the following sequence would lose events:
1. Event stream connects.
2. Objects API is queried and all the requests are completed successfully.
3. Now imagine the notifications daemon has some kind of weird network issue while it's still processing the events generated based on the objects API where the event stream connection is reset but the database connection just hangs for a bit.
4. Event stream reconnects but doesn't start a new replay phase because the previous one didn't complete yet.

Any events between the last one read from the old event stream and the first one from the new event stream would be lost now.

julianbrost (changes requested):

internal/eventstream/client.go
oxzi: > Did you consider the case where `case <-replayFinCh:` and `case ev := <-client.eventDispatcherReplay:` might become ready at the same time? I think currently they can't, but only because `client.eventDispatcherReplay` is unbuffered.

Correct. Currently this should not be possible, as the channel is closed after all producing goroutines have finished.
https://github.com/Icinga/icinga-notifications/blob/500b2a974d5c03d6e811735d62d3db0f6a433d93/internal/eventstream/client.go#L283-L290

Furthermore, I'd like to keep those channels unbuffered as otherwise the producer cannot know if its message was really sent, even as this is currently not completely relevant.

> `startReplayWorker()` is already creating `replayFinCh` now, I think this would be more robust if instead of a `chan struct{}`, this was also the output stream, i.e. `chan *eventMsg` which completely replaces `client.eventDispatcherReplay`. Then with a single case for both cases that uses the `ev, ok := <- ...` receive to check whether the channel was closed.

I like this idea.

internal/eventstream/client.go
oxzi: Valid point. To not have to worry about huge callback delays when working on multiple cached events, I am going to implement a second independent FIFO-like queue which gets feed with events where currently `CallbackFn` is called.

internal/eventstream/client.go
julianbrost: > Furthermore, I'd like to keep those channels unbuffered as otherwise the producer cannot know if its message was really sent, even as this is currently not completely relevant.

That also wasn't intended as a suggestion to make the channels buffered. I just wanted to say that arguing why this behaves correctly requires some indirection (closing `replayFinCh` only happens after other goroutines finished, they finish only when they sent to the channel, sending to the channel requires active receiver, ...), to making this more obvious should be a good idea.

oxzi:
Rebased onto current changes within `main`.

A more useful diff between the code, generated before force-pushing, follows for the relevant changes within `internal/eventstream`.

```
$ git diff origin/icinga2-source..HEAD -- internal/eventstream/
```

```diff
diff --git a/internal/eventstream/client.go b/internal/eventstream/client.go
index 56d696b..d02231f 100644
--- a/internal/eventstream/client.go
+++ b/internal/eventstream/client.go
@@ -6,6 +6,7 @@ import (
        "crypto/x509"
        "fmt"
        "github.com/icinga/icinga-notifications/internal/config"
+       "github.com/icinga/icinga-notifications/internal/daemon"
        "github.com/icinga/icinga-notifications/internal/event"
        "github.com/icinga/icingadb/pkg/icingadb"
        "github.com/icinga/icingadb/pkg/logging"
@@ -68,7 +69,7 @@ func NewClientsFromConfig(
        logs *logging.Logging,
        db *icingadb.DB,
        runtimeConfig *config.RuntimeConfig,
-       conf *config.ConfigFile,
+       conf *daemon.ConfigFile,
 ) ([]*Client, error) {
        clients := make([]*Client, 0, len(conf.Icinga2Apis))

@@ -101,7 +102,7 @@ func NewClientsFromConfig(
                        IcingaNotificationsEventSourceId: icinga2Api.NotificationsEventSourceId,
                        IcingaWebRoot:                    conf.Icingaweb2URL,

-                       CallbackFn: makeProcessEvent(ctx, db, callbackLogger, logs, runtimeConfig, conf),
+                       CallbackFn: makeProcessEvent(ctx, db, callbackLogger, logs, runtimeConfig),
                        Ctx:        ctx,
                        Logger:     logger,
                }
diff --git a/internal/eventstream/util.go b/internal/eventstream/util.go
index b203733..596d586 100644
--- a/internal/eventstream/util.go
+++ b/internal/eventstream/util.go
@@ -23,7 +23,6 @@ func makeProcessEvent(
        logger *logging.Logger,
        logs *logging.Logging,
        runtimeConfig *config.RuntimeConfig,
-       configFile *config.ConfigFile,
 ) func(*event.Event) {
        return func(ev *event.Event) {
                obj, err := object.FromEvent(ctx, db, ev)
@@ -39,7 +38,6 @@ func makeProcessEvent(
                        obj,
                        logs.GetChildLogger("incident"),
                        runtimeConfig,
-                       configFile,
                        createIncident)
                if err != nil {
                        logger.Errorw("Failed to get current incident", zap.Error(err))
```

internal/eventstream/client.go
```diff
@@ -0,0 +1,455 @@
...
+		catchupCancel context.CancelFunc
+
+		// catchupBuffer holds Event Stream events to be replayed after the catch-up-phase has finished.
+		catchupBuffer = make([]*event.Event, 0)
+		// catchupCache maps eventCache derived from event.Events to API time to skip replaying outdated events.
+		catchupCache = make(map[eventCache]time.Time)
+
+		// dispatchQueue is a FIFO-like queue for events to be dispatched to the callback function without having to
+		// wait for the callback to finish, which, as being database-bound, might take some time during bulk phases.
+		dispatchQueue = make(chan *event.Event, 1<<16)
```
julianbrost, line 327: I don't really get the idea behind this extra buffered channel. There's already `catchupBuffer` for buffering and combined with how it's used, it looks like it introduces some 65536 events + how many events can be handled within a minute limit.

Can you please explain what this is intended to do? I'm not really sure if it's supposed to smooth out some database latency, or take up the entire contents of `catchupBuffer` at times, or both.

oxzi: All events - fetched over the Event Stream API during normal operation, fetched from the Icinga 2 API during the catch-up-phase, and replayed Event Stream API events when leaving the catch-up-phase - are being sent to the `dispatchQueue` for processing. This is done through the `dispatchEvent` helper, which also checks the `context.Context` and has an additional huge timeout if there is very much to do or something went really wrong.

<details>
<summary>Code snippets writing into dispatchQueue</summary>

https://github.com/Icinga/icinga-notifications/blob/9ccd2c28d7b7240dd47e26118d93ba532dc4aa93/internal/eventstream/client.go#L344-L359
https://github.com/Icinga/icinga-notifications/blob/9ccd2c28d7b7240dd47e26118d93ba532dc4aa93/internal/eventstream/client.go#L391-L397
https://github.com/Icinga/icinga-notifications/blob/9ccd2c28d7b7240dd47e26118d93ba532dc4aa93/internal/eventstream/client.go#L401-L414
https://github.com/Icinga/icinga-notifications/blob/9ccd2c28d7b7240dd47e26118d93ba532dc4aa93/internal/eventstream/client.go#L422-L430
</details>

In parallel, another goroutine consumes events from the buffered `dispatchQueue` goroutine and sends them sequentially to the defined callback function.

<details>
<summary>Consuming goroutine around dispatchQueue</summary>

https://github.com/Icinga/icinga-notifications/blob/9ccd2c28d7b7240dd47e26118d93ba532dc4aa93/internal/eventstream/client.go#L330-L342
</details>

The `catchupBuffer`, on the other hand, is only used to delay Event Stream events during the catch-up-phase and gets filled until the phase ends and is only then being consumed.

As the `dispatchQueue` covers all events (see above) and is buffered with an huge buffer of up to 2^16 possible events, all three possible producers should not block, or under an immense load only for a very short time, when enqueuing an event to be delivered to the callback.

In contrast, the prior and unbuffered implementation which just sent events directly to the callback function always blocked for a short amount of time. For just a few elements, this delay was too short to be relevant. However, in my test instance with round about 4k initial state changes being sent to a Rocket.Chat, this delay summed up to over a minute due to database queries and interacting with the chat.

By the way, I made those change in regard to your earlier comment, <https://github.com/Icinga/icinga-notifications/pull/112#discussion_r1385106470>.

I hope it could made my motivations clear. Please feel free to argue on them if I'm mistaking something or if something is still unclear.

julianbrost: > By the way, I made those change in regard to your earlier comment, [#112 (comment)](https://github.com/Icinga/icinga-notifications/pull/112#discussion_r1385106470).

The first idea I had in mind was to rewrite the following case (code from the PR when that comment was written)

https://github.com/Icinga/icinga-notifications/blob/500b2a974d5c03d6e811735d62d3db0f6a433d93/internal/eventstream/client.go#L348-L366

something like this (reduced to the parts relevant for what I'm trying to say, so missing check of `replayCache`, logging, reset of other variable):

```go
case <-replayFinCh:
    // process a single event
    client.CallbackFn(replayBuffer[0])
    replayBuffer = replayBuffer[1:]

    if len(replayBuffer) == 0 {
        // no more events left, disable this case
        replayFinCh = nil
    } else {
        // nothing explicit to do here, implicitly continue the loop,
        // allowing the other cases to add another event to replayBuffer
    }
```

That on its own probably isn't perfect. For one, it heavily relies on the behavior when multiple cases are ready and still blocks the loop, but limited to the processing time of a single event now. While there's room for improvement, it should convey the idea.

With the current PR, there are these extra 65536 and 1 minute limits, the need to clear an extra buffer, and there are cases where it's ignored that events are dismissed. For now, this looks like complexity where I don't see what it's for, as there should be no need for a second stage of buffering and this all seems to exist just for managing that second buffer.

oxzi: As I felt unsure, I just "verified" the fairness of the current implementation within the debugger. It behaves as the [Go Language Specification says about `select`](https://go.dev/ref/spec#Select_statements) that "[i]f one or more of the communications can proceed, a single one that can proceed is chosen via a uniform pseudo-random selection". Just to be sure, I also wrote a [small mockup](https://go.dev/play/p/qhNf6C_LM1p) which shows - please note that the runs are not deterministic, as stated in the specs above - that the other producer may also be selected when working on the closed producer.

oxzi:
Rebased on `main` and thought I found a bug within my code, which, however, was one in my testing setup..

internal/eventstream/client.go
```diff
@@ -0,0 +1,421 @@
...
+	)
+
+	// catchupCacheUpdate updates the catchupCache if this eventMsg seems to be the latest of its kind.
+	catchupCacheUpdate := func(ev *eventMsg) {
+		evc := eventCache{ev.event.SourceId, ev.event.Name, ev.event.Type}
+		ts, ok := catchupCache[evc]
+		if !ok || ev.apiTime.After(ts) {
+			catchupCache[evc] = ev.apiTime
+		}
+	}
```
julianbrost, line 333: `eventCache.Type` is something like state/ack here, not host/service, isn't it? So given that both use the same value for `apiTime`, it shouldn't be necessary to even include this in the cache key.

https://github.com/Icinga/icinga-notifications/blob/a61c6efbebac100a97c44f198c272f2ad6944a87/internal/eventstream/client_api.go#L231-L232

https://github.com/Icinga/icinga-notifications/blob/a61c6efbebac100a97c44f198c272f2ad6944a87/internal/eventstream/client_api.go#L258-L259

julianbrost (changes requested):

internal/eventstream/client_api.go
```diff
@@ -0,0 +1,421 @@
...
+		typeStateChange,
+		typeAcknowledgementSet,
+		// typeAcknowledgementCleared,
+		// typeCommentAdded,
+		// typeCommentRemoved,
+		// typeDowntimeAdded,
+		// typeDowntimeRemoved,
+		// typeDowntimeStarted,
+		// typeDowntimeTriggered,
+	})
```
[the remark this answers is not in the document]
oxzi: Yes, in theory I would agree. However, in this very case creating a local sub-context is kind of a problem, as this context is only used to allow aborting the web request after a timeout. Thus, the local `context.CancelFunc` is only called in case of an early error, as the Event Stream connection is bound to it.

When leaving the `connectEventStream` method with an established connection, the context is still in use. But in this case the `context.CancelFunc` was not called and would go out of scope.

Against this, the `context.CancelFunc` needs still to be returned - as in the current implementation -, resulting in an even more oddly looking pattern with then two `context.CancelFunc`s to be called. Like so:
```go
	ctx, cancel := context.WithCancel(client.Ctx)
	defer cancel()

	response, rseponseCancel, err := client.connectEventStream(ctx, []string{ … })
	if err != nil { … }
	defer func() {
		cancel()

		_, _ = io.Copy(io.Discard, response.Body)
		_ = response.Body.Close()
	}()
```

As an alternative, one could rely on our "domain knowledge" and just accept that the method-local `context.CancelFunc` will not be called directly, but as being derived from a context which will eventually be canceled. This, however, will look even more unconventional, might result in future bugs when this knowledge - documented or not - gets forgotten, and will trigger each linter. In fact, as I have just played with this idea, even GoLand's quite conservative linter lighted up this case.

To my knowledge, there are only two reasonable ways to terminate a web request early: by timeouts, which is a bit tricky for long-lived connections and which was already discussed <https://github.com/Icinga/icinga-notifications/pull/112#discussion_r1376514352>, and by a `context.Context`.

That's why I would stay with the current architecture for the moment. However, I'm open to other suggestions.

internal/eventstream/client.go
oxzi: Yes, `eventCache.Name` is the host or service name (either `$host` for host events or `${host}!${service}` for service events) and `eventCache.Type` the`Event.Type`.
https://github.com/Icinga/icinga-notifications/blob/3b9bef52be6bfba808460da34a56590378c0ddfa/internal/event/event.go#L23
https://github.com/Icinga/icinga-notifications/blob/3b9bef52be6bfba808460da34a56590378c0ddfa/internal/event/event.go#L31-L35

However, when dropping the `.Type` field, a state change would not be distinguishable from an acknowledgement, right?

internal/eventstream/util.go
```diff
@@ -0,0 +1,83 @@
...
+//
+// This function contains glue code similar to those from Listener.ProcessEvent to check for incidents for the Event
+// and, if existent, call *Incident.ProcessEvent on this incident.
+func makeProcessEvent(
+	ctx context.Context,
+	db *icingadb.DB,
+	logger *logging.Logger,
+	logs *logging.Logging,
+	runtimeConfig *config.RuntimeConfig,
+) func(*event.Event) {
```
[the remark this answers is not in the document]
oxzi: I am not quite sure if I got this comment right, but I understood this as the opportunity to refactor the common setup code from  `Listener.ProcessEvent` and this `makeProcessEvent` together. Thus, I have created `incident.ProcessEvent` in the recent commit d851d78dabb671c658b5d2b84ab2b9bde09c97c8.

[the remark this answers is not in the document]
julianbrost: Yes, exactly that.

internal/eventstream/client_api.go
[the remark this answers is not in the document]
julianbrost: > This, however, will look even more unconventional, might result in future bugs when this knowledge - documented or not - gets forgotten, and will trigger each linter. In fact, as I have just played with this idea, even GoLand's quite conservative linter lighted up this case.

Okay, I didn't consider that part. (Would be kind of nice if there was some way to cancel the `CancelFunc`, so that the child context becomes pretty much the same as its parent. But I don't know of any such way except implementing a new `context.Context`, so probably not.)

> However, I'm open to other suggestions.

I think my suggestion was mainly caused by the defer function having to deal with HTTP and context internals of `connectEventStream()` about which we don't really care here. So another quick idea: maybe just return an `io.ReadCloser` that just does all the right cleanup on both read errors and close. (Needs implementing an extra type though, so not fully sure how much nicer this will be in the end.)

internal/eventstream/client.go
julianbrost: > However, when dropping the `.Type` field, a state change would not be distinguishable from an acknowledgement, right?

Yes, but that's actually a good thing. Currently, there are two scenarios:
1. The checkable was acknowledged, so `catchupCache["$name", "state"]` and `catchupCache["$name", "acknowledgement"]` store the same value, so if it was just `catchupCache["$name"]`, there would be no difference.
2. The checkable was not acknowledged. Now `catchupCache["$name", "acknowledgement"]` won't exist and all acknowledgements from the event stream will be processed, no matter if they have a timestamp before `catchupCache["$name", "state"]` or not. If there's a buffered acknowledgement from the event stream with a timestamp earlier than that, this would mean that this acknowledgment got already dismissed when `/v1/objects` API was queried. So by the time it's processed, it's already obsolete and the processing would happen out of order. If it was just a common value for all event types here, that wouldn't happen.

internal/eventstream/client_api.go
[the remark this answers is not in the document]
oxzi: I liked the `io.ReadCloser` idea and went with it in c305b9e0ce548109396fe681bfd3506dc2ea47ef.

yhabteab:
Here are some notes I noticed during my tests!

1. I think it's just fine to use the same logger for callback and client logging without any suffixes, i.e. just use `eventstream` as the child logger name.
2. When you restart the Icinga Notifications daemon or Icinga 2, all superfluous state changes should be silently dropped. Since all active incidents are loaded before the event stream api is listened to, you should be able to easily figure out the actual severity of the affected incident.
   ```bash
   2023-11-28T15:32:22.155Z	WARN	eventstream-callback-1	Cannot process event	{"error": "Ignoring superfluous \"err\" state event from source 1"}
   2023-11-28T15:32:22.160106503Z 2023-11-28T15:32:22.159Z	WARN	eventstream-callback-1	Cannot process event	{"error": "ignoring superfluous OK state event from source 1"}
   2023-11-28T15:32:22.165573962Z 2023-11-28T15:32:22.165Z	WARN	eventstream-callback-1	Cannot process event	{"error": "ignoring superfluous OK state event from source 1"}
   2023-11-28T15:32:22.174958670Z 2023-11-28T15:32:22.174Z	WARN	eventstream-callback-1	Cannot process event	{"error": "ignoring superfluous OK state event from source 1"}
   2023-11-28T15:32:22.187196795Z 2023-11-28T15:32:22.186Z	WARN	eventstream-callback-1	Cannot process event	{"error": "ignoring superfluous OK state event from source 1"}
   ```
3. When stopping Icinga 2 (I think the maximum retry backoff is `3s`? Didn't verify it!), you see a log message every 3s that the connection to Icinga 2 cannot be established. You may want to log only every `1m` or `5m` instead (just a random number).
   ```
   2023-11-28T16:01:32.254Z	WARN	eventstream-1	Establishing an Event Stream API connection failed, will be retried	{"error": "Post \"https://icinga2:5665/v1/events\": dial tcp: lookup icinga2 on 127.0.0.11:53: no such host"}
   2023-11-28T16:01:35.258409798Z 2023-11-28T16:01:35.258Z	WARN	eventstream-1	Establishing an Event Stream API connection failed, will be retried	{"error": "Post \"https://icinga2:5665/v1/events\": dial tcp: lookup icinga2 on 127.0.0.11:53: no such host"}
   2023-11-28T16:01:38.261707049Z 2023-11-28T16:01:38.261Z	WARN	eventstream-1	Establishing an Event Stream API connection failed, will be retried	{"error": "Post \"https://icinga2:5665/v1/events\": dial tcp: lookup icinga2 on 127.0.0.11:53: no such host"}
    ```

oxzi:
Thanks for your input, @yhabteab! 

I have reworked the incident handling by creating a custom error, allowing to hide (or debug log) the superfluous state changes from Event Stream API Events. While doing so, I removed the custom logger, which, iirc, was a leftover from something else.

The reconnection backoff was changed to an exponential backoff topped at three minutes, instead of a linear backoff being topped at three seconds. During my tests, I liked the short interval, especially as being a bit impatient. But, when running in production, it would be unwise to flood the logs with an unnecessary amount of reconnection attempts.

yhabteab:
The event stream channels also seem to block on shutdown. Currently you would have to wait at worst `3` minutes before the daemon is successfully shut down. I think you need some mechanics to immediately react upon the syscall signals.

oxzi:
Thanks for the feedback, @yhabteab. Please feel free to take another look :)

julianbrost:
With #132 merged, please update this PR so that it also gets its config from the database.

oxzi:
I have moved the source's configuration to the database in 245ad9f239c5c4e51de593119e93d71f3911569c as prepared in #132.

julianbrost (commented):
A few high-level observations for now, I haven't looked at the new code so far.

README.md
```diff
@@ -23,10 +23,27 @@ It is required that you have created a new database and imported the [schema](sc
...
+  Therefore, you need to copy the [Icinga 2 config](icinga2.conf) to `/etc/icinga2/features-enabled` on your master node(s) and restart the Icinga 2 service.
+  At the top of this file, you will find multiple configurations options that can be set in `/etc/icinga2/constants.conf`.
+  There are also Icinga2 `EventCommand` definitions in this file that will automatically match all your **checkables**, which may not work properly if the configuration already uses event commands for something else.
+
 Additionally, it also requires you to manually insert items into the **source** table before starting the daemon.
 ```sql
+-- Allow incoming connections being pushed to icinga-notification's web server:
 INSERT INTO source (id, type, name, listener_password_hash)
-VALUES (1, 'icinga2', 'Icinga 2', '$2y$10$QU8bJ7cpW1SmoVQ/RndX5O2J5L1PJF7NZ2dlIW7Rv3zUEcbUFg3z2');
+  VALUES (1, 'web-api', 'Icinga 2', '$2y$10$QU8bJ7cpW1SmoVQ/RndX5O2J5L1PJF7NZ2dlIW7Rv3zUEcbUFg3z2');
```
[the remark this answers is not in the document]
oxzi: `web-api` was just used because the current logic says either `icinga2` or something else.

internal/eventstream/api_responses.go
```diff
@@ -0,0 +1,261 @@
...
+//   - Acknowledgement type is 0 = NONE, 1 = NORMAL, 2 = STICKY.
+//
+// https://icinga.com/docs/icinga-2/latest/doc/09-object-types/#host
+// https://icinga.com/docs/icinga-2/latest/doc/09-object-types/#service
+type HostServiceRuntimeAttributes struct {
+	Name                      string      `json:"name"`
+	Host                      string      `json:"host_name,omitempty"`
+	Groups                    []string    `json:"groups"`
+	State                     int         `json:"state"`
+	StateType                 int         `json:"state_type"`
```
julianbrost, line 85: This and other structs in this file have `StateType` attributes but they seem to be used nowhere. So far, only hard states where processed but the current implementation processes any state.

internal/eventstream/api_responses.go
```diff
@@ -0,0 +1,261 @@
+package eventstream
```
julianbrost, line 1: I'd rename this package (and directory). For this project, it's more relevant that we're interacting with Icinga 2 to make it work as a source, not that this is done using its event stream API. Maybe just go for `icinga2`, alternative I have in mind would be 'icinga2source` (looks cumbersome) or maybe something like `internal/sources/icinga2/` (looks like we'd want to implement more in this style, which I don't think we want).

julianbrost (changes requested):
I gave this a basic test where I noticed a few things so far (see inline comments), but all in all it looks like we're getting close to the finish line. Don't be scared by the sheer size of the review, quite a few comments are just on comment texts for example.

internal/incident/incidents.go
```diff
@@ -132,3 +136,48 @@ func GetCurrentIncidents() map[int64]*Incident {
...
+		return fmt.Errorf("cannot get current incident for %q: %w", obj.DisplayName(), err)
+	}
+
+	if currentIncident == nil {
+		switch {
+		case ev.Type == event.TypeAcknowledgement:
+			return fmt.Errorf("%q does not have an active incident, ignoring acknowledgement event from source %d",
+				obj.DisplayName(), ev.SourceId)
+		case ev.Severity != event.SeverityOK:
+			return fmt.Errorf("cannot process event with a non-OK state without a known incident")
```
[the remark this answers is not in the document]
oxzi: Changed.

I just disliked raising panics for things which are imo not a total catastrophe, as this goes on the cost resilience. However, in this case this would indicate either an rogue or faulty Icinga 2 daemon.

internal/incident/incidents.go
```diff
@@ -132,3 +136,48 @@ func GetCurrentIncidents() map[int64]*Incident {
...
+		default:
+			return fmt.Errorf("%w: ok state event from source %d", ErrSuperfluousStateChange, ev.SourceId)
```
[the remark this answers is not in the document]
oxzi: A similar example is at least shown in the Go blog post [Working with Errors in Go 1.13](https://go.dev/blog/go1.13-errors) within the _Wrapping errors with %w_ section.

I kind of like this approach as it is short on code, but does not rely on some magic effects. Otherwise, implementing a custom error type feels like creating boilerplate code.

internal/eventstream/api_responses.go
```diff
@@ -0,0 +1,261 @@
...
+	Host    string `json:"host_name"`
+	Service string `json:"service_name"`
+	Author  string `json:"author"`
+	Comment string `json:"comment"`
+}
+
+// HostServiceRuntimeAttributes are common attributes of both Host and Service objects.
+//
+// When catching up potentially missed changes, the following fields seems to be the relevant ones which, fortunately,
+// are identical for Icinga 2 Host and Service objects.
```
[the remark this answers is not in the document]
oxzi: Well.. there were at least some uncertainties when I wrote it first back then.

internal/incident/incidents.go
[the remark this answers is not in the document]
julianbrost: > However, in this case this would indicate either an rogue or faulty Icinga 2 daemon.

Can this happen? When I wrote that code, this was like "this should never happen, if `ev.Severity != event.SeverityOK`, there should be an incident (created if it didn't exist so far), add this check with a more helpful message than the nil dereference that would probably happen right afterwards".

internal/eventstream/client.go
```diff
@@ -0,0 +1,336 @@
...
+
+			// The channel is closed, replay cache and eventually switch modes
+			if len(catchupBuffer) > 0 {
+				// To not block the select and all channels too long, only one event will be processed per iteration.
+				ev := catchupBuffer[0]
+				catchupBuffer = catchupBuffer[1:]
+
+				ts, ok := catchupCache[ev.Name]
+				if !ok {
+					client.Logger.Errorw("Event to be replayed is not in cache", zap.Stringer("event", ev))
```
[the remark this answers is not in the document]
oxzi: Good catch. I have only thought about a somehow static Icinga 2 configuration, but yes, you are totally right.

internal/eventstream/client_api.go
```diff
@@ -0,0 +1,438 @@
...
+	}
+
+	return res.Body, nil
+}
+
+// queryObjectsApiDirect performs a direct resp. "fast" API query against an object, optionally identified by its name.
+func (client *Client) queryObjectsApiDirect(ctx context.Context, objType, objName string) (io.ReadCloser, error) {
+	return client.queryObjectsApi(
+		ctx,
+		[]string{"/v1/objects/", objType + "s/", rawurlencode(objName)},
```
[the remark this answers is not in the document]
oxzi: Tbh, I am not quite sure why I also substituted `url.PathEscape` to `rawurlencode` at this place. I went down a rabbit hole inspecting how Icinga 2 (or more likely boost) was doing things and somewhat certain that `url.PathEscape` should do the right thing, even for very strange host or service names.

internal/eventstream/client_api.go
```diff
@@ -0,0 +1,438 @@
...
+		ctx,
+		[]string{"/v1/objects/", objType + "s"},
+		http.MethodPost,
+		bytes.NewReader(reqBody),
+		map[string]string{
+			"Accept":                 "application/json",
+			"Content-Type":           "application/json",
+			"X-Http-Method-Override": "GET",
+		})
+}
```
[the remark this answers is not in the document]
oxzi: This has sent me deep down a rabbit hole…

Yes, in theory or when directly speaking to the Icinga 2 API, this will work - as I have also verified this. However, when some industry grade firewall/proxy/middleware thingy appears, this will might stop working as I have tried to explain in <https://github.com/Icinga/icinga2/pull/9958>.
Thus, I would like to leave the code as it is, as otherwise strange bugs might appear in the future.

internal/eventstream/api_responses.go
oxzi: @julianbrost: I did some small refactoring to reduce the number of magic integer constant values mirrored from the Icinga 2 API and introduced checks to only process HARD state changes.

julianbrost (changes requested):

internal/icinga2/api_responses.go
```diff
@@ -0,0 +1,290 @@
...
+	StateTypeSoft = 0
+	StateTypeHard = 1
+)
+
+// Comment represents the Icinga 2 API Comment object.
+//
+// NOTE:
+//   - An empty Service field indicates a host comment.
+//   - The optional EntryType should be User = EntryTypeUser, Downtime = EntryTypeDowntime,
+//     Flapping = EntryTypeFlapping, Acknowledgement = EntryTypeAcknowledgement (EntryType* consts)
```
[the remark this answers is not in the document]
oxzi: > Also, why is EntryType optional?

It is described as an optional type in the Icinga 2 Object Types API, <https://icinga.com/docs/icinga-2/latest/doc/09-object-types/#objecttype-comment>.

internal/icinga2/api_responses_test.go
```diff
@@ -0,0 +1,524 @@
...
+			expected: &StateChange{
+				Timestamp: UnixFloat(time.UnixMicro(1697188278203504)),
+				Host:      "dummy-158",
+				State:     StateHostDown,
+				StateType: StateTypeSoft,
+				CheckResult: CheckResult{
+					ExitStatus:     2,
+					Output:         "If two people love each other, there can be no happy end to it.\n\t\t-- Ernest Hemingway",
+					State:          2, // check returned an invalid state for a host check
```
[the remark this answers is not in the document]
oxzi: Thanks a lot for the clarification!

julianbrost:
The following just happened during testing when I restarted Icinga 2:

Starts reconnecting as expected:
```
2024-01-09T15:29:10.789Z	ERROR	icinga2	Event Stream processing was closed	{"source-id": 1}
2024-01-09T15:29:10.789Z	DEBUG	icinga2	Try to establish an Event Stream API connection	{"source-id": 1}
2024-01-09T15:29:10.792Z	WARN	icinga2	Establishing an Event Stream API connection failed, will be retried	{"source-id": 1, "error": "Post \"https://master-1:5665/v1/events\": dial tcp 172.31.0.22:5665: connect: connection refused"}
2024-01-09T15:29:10.792Z	DEBUG	icinga2	Try to establish an Event Stream API connection	{"source-id": 1}
2024-01-09T15:29:10.793Z	WARN	icinga2	Establishing an Event Stream API connection failed, will be retried	{"source-id": 1, "error": "Post \"https://master-1:5665/v1/events\": dial tcp 172.31.0.22:5665: connect: connection refused"}
2024-01-09T15:29:10.793Z	DEBUG	icinga2	Try to establish an Event Stream API connection	{"source-id": 1}
2024-01-09T15:29:10.793Z	WARN	icinga2	Establishing an Event Stream API connection failed, will be retried	{"source-id": 1, "error": "Post \"https://master-1:5665/v1/events\": dial tcp 172.31.0.22:5665: connect: connection refused"}
2024-01-09T15:29:10.793Z	DEBUG	icinga2	Try to establish an Event Stream API connection	{"source-id": 1}
2024-01-09T15:29:10.794Z	WARN	icinga2	Establishing an Event Stream API connection failed, will be retried	{"source-id": 1, "error": "Post \"https://master-1:5665/v1/events\": dial tcp 172.31.0.22:5665: connect: connection refused"}
2024-01-09T15:29:10.794Z	DEBUG	icinga2	Try to establish an Event Stream API connection	{"source-id": 1}
2024-01-09T15:29:10.794Z	WARN	icinga2	Establishing an Event Stream API connection failed, will be retried	{"source-id": 1, "error": "Post \"https://master-1:5665/v1/events\": dial tcp 172.31.0.22:5665: connect: connection refused"}
2024-01-09T15:29:11.795Z	DEBUG	icinga2	Try to establish an Event Stream API connection	{"source-id": 1}
2024-01-09T15:29:11.796Z	WARN	icinga2	Establishing an Event Stream API connection failed, will be retried	{"source-id": 1, "error": "Post \"https://master-1:5665/v1/events\": dial tcp 172.31.0.22:5665: connect: connection refused"}
2024-01-09T15:29:14.796Z	DEBUG	icinga2	Try to establish an Event Stream API connection	{"source-id": 1}
2024-01-09T15:29:14.911Z	INFO	icinga2	Start listening on Icinga 2 Event Stream	{"source-id": 1}
2024-01-09T15:29:14.911Z	INFO	icinga2	Worker enters catch-up-phase, start caching up on Event Stream events	{"source-id": 1}
```

However, the WARN messages happen quite rapidly, maybe it would make sense to only log the first few attempts only at DEBUG so that when a disconnect happens, you either get a reconnect, or, if that doesn't happen within a few seconds, an error message (if there's a few seconds between later attempts, it's more reasonable to log that again).

Also, shouldn't there be at least 100ms between attempts?

Then starts catching up emitting a bunch of events like the following  (to be expected):
```
2024-01-09T15:29:15.595Z	DEBUG	icinga2	Stopped processing event with superfluous state change	{"source-id": 1, "event": "[time=2024-01-09 15:29:15.577166884 +0000 UTC m=+2145.295829142 type=\"state\" severity=ok]", "error": "ignoring superfluous state change: ok state event from source 1"}
2024-01-09T15:29:15.759Z	DEBUG	icinga2	Skipping SOFT event, icinga2.HostServiceRuntimeAttributes{Name:"host_with_disabled_checks", Host:"", Groups:[]string{}, State:1, StateType:0, LastCheckResult:icinga2.CheckResult{ExitStatus:0, Output:"", State:0, ExecutionStart:icinga2.UnixFloat{wall:0x0, ext:0, loc:(*time.Location)(nil)}, ExecutionEnd:icinga2.UnixFloat{wall:0x0, ext:0, loc:(*time.Location)(nil)}}, LastStateChange:icinga2.UnixFloat{wall:0x0, ext:62135596800, loc:(*time.Location)(0xc10200)}, DowntimeDepth:0, Acknowledgement:0, AcknowledgementLastChange:icinga2.UnixFloat{wall:0x0, ext:62135596800, loc:(*time.Location)(0xc10200)}}	{"source-id": 1}
```

And then the following happened:
```
2024-01-09T15:29:17.867Z	DEBUG	icinga2	Querying API emitted events	{"source-id": 1, "object type": "host", "state changes": 45, "acknowledgements": 0}
2024-01-09T15:29:17.895Z	DEBUG	icinga2	Stopped processing event with superfluous state change	{"source-id": 1, "event": "[time=2024-01-09 15:29:17.867313181 +0000 UTC m=+2147.585975451 type=\"state\" severity=crit]", "error": "ignoring superfluous state change: crit state event from source 1"}
2024-01-09T15:29:17.913Z	ERROR	icinga2	Catch-up-phase event worker failed	{"source-id": 1, "object type": "service", "error": "Get \"https://master-1:5665/v1/objects/services\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)"}
2024-01-09T15:29:17.914Z	ERROR	icinga2	Catching up the API failed	{"source-id": 1, "error": "Get \"https://master-1:5665/v1/objects/services\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)", "duration": "3.002244697s"}
2024-01-09T15:29:17.914Z	INFO	icinga2	Worker leaves catch-up-phase, returning to normal operation	{"source-id": 1}
```

So looks like some timeout might be a bit too restrictive and also doesn't sound like it should be "returning to normal operation" in that situation.

oxzi:
> However, the WARN messages happen quite rapidly, maybe it would make sense to only log the first few attempts only at DEBUG so that when a disconnect happens, you either get a reconnect, or, if that doesn't happen within a few seconds, an error message (if there's a few seconds between later attempts, it's more reasonable to log that again).

I increased the rate limiting times to a more reasonable range. This will now result in less warning messages. However, I was not convinced to reduce the first failures to a debug message, as they definitely represent a failure which imho should not be overlooked.

> ```
> 2024-01-09T15:29:17.867Z	DEBUG	icinga2	Querying API emitted events	{"source-id": 1, "object type": "host", "state changes": 45, "acknowledgements": 0}
> 2024-01-09T15:29:17.895Z	DEBUG	icinga2	Stopped processing event with superfluous state change	{"source-id": 1, "event": "[time=2024-01-09 15:29:17.867313181 +0000 UTC m=+2147.585975451 type=\"state\" severity=crit]", "error": "ignoring superfluous state change: crit state event from source 1"}
> 2024-01-09T15:29:17.913Z	ERROR	icinga2	Catch-up-phase event worker failed	{"source-id": 1, "object type": "service", "error": "Get \"https://master-1:5665/v1/objects/services\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)"}
> 2024-01-09T15:29:17.914Z	ERROR	icinga2	Catching up the API failed	{"source-id": 1, "error": "Get \"https://master-1:5665/v1/objects/services\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)", "duration": "3.002244697s"}
> 2024-01-09T15:29:17.914Z	INFO	icinga2	Worker leaves catch-up-phase, returning to normal operation	{"source-id": 1}
> ```
> 
> So looks like some timeout might be a bit too restrictive and also doesn't sound like it should be "returning to normal operation" in that situation.

I reworked the catch-up-phase logic to allow propagating back an error. This will lead to another catch-up-phase attempt.

Please feel free to take another look. Thanks.

julianbrost:
After being asked something about the config, I noticed that it would probably be a good idea to have a separate config attribute for the Icinga 2 endpoint name, i.e. the name expected in the certificate. Given that both are configured separately in Icinga 2 (and you have to specify both, one isn't implicitly used for the other), it's quite possible to have configs like this which wouldn't be possible with the current implementation here without disabling certificate validation altogether:

```
object Endpoint "master" {
  host = "icinga.example.com"
  port = 5665
}
```

oxzi:
@julianbrost: A distinct Common Name was implemented in 24a484370dd372d6f657f92a2c554ec0965228bd, with additional information regarding adjusting one's testing environment in the commit message.

schema/pgsql/schema.sql
```diff
@@ -124,21 +124,34 @@ CREATE TABLE schedule_member (
...
+    -- icinga2_common_name requires Icinga 2's certificate to hold this Common Name if not NULL. This allows using a
+    -- differing Common Name - maybe an Icinga 2 Endpoint object name - from the FQDN within icinga2_base_url.
+    icinga2_common_name text,
+    icinga2_insecure_tls boolenum NOT NULL DEFAULT 'n',
 
     -- The hash is a PHP password_hash with PASSWORD_DEFAULT algorithm, defaulting to bcrypt. This check roughly ensures
     -- that listener_password_hash can only be populated with bcrypt hashes.
     -- https://icinga.com/docs/icinga-web/latest/doc/20-Advanced-Topics/#manual-user-creation-for-database-authentication-backend
-    CHECK (listener_password_hash LIKE '$2y$%'),
+    CHECK (listener_password_hash IS NULL OR listener_password_hash LIKE '$2y$%'),
```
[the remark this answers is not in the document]
oxzi: This was introduced in #132. Please refer to this PR and the comment above this `CHECK`.

As there is no function matching exactly [PHP's generic `password_hash` function](https://www.php.net/manual/en/function.password-hash.php) in Go - especially regarding future PHP changes -, we are verifying against the current default hashing algorithm, bcrypt, in icinga-notifications.

If a new default will be introduced, e.g., Argon2id, we can make the code more dynamic to support different PHP version defaults. But for the moment, this seems useless as icingaweb2 uses `PASSWORD_DEFAULT` and, afaik, no upcoming change to the default was announced.

To make sure that no other hashing algorithm is used, this additional SQL `CHECK` was added. This results in an early failure which is easy to catch. An additional check was also introduced to the Go code just to be sure.

https://github.com/Icinga/icinga-notifications/blob/4e5716e308feb3e954456f7db888c1804ab9c0c4/internal/config/runtime.go#L170-L179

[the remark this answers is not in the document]
nilmerg: Okay. Though, is it intended to necessitate a schema change in such a case?

You handle the error, which can't occur as no such invalid value can be stored. Which makes the error handling redundant. (For a postgres backend at least)

Updating the daemon is much less hassle and I guess you need to support the *old* algorithm anyway, just like Web does (that's why the algorithm is denoted in the hash to begin with), so the check must likely go away anyway.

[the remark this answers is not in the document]
oxzi: As I perhaps did not phrase this clearly enough: In our Go code, the hash is verified and we simply cannot know what future PHP `password_hash` defaults would be. Not only PHP code uses this value.
That's why we let it fail hard if the default has changed and we failed to patch in time.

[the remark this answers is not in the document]
nilmerg: *hard*, as in no source can be added or it's password changed. Okay :sweat_smile:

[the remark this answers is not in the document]
julianbrost: Notifications Web should just use `PASSWORD_BCRYPT` instead of relying on the default when creating a password hash for this column. Anything else would not be understood by the daemon anyways.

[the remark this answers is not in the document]
oxzi: According to the [docs](https://www.php.net/manual/en/function.password-hash.php), the default was last changed for PHP 5.5, [release in 2013](https://www.php.net/ChangeLog-5.php#5.5.0) (w/o a mention in the change log, of course).

I mean, of course we could lift the SQL `CHECK` but this would result in a Go error, which might be harder to see for an end user. Dunno, what the best behavior should be.

However, this "only" affects the `listener_password_hash` and not the upcoming Icinga 2 API communication.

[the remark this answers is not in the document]
oxzi: Sorry @nilmerg, I missed your first reply as I almost instantly wrote my second answer and somehow GitHub does not shown yours (or I was blind).

> You handle the error, which can't occur as no such invalid value can be stored. Which makes the error handling redundant. (For a postgres backend at least)

An invalid value can be stored if a future PHP version changes the default, resulting in a hash created by another hashing algorithm being stored in the database.

Thus, we should maybe go with @julianbrost's suggestion and don't rely on a default, but settle for an algorithm.

As I wrote in another answer, the idea of this double error check - I wouldn't call it redundant, it has such a bad connotation - was to let it fail early so that novice users don't have to consult icinga-notification's logs.
Of course, this brings the burden to the web team regarding user feedback.. mea culpa.

> Updating the daemon is much less hassle and I guess you need to support the _old_ algorithm anyway, just like Web does (that's why the algorithm is denoted in the hash to begin with), so the check must likely go away anyway.

Why shouldn't go a daemon update hand-in-hand with a schema update?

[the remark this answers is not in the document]
nilmerg: Of course, with appropriate tooling a schema update isn't much of a deal.

I've went with using `PASSWORD_BCRYPT` now. But only because the actual advantage of `PASSWORD_DEFAULT`, to automatically migrate to a stronger algorithm, is only effectively useful in PHP's own ecosystem. Once dealing with anything outside of that, not relying on this *dynamic upgrade* is the better choice.

internal/icinga2/client.go
```diff
@@ -0,0 +1,385 @@
...
+
+		case <-client.catchupPhaseRequest:
+			if catchupEventCh != nil {
+				client.Logger.Warn("Switching to catch-up-phase was requested while already catching up, restarting phase")
+
+				// Drain the old catch-up-phase producer channel until it is closed as its context will be canceled.
+				go func(catchupEventCh chan *catchupEventMsg) {
+					for _, ok := <-catchupEventCh; ok; {
+					}
+				}(catchupEventCh)
```
julianbrost, line 298: Wait, am I missing something completely or does that loop try to receive one element and then goes into an endless loop depending on whether that worked?

`for range catchupEventCh {}` should actually drain the channel, but is it actually necessary? Is there anything that's sending into the channel that has to be unblocked by this (i.e. doesn't do so in a select together with the context cancel channel)? If not, shouldn't just cancelling the context suffice?

internal/icinga2/client.go
```diff
@@ -0,0 +1,385 @@
...
+					select {
+					case <-client.Ctx.Done():
+					case client.catchupPhaseRequest <- struct{}{}:
+					}
+				}()
+			} else {
+				client.Logger.Info("Worker leaves catch-up-phase, returning to normal operation")
+			}
+
+			catchupReset()
```
julianbrost, line 351: If I understand the code correctly, if an error occurs, it will still continue processing `catchupEventCh` until it's closed, then, process all buffered events from the event stream and then handle the error. Did I get this correctly and if so, why is it done this way? I think I'd have treated all errors pretty much the same: stop everything and start from the beginning, like it should also happen if Icinga 2 was restarted.

julianbrost (changes requested):

internal/icinga2/client.go
oxzi: You are totally right. If I have to guess how this happened, I would say the `_, ok := <- catchupEventCh` part was prior within a `for`/`select` and messy refactoring resulted in this.

But even after reading both the git log as well the code, I don't see any real reason to drain the channel, as every writing channel access is being protected by a `<-ctx.Done():` check.

internal/icinga2/client.go
oxzi: Currently, the only case when an error will be returned to the `catchupEventCh` is after the `startCatchupWorker`'s `errgroup` has finished with an error which is not a `context.Canceled`. Directly following, the `catchupEventCh` will be closed.

https://github.com/Icinga/icinga-notifications/blob/d6c4d36f07a3a27e6bbd62c522d4aef3de6dd2ad/internal/icinga2/client.go#L228-L238

Thus, next time the `case catchupMsg, ok := <-catchupEventCh:` will be entered, it is because `catchupEventCh` is closed and all `catchupBuffer` events will be processed. Please note, those events are not from the catch-up phase, but received from the Event Stream API in the meantime.

However, you have a point that there is no use in replaying those when the catch-up failed. By restarting the catch-up phase in case of an error, those "new" events will be queried from the Icinga 2 API again.

I am going to rework this. You might wanna ignore my response, it's mostly here to collect my thoughts.

internal/icinga2/client.go
```diff
@@ -0,0 +1,388 @@
...
+
+			// Abort and restart the catch-up-phase when receiving an error.
+			if ok && catchupMsg.error != nil {
+				client.Logger.Warnw("Worker leaves catch-up-phase with an error, another attempt will be made", zap.Error(catchupMsg.error))
+				go func() {
+					select {
+					case <-client.Ctx.Done():
+					case client.catchupPhaseRequest <- struct{}{}:
+					}
+				}()
```
julianbrost, line 327: IMO a bit strange to start a go routine that only has the purpose to send an event to the very event loop that started it. Also, it intentional that `catchupReset()` is done twice?

An alternative would be to just do the `client.startCatchupWorkers()` here as well. Also shouldn't this here be the only place then where an delay is needed? In that case, you wouldn't even have to duplicate that delay.

Maybe it's also an option to signal a restart due to an error via `client.catchupPhaseRequest` directly instead of `catchupEventCh`. (To be fair, I haven't fully thought this through, so that might result in other strange things.)

internal/icinga2/client.go
```diff
@@ -0,0 +1,388 @@
...
+		case <-client.catchupPhaseRequest:
+			if catchupEventCh != nil {
+				client.Logger.Warn("Switching to catch-up-phase was requested while already catching up, restarting phase")
+				catchupCancel()
+			}
+
+			client.Logger.Info("Worker enters catch-up-phase, start caching up on Event Stream events")
+			catchupReset()
+			catchupEventCh, catchupCancel = client.startCatchupWorkers(
+				min(3*time.Minute, time.Duration(math.Exp2(float64(catchupFailCounter))-1)*time.Second))
```
julianbrost, line 309: IMO the use of `math.Exp2` makes this harder to read than it has to be with the multiple casts required. As you don't need `catchupFailCounter` as an actual counter, you could consider something like a `var catchupDelay time.Duration` instead, where instead of of the `catchupFailCounter++`, you do something like this:

```go
// one-liner, maybe a the purpose of the max() in the middle is a bit unclear
catchupDelay = min(3*time.Minute, max(time.Second, 2*catchupDelay))

// expanded into a hopefully more readable if
if catchupDelay == 0 {
	catchupDelay = time.Second
} else {
	catchupDelay = min(3*time.Minute, 2*catchupDelay)
}
```

internal/icinga2/client.go
```diff
@@ -0,0 +1,388 @@
...
+
+	objTypes := []string{"host", "service"}
+	for _, objType := range objTypes {
+		objType := objType // https://go.dev/doc/faq#closures_and_goroutines
+		group.Go(func() error {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-time.After(delay):
+			}
```
julianbrost, line 219: Is the event loop in `client.worker()` expected to do anything while these are sleeping? Otherwise, wouldn't it make more sense to sleep in that loop before calling this function?

julianbrost (changes requested):

internal/icinga2/client.go
oxzi: I don't want to block the `worker`'s event loop with a busy sleep. Otherwise, all other signals have to wait, including the cancellation context. Thus, by moving the sleep within the `startCatchupWorkers` worker Goroutines, they don't block anything and can easily be canceled through their custom context.

Yes, I am not so proud of this design, but it has the least implications.

internal/icinga2/client.go
oxzi: Good idea. This is definitely a more readable approach.

julianbrost:
I noticed a small detail that would be pretty helpful: setting the user agent in all the requests. Icinga 2 logs the user agent for each request, so it would be helpful to easily identify the requests in these logs.

PR #114, opened 2023-10-19
Use `icinga-go-library`

yhabteab:
## Blocked By
- https://github.com/Icinga/icinga-go-library/pull/22


yhabteab:
I had to use the new, not yet [merged PR](https://github.com/Icinga/icinga-go-library/pull/17) as the `DB#BuildColumns()` was dropped during the migration to the new library and we have no other direct way to access the internal DB column mapper.
https://github.com/Icinga/icinga-notifications/blob/4ae3893165791b0ee36655e0cdbc54963b517cff/internal/event/event.go#L136-L137

internal/utils/utils.go
```diff
@@ -5,19 +5,17 @@ import (
...
 	"github.com/pkg/errors"
 	"slices"
 	"strings"
 )
 
 // BuildInsertStmtWithout builds an insert stmt without the provided column.
-func BuildInsertStmtWithout(db *icingadb.DB, into interface{}, withoutColumn string) string {
-	columns := db.BuildColumns(into)
+func BuildInsertStmtWithout(db *database.DB, into interface{}, withoutColumn string) string {
+	columns := slices.Clone(db.BuildColumns(into))
```
julianbrost, line 18: I think it might be a good idea to do the clone within the library, i.e. never expose the internal slice from the cache so that there's no way to break it from the outside.

go.mod
```diff
@@ -1,14 +1,16 @@
 module github.com/icinga/icinga-notifications
 
-go 1.21
+go 1.22
+
+toolchain go1.22.3
```
julianbrost, line 5: Seems like the `toolchain` directive was added because Go hat to increase the `go` directive for you:

> For reproducibility, the go command writes its own toolchain name in a toolchain line any time it is updating the go version in the go.mod file (usually during go get).

-- https://go.dev/ref/mod#go-mod-file-toolchain

But I don't see a particular reason why we would need or want it here.

cmd/icinga-notifications-daemon/main.go
```diff
@@ -89,7 +90,7 @@ func main() {
 
 	runtimeConfig := config.NewRuntimeConfig(icinga2Launcher.Launch, logs, db)
 	if err := runtimeConfig.UpdateFromDatabase(ctx); err != nil {
-		logger.Fatalw("failed to load config from database", zap.Error(err))
+		logger.Fatalf("%+v", errors.Wrap(err, "failed to load config from database"))
```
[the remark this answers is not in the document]
yhabteab: Nope!

cmd/icinga-notifications-daemon/main.go
```diff
@@ -62,24 +62,25 @@ func main() {
...
-	}
-
-	channel.UpsertPlugins(conf.ChannelPluginDir, logs.GetChildLogger("channel"), db)
 
 	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
 	defer cancel()
 
+	logger.Infof("Connecting to database at '%s'", db.GetAddr())
+	if err := db.PingContext(ctx); err != nil {
+		logger.Fatalf("%+v", err)
```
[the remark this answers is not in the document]
yhabteab: Fixed!

internal/utils/utils.go
oxzi, line 1: Unless I'm missing something, both `RemoveIf` and `RemoveNils` can be removed now.

oxzi (commented):
Please feel free removing the now unused `utils` functions. Otherwise, LGTM.

oxzi (dismissed):

go.mod
```diff
@@ -1,14 +1,14 @@
...
+go 1.22
 
 require (
 	github.com/creasty/defaults v1.7.0
 	github.com/emersion/go-sasl v0.0.0-20231106173351-e73c9f7bad43
 	github.com/emersion/go-smtp v0.21.2
 	github.com/goccy/go-yaml v1.11.3
 	github.com/google/uuid v1.6.0
-	github.com/icinga/icingadb v1.1.1-0.20230418113126-7c4b947aad3a
+	github.com/icinga/icinga-go-library v0.1.1-0.20240529153148-5054d55f8005
```
julianbrost, line 11: I just tagged the new version as `v0.2.0`, please update this to use the tagged version.

julianbrost (approved):

PR #117, opened 2023-10-25
golangci-lint: add as a CI and fix raised concerns

oxzi:
Implement the well known meta linter golangci-lint in the CI to check for common errors, also in Pull Requests.

This commit also fixes existing issues raised by golangci-lint: simplifications, cleaning up an unused private struct field, and setting an http.Server's timeouts, which are indefinite by default.

### DIFF
--- a/internal/listener/listener.go
+++ b/internal/listener/listener.go
@@ -48,7 +48,14 @@ func (l *Listener) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 
 func (l *Listener) Run() error {
 	l.logger.Infof("Starting listener on http://%s", l.configFile.Listen)
-	return http.ListenAndServe(l.configFile.Listen, l)
+	server := &http.Server{
+		Addr:         l.configFile.Listen,
+		Handler:      l,
+		ReadTimeout:  10 * time.Second,
+		WriteTimeout: 10 * time.Second,
+		IdleTimeout:  30 * time.Second,
+	}
+	return server.ListenAndServe()
 }
 
 func (l *Listener) ProcessEvent(w http.ResponseWriter, req *http.Request) {
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -5,6 +5,10 @@ on:
     branches: [ main ]
   pull_request: {}
 
+permissions:
+  # https://docs.github.com/en/rest/overview/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents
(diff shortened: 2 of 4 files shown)


.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,26 @@
...
+  push:
+    branches: [ main ]
+  pull_request: {}
+
+permissions:
+  # https://docs.github.com/en/rest/overview/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents
+  contents: read
+
+jobs:
+  go:
```
julianbrost, line 13: That's the same name as a job in `go.yml`, which is not a good idea, see #113.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,26 @@
...
+    - uses: actions/setup-go@v4
+      with:
+        go-version: stable
+    - name: golangci-lint
+      uses: golangci/golangci-lint-action@v3
+      with:
+        version: latest
+        # Enable the gosec linter w/o having to create a .golangci.yml config
+        args: -E gosec
+        only-new-issues: true
```
julianbrost, line 26: I would just add this (and the `permission:`, I wasn't aware that this existed so far) to the existing `go` job in `go.yml`. Both workflows don't take that long that we gain much by running them in parallel.

internal/listener/listener.go
```diff
@@ -48,7 +48,14 @@ func (l *Listener) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 
 func (l *Listener) Run() error {
 	l.logger.Infof("Starting listener on http://%s", l.configFile.Listen)
-	return http.ListenAndServe(l.configFile.Listen, l)
+	server := &http.Server{
+		Addr:         l.configFile.Listen,
+		Handler:      l,
+		ReadTimeout:  3 * time.Second,
+		WriteTimeout: 3 * time.Second,
```
julianbrost, line 55: I think the goal is to have a timeout at all, not a very tight one. I guess all values are pretty much chosen by a fair dice roll, but I'd go for maybe 10 seconds.

julianbrost (changes requested):
In general: I haven't used golangci-lint before, but why not, let's give it a try.

internal/listener/listener.go
oxzi: I just thought three seconds would be a very long time..

julianbrost (approved):

PR #122, opened 2023-11-02
Shutdown Context and Fixes

oxzi:
This PR works on the initial signal-based context changes from #112. During testing I found my Icinga Notifications container always being stuck when being shut downed until it got SIGKILLed. Thus, next to the introduction of the signal context, a context bug in the `RuntimeConfig` was fixed and a simple context wrapper was added to the `Listener`.

- __listener: context for Listener.Run__
  Allow gracefully stopping the internal web server when the parent context is closed.
- __runtime: fix RuntimeConfig.PeriodicUpdates context__
  The break only broke out of the select, but continued to for loop.
- __icinga-notifications-daemon: handle SIGINT/SIGTERM__
  Switch to a signal based context which is being canceled for SIGINT and SIGTERM. This change was extracted from the ongoing PR #112.

### DIFF
--- a/internal/listener/listener.go
+++ b/internal/listener/listener.go
@@ -46,7 +46,13 @@ func (l *Listener) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	l.mux.ServeHTTP(rw, req)
 }
 
-func (l *Listener) Run() error {
+// Run the Listener's web server and block until the server has finished.
+//
+// The web server either returns (early) when its ListenAndServe fails or when the given context is finished. After the
+// context is done, the web server shuts down gracefully with a hard limit of three seconds.
+//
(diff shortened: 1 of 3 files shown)


internal/listener/listener.go
```diff
@@ -46,7 +46,7 @@ func (l *Listener) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	l.mux.ServeHTTP(rw, req)
 }
 
-func (l *Listener) Run() error {
+func (l *Listener) Run(ctx context.Context) error {
```
julianbrost, line 49: Probably a good idea to document the new context + error return behavior of the function.

julianbrost (changes requested):

julianbrost (approved):

PR #123, opened 2023-11-13
Add encryption and pass auth to email channel

sukhwinder33445:


cmd/channel/email/main.go
```diff
@@ -13,9 +17,11 @@ import (
...
 
 type Email struct {
-	Host string `json:"host"`
-	Port string `json:"port"`
-	From string `json:"from"`
+	Host     string `json:"host"`
+	Port     string `json:"port"`
+	From     string `json:"from"`
+	Password string `json:"password"`
+	UseTLS   bool   `json:"tls"`
```
oxzi, line 24: I'm not sure if this is the right place to discuss this, but SMTP transport encryption is more ternary than binary.
There is:
- plaintext: no transport encryption at all
- STARTTLS: start with a plaintext SMTP connection and try to upgrade it to TLS (with all possible issues, <https://nostarttls.secvuln.info/>)
- TLS: TLS tunnel with SMTP within

By both calling the variable `UseTLS` and also using `tls` as the element's name, I find this misleading as a STARTTLS connection will be established.
https://github.com/Icinga/icinga-notifications/blob/64067793a9fccc42498051561eb91d78ea55638c/cmd/channel/email/main.go#L159-L166
https://github.com/Icinga/icinga-notifications/blob/64067793a9fccc42498051561eb91d78ea55638c/cmd/channel/email/main.go#L55-L59

cmd/channel/email/main.go
```diff
@@ -40,12 +46,46 @@ func (ch *Email) SendNotification(req *plugin.NotificationRequest) error {
...
+		if err = c.Rcpt(addr); err != nil {
+			return err
+		}
+	}
+	w, err := c.Data()
+	if err != nil {
+		return err
+	}
+
+	if _, err = w.Write(bytes.ReplaceAll(msg.Bytes(), []byte("\n"), []byte("\r\n"))); err != nil {
```
oxzi, line 80: Why replace the newline characters? Maybe it might be better to define a proper Content-Type.

Btw, the handmade email header field generation looks a bit thin and not RFC-compliant, especially regards special characters and the like.

https://github.com/Icinga/icinga-notifications/blob/64067793a9fccc42498051561eb91d78ea55638c/cmd/channel/email/main.go#L44-L45

cmd/channel/email/main.go
```diff
@@ -80,7 +120,63 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
...
+			},
+		},
+		{
+			Name: "port",
+			Type: "number",
+			Label: map[string]string{
+				"en_US": "SMTP Port",
+				"de_DE": "SMTP Port",
+			},
+			Min: types.Int{NullInt64: sql.NullInt64{Int64: 0, Valid: true}},
```
oxzi, line 139: Might be nitpicking, but port 0 might be unwanted.

julianbrost:
By the way, so far this uses `net/smtp` just because it's in the standard library and is good enough to send something as a first test. If there's a suitable library that makes it much easier to do things right, using that can be an option for sure.

cmd/channel/email/main.go
```diff
@@ -38,17 +44,49 @@ func (ch *Email) SendNotification(req *plugin.NotificationRequest) error {
...
+			}
+			if err = c.Auth(auth); err != nil {
+				return err
+			}
+		}
+		if err := c.SendMail(reversePath, recipients, msgReader); err != nil {
+			return err
+		}
+
+		return c.Quit()
```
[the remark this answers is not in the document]
oxzi: However, instead of going with this home-brew approach, I would consider using a `smtp.Client` as I tried in <https://github.com/Icinga/icinga-notifications/commit/9308b97c2b615eb291e45bb317b9a997b7482866>. Please be aware that I have wrote this according to both go-smtp's API doc and source, but haven't tested it :-)

cmd/channel/email/main.go
```diff
@@ -38,17 +44,49 @@ func (ch *Email) SendNotification(req *plugin.NotificationRequest) error {
...
+	if ch.Password != "" {
+		auth = sasl.NewPlainClient("", ch.User, ch.Password)
 	}
 
-	return nil
+	switch ch.Encryption {
+	case "tls":
+		return smtp.SendMailTLS(ch.GetServer(), auth, reversePath, recipients, msgReader)
+	case "starttls":
+		return smtp.SendMail(ch.GetServer(), auth, reversePath, recipients, msgReader)
```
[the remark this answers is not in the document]
oxzi: Before working on this one, please read my second comment in the other thread after the `default` case.

oxzi (commented):
Thanks for the changes. Please feel free to have a look at the comments I made.

cmd/channel/email/main.go
```diff
@@ -38,17 +44,49 @@ func (ch *Email) SendNotification(req *plugin.NotificationRequest) error {
 	}
 
 	var msg bytes.Buffer
-	_, _ = fmt.Fprintf(&msg, "To: %s\n", strings.Join(to, ","))
-	_, _ = fmt.Fprintf(&msg, "Subject: [#%d] %s %s is %s\n\n", req.Incident.Id, req.Event.Type, req.Object.Name, req.Event.Severity)
-
 	plugin.FormatMessage(&msg, req)
 
-	err := smtp.SendMail(ch.GetServer(), nil, ch.From, to, bytes.ReplaceAll(msg.Bytes(), []byte("\n"), []byte("\r\n")))
+	return enmime.Builder().
```
oxzi, line 49: It seems like `enmime` does not add a Message-ID by default.

Thus, might you please add this header manually. For example like this:
```go
Header("Message-Id", fmt.Sprintf("<%d-%x-%s>", req.Incident.Id, time.Now().UnixNano(), ch.SenderMail)).
```

oxzi (approved):

cmd/channel/email/main.go
```diff
@@ -5,31 +5,39 @@ import (
...
 
 type Email struct {
-	Host string `json:"host"`
-	Port string `json:"port"`
-	From string `json:"from"`
+	Host       string `json:"host"`
+	Port       string `json:"port"`
+	SenderName string `json:"sender_name"`
+	SenderMail string `json:"sender_mail"`
+	From       string `json:"from"`
```
julianbrost, line 26: `From` looks unused. Was it replaced completely by these other two options?

cmd/channel/email/main.go
```diff
@@ -38,17 +46,50 @@ func (ch *Email) SendNotification(req *plugin.NotificationRequest) error {
...
-	_, _ = fmt.Fprintf(&msg, "Subject: [#%d] %s %s is %s\n\n", req.Incident.Id, req.Event.Type, req.Object.Name, req.Incident.Severity)
-
 	plugin.FormatMessage(&msg, req)
 
-	err := smtp.SendMail(ch.GetServer(), nil, ch.From, to, bytes.ReplaceAll(msg.Bytes(), []byte("\n"), []byte("\r\n")))
+	return enmime.Builder().
+		ToAddrs(to).
+		From(ch.SenderName, ch.SenderMail).
+		Subject(fmt.Sprintf("[#%d] %s %s is %s", req.Incident.Id, req.Event.Type, req.Object.Name, req.Incident.Severity)).
+		Header("Message-Id", fmt.Sprintf("<%d-%x-%s>", req.Incident.Id, time.Now().UnixNano(), ch.SenderMail)).
```
julianbrost, line 55: I'd just go for a UUID instead of `req.Incident.Id` and `time.Now().UnixNano()` here.

cmd/channel/email/main.go
```diff
@@ -38,17 +46,50 @@ func (ch *Email) SendNotification(req *plugin.NotificationRequest) error {
...
 		return err
 	}
+	defer client.Close()
 
-	return nil
+	if ch.Encryption == "starttls" {
+		if err = client.StartTLS(nil); err != nil {
+			return err
+		}
+	}
```
julianbrost, line 80: What's the default certificate validation behavior of `DialTLS()` and `StartTLS()`?

julianbrost (changes requested):

cmd/channel/email/main.go
sukhwinder33445: Yes, I forgot to remove it from here.

cmd/channel/email/main.go
sukhwinder33445: The `serverName` is required to validate the certificate. We already pass this to `smtp.DialTLS/smtp.Dial`.
It is then passed on and during the handshake it is checked whether the certificates are valid.

go.mod
```diff
@@ -4,8 +4,11 @@ go 1.21
 
 require (
 	github.com/creasty/defaults v1.7.0
+	github.com/emersion/go-sasl v0.0.0-20200509203442-7bfe0ed36a21
+	github.com/emersion/go-smtp v0.20.1
 	github.com/goccy/go-yaml v1.11.2
 	github.com/icinga/icingadb v1.1.1-0.20230418113126-7c4b947aad3a
+	github.com/jhillyerd/enmime v1.0.1
```
julianbrost, line 11: There was a new release: https://github.com/jhillyerd/enmime/releases/tag/v1.1.0

Please update now, otherwise we'd just get a dependabot PR anyways.

go.mod
```diff
@@ -4,8 +4,11 @@ go 1.21
 
 require (
 	github.com/creasty/defaults v1.7.0
+	github.com/emersion/go-sasl v0.0.0-20200509203442-7bfe0ed36a21
```
julianbrost, line 7: No releases? That's a bit unfortunate. But the commit you're referencing isn't the latest one, so you might want to update that one as well.

julianbrost (changes requested):

cmd/channel/email/main.go
```diff
@@ -5,31 +5,43 @@ import (
...
+	"net/mail"
 	"os"
 	"os/user"
-	"strings"
+)
+
+const (
+	TypeNone     = "none"
+	TypeStartTls = "starttls"
+	TypeTls      = "tls"
```
julianbrost, line 23: Why `Type...`? I'd name them something like `EncryptionNone`. Also, acronyms are typically treated special (see for example [`http.ParseHTTPVersion`](https://pkg.go.dev/net/http#ParseHTTPVersion)), so it should be `EncryptionStartTLS` and `EncryptionTLS`.

julianbrost (changes requested):
Looks like merging #147 introduced merge conflicts here, so you also have to rebase once again unfortunately.

julianbrost:
Code looks fine to me, I just have to spin it up for some testing. @oxzi you could take another look at it in the meantime.

julianbrost (dismissed):

julianbrost:
I just had a look whether there were any significant changes since https://github.com/Icinga/icinga-notifications/pull/123#pullrequestreview-1742193488 ([diff for that range](https://github.com/Icinga/icinga-notifications/compare/cdcc1392ce18550997f9390f51ebacadd5ec24e1..d5090338ffec55f000a365b96264fb054029f892#diff-87f9e9ce351d7546b7247986660ac86aad794499d6c0aa961b25333adcf24180)) and noticed that the `Message-ID` was removed (done in https://github.com/Icinga/icinga-notifications/pull/123#event-11419299418).

@sukhwinder33445 Was this removed accidentally during rebase? If not, why was it removed?

cmd/channel/email/main.go
julianbrost: You've pushed an outdated version which does not include the changes that were already done in https://github.com/Icinga/icinga-notifications/pull/123#event-11168199610.

(Hint: this is a reply to https://github.com/Icinga/icinga-notifications/pull/123/files#r1414033413, GitHub sometimes doesn't show it clearly if you reply to existing threads in a review.)

julianbrost (changes requested):

julianbrost (dismissed):

cmd/channel/email/main.go
```diff
@@ -76,14 +122,35 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
 			return fmt.Errorf("failed to get the os's current user: %w", err)
 		}
 
-		ch.From = usr.Username + "@" + hostname
+		ch.SenderMail = usr.Username + "@" + hostname
+	}
+
+	if ch.User == "" {
+		ch.User = ch.SenderMail
 	}
```
oxzi, line 131: Honestly, I am a bit unhappy with all those undocumented fallback or default values.

- `Host` and `Port` might each fall back to `localhost` and `25`, respectively.
  Does this even make sense from a user's perspective? Why should one leave those fields empty anyway?
- `SenderName` is allowed to be left out, resulting in emails with a _From_ header like `From: <${SenderMail}>`, which is totally fine, imho.
- `SenderMail` will fallback to `${OS_USER}@${OS_HOSTNAME}`, which is also quite arbitrary, but something is required.
- `User` and `Password` are being used for the SMTP authentication, but `User` shouldn't fall back to something else, as I mentioned above.
- `Encryption` is the only "required" argument - which, however, is only a web check, relying on this component, and, if being set to neither `starttls` nor `tls`, implicitly falling back to the `none` logic.

Might I suggest not setting any defaults, but marking all necessary fields as "required" and checking if they have valid values in this method, resulting in an error otherwise.

cmd/channel/email/main.go
```diff
@@ -76,14 +122,35 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
 			return fmt.Errorf("failed to get the os's current user: %w", err)
 		}
 
-		ch.From = usr.Username + "@" + hostname
+		ch.SenderMail = usr.Username + "@" + hostname
+	}
+
+	if ch.User == "" {
+		ch.User = ch.SenderMail
```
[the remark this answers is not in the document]
oxzi: Additionally, when `User` and `Password` are only used together. Thus, it should be checked that if one is set, the other is also set. An assumption like `(ch.User == "") == (ch.Password == "")`.

cmd/channel/email/main.go
```diff
@@ -38,17 +51,50 @@ func (ch *Email) SendNotification(req *plugin.NotificationRequest) error {
...
+
+	if ch.Encryption == EncryptionTLS {
+		client, err = smtp.DialTLS(ch.GetServer(), nil)
+	} else {
+		client, err = smtp.Dial(ch.GetServer())
+	}
 	if err != nil {
 		return err
 	}
+	defer client.Close()
```
oxzi, line 79: `Close()` returns an error. If we are going to ignore `defer` errors, maybe make it explicit.
```suggestion
	defer func() { _ = client.Close() }()
```

cmd/channel/email/main.go
```diff
@@ -99,17 +166,39 @@ func (ch *Email) GetInfo() *plugin.Info {
...
+				"de_DE": "SMTP Passwort",
+			},
+		},
+		{
+			Name:     "encryption",
+			Type:     "option",
+			Required: true,
+			Label: map[string]string{
+				"en_US": "Transport Encryption",
+				"de_DE": "Transportverschlüsselung",
```
oxzi, line 195:
```suggestion
				"en_US": "SMTP Transport Encryption",
				"de_DE": "SMTP Transportverschlüsselung",
```

All other SMTP related fields are prefixed with "SMTP". I guess, for a user it eliminates some uncertainties.

oxzi (commented):
I have done some basic tests on this PR and it seems to work as intended. However, when going over the code, the fallback configuration values bothered me.

oxzi:
Pardon me if some of my comments missed the point of this PR.

Personally, I'm just not a fan of opinionated defaults, especially when they're not documented clearly. There have been too many times when I have had to deal with software that has made decisions that were not obvious to me.

However, at the very least default settings should be consistent, which in my opinion they are not in the current state: <https://github.com/Icinga/icinga-notifications/pull/123#discussion_r1447162261>, <https://github.com/Icinga/icinga-notifications/pull/123#discussion_r1447142112>.

oxzi (approved):
Thanks for your changes, LGTM!

PR #124, opened 2023-11-14
Improve channel plugin

sukhwinder33445:


pkg/rpc/rpc.go
```diff
@@ -45,10 +45,9 @@ type RPC struct {
 	lastRequestId   uint64
 	requestsMu      sync.Mutex
 
-	errChannel        chan struct{} // never transports a value, only closed through setErr() to signal an occurred error
-	err               *Error        // only initialized via setErr(), if a rpc (Fatal/non-recoverable) error has occurred
-	errOnce           sync.Once
-	requestedShutdown bool
+	errChannel chan struct{} // never transports a value, only closed through setErr() to signal an occurred error
+	err        *Error        // only initialized via setErr(), if a rpc (Fatal/non-recoverable) error has occurred
```
julianbrost, line 49: I'd rename those to something like `processResponsesErr(Ch)` as these represent whether that loop has terminated due to an error (counting orderly shutdown due to EOF as an error here).

pkg/rpc/rpc.go
```diff
@@ -87,13 +86,14 @@ func (r *RPC) Call(method string, params json.RawMessage) (json.RawMessage, erro
 	r.requestsMu.Unlock()
 
 	r.encoderMu.Lock()
+	if r.encoder == nil {
+		return nil, errors.New("cannot process any further requests, writer already closed")
```
julianbrost, line 90: You're not unlocking the mutex here, so this will only work once.

pkg/rpc/rpc.go
```diff
@@ -45,10 +45,9 @@ type RPC struct {
 	lastRequestId   uint64
 	requestsMu      sync.Mutex
 
-	errChannel        chan struct{} // never transports a value, only closed through setErr() to signal an occurred error
-	err               *Error        // only initialized via setErr(), if a rpc (Fatal/non-recoverable) error has occurred
-	errOnce           sync.Once
-	requestedShutdown bool
```
julianbrost, line 51: Please create a separate PR for removing that unused field so that we can make the CI happy again sooner.

julianbrost (changes requested):

internal/channel/plugin.go
```diff
@@ -36,33 +36,49 @@ func NewPlugin(pluginType string, logger *zap.SugaredLogger) (*Plugin, error) {
...
 	}
+	ioPipesToClose = append(ioPipesToClose, resWrite)
 
-	errPipe, err := cmd.StderrPipe()
+	logRead, logWrite, err := os.Pipe()
 	if err != nil {
-		return nil, fmt.Errorf("failed to create stderr pipe: %w", err)
+		ioPipesToClose = append(ioPipesToClose, reqWrite, resRead)
+		return nil, fmt.Errorf("failed to create logRead/logWrite pipe: %w", err)
 	}
```
julianbrost, line 63: In those two error cases, you're listing the names of files from all previous steps. Yes, it's only two times, so not a lot of repetition, but I quite like how `os/exec` does this internally, by tracking separately which ones are for the child process (those are closed always) and which ones are for the parent process (those are only closed in case of an error). That way, you only need to reference specific files in the block that's responsible for creating them.

pkg/rpc/rpc.go
```diff
@@ -86,13 +85,17 @@ func (r *RPC) Call(method string, params json.RawMessage) (json.RawMessage, erro
...
 	r.encoderMu.Unlock()
 	if err != nil {
-		err = fmt.Errorf("failed to write request: %w", err)
-		r.setErr(err)
-
-		return nil, r.Err()
+		r.encoder = nil
+		_ = r.writer.Close()
+		return nil, fmt.Errorf("failed to write request: %w", err)
 	}
```
julianbrost, line 99: The operations in line 96 and 97 must be done while the mutex is still locked.

pkg/rpc/rpc.go
```diff
@@ -112,48 +115,39 @@ func (r *RPC) Call(method string, params json.RawMessage) (json.RawMessage, erro
...
 	for r.Err() == nil {
 		var response Response
-		err := r.decoder.Decode(&response)
-
-		if err != nil {
-			r.setErr(fmt.Errorf("failed to read json response: %w", err))
+		if err := r.decoder.Decode(&response); err != nil {
+			r.processResponsesErr = &Error{cause: fmt.Errorf("failed to read json response: %w", err)}
+			_ = r.Close()
+			close(r.processResponsesErrCh)
```
julianbrost, line 150: I'd swap those two operations around. Closing the channel can always proceed immediately whereas `r.Close()` might have to wait for the lock for a short while.

julianbrost (changes requested):

pkg/plugin/plugin.go
```diff
@@ -132,19 +132,23 @@ func RunPlugin(plugin Plugin) {
...
 		var req rpc.Request
 		err := decoder.Decode(&req)
 		if err != nil {
 			if errors.Is(err, io.EOF) {
 				// plugin shutdown requested
-				return
+				break
 			}
 
 			log.Fatal("failed to read request:", err)
```
oxzi, line 146: If I'm not missing something, this would go to the stdout where the parent's `RPC` through the `resRead` pipe would read.

Thus, maybe configure logging to stderr or change logging to zap and let it log to stderr to match with the rest of the code base.

pkg/rpc/rpc.go
```diff
@@ -86,14 +85,19 @@ func (r *RPC) Call(method string, params json.RawMessage) (json.RawMessage, erro
 	r.requestsMu.Unlock()
 
 	r.encoderMu.Lock()
```
oxzi, line 87: Might be nit-picky, but there are three following unlock cases which makes this a bit hard to read. Even as I dislike kind of "useless" or "inline-able" methods, maybe consider refactoring this into a helper method where you can just `defer` the unlocking and return the error?

oxzi (commented):
Went over the change and it looks okay so far; please look at the two comments.

pkg/plugin/plugin.go
julianbrost: The `log` package uses stderr by default: https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/log/log.go;l=87 (interestingly, this seems to be mentioned nowhere on https://pkg.go.dev/log)

You probably get some duplicate timestamps with that though.

oxzi: Oh thanks, my bad. I wrongly assumed that it would be logged to stdout.

oxzi (approved):
I have successfully tested the changes and couldn't trigger an error during testing, killing plugins included. LGTM.

PR #125, opened 2023-11-14
golangci-lint: disable only-new-issues

julianbrost:
`only-new-issues` seems to hide some issues in PRs even if they are introduced there.  For example, removing all usages of a struct field without touching the field itself is not detected. This can result in CI failures after merging, see #108 and 6f4b437eaa2f6f42ec56f9cfdd715ace071be04d for example.


oxzi (approved):
When searching the net for problems with `only-new-issues`, some others are popping up. Thus, seems like there are some issues in this regard.

As there should be nothing to complain on our current `main` branch anyway, we should be able to drop this option.

PR #129, opened 2023-11-21
Load active incidents gracefully

### LINKED ISSUE #119: Improve incident opening/loading code
Follow-up issue for: #68

That PR adds code to load all open incidents on daemon startup. Currently, due to the existing code, it does so by querying the `object_id` of all open incidents and then loading the current incident for each object individually.

So the task here is to look into how to restructure `internal/incident/` and `internal/object/` so that it better fits its current uses, i.e. allow for some "just select everything from a few tables and construct the in-memory data structures for many objects/incidents from that".

yhabteab:
resolves #119

### DIFF
--- a/.github/workflows/compliance.yml
+++ b/.github/workflows/compliance.yml
@@ -31,4 +31,4 @@ jobs:
         # and this very icinga-notifications are licensed under GPL-2.0.
         # https://github.com/Icinga/icingadb/blob/v1.1.1/.github/workflows/compliance/check-licenses.sh
         go-licenses check github.com/icinga/icinga-notifications/... \
-          --allowed_licenses BSD-2-Clause,BSD-3-Clause,GPL-2.0,MIT,MPL-2.0
+          --allowed_licenses BSD-2-Clause,BSD-3-Clause,GPL-2.0,ISC,MIT,MPL-2.0
--- a/internal/incident/incidents.go
+++ b/internal/incident/incidents.go
@@ -6,10 +6,10 @@ import (
 	"github.com/icinga/icinga-notifications/internal/config"
 	"github.com/icinga/icinga-notifications/internal/event"
 	"github.com/icinga/icinga-notifications/internal/object"
+	"github.com/icinga/icinga-notifications/internal/utils"
 	"github.com/icinga/icingadb/pkg/com"
 	"github.com/icinga/icingadb/pkg/icingadb"
 	"github.com/icinga/icingadb/pkg/logging"
-	"github.com/jmoiron/sqlx"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
@@ -77,8 +77,8 @@ func LoadOpenIncidents(ctx context.Context, db *icingadb.DB, logger *logging.Log
 					chunkLen := len(bulk)
 					objectIds := make([]any, chunkLen)
 					incidentIds := make([]any, chunkLen)
-					incidentsById := make(map[any]*Incident, chunkLen)
-					incidentsByObjId := make(map[any]*Incident, chunkLen)
+					incidentsById := make(map[int64]*Incident, chunkLen)
+					incidentsByObjId := make(map[string]*Incident, chunkLen)
 
 					for k, i := range bulk {
 						incidentsById[i.Id] = i
@@ -89,27 +89,52 @@ func LoadOpenIncidents(ctx context.Context, db *icingadb.DB, logger *logging.Log
 					}
 
 					// Restore all incident objects matching the given object ids
-					if err := restoreAttrsFor(chCtx, db, incidentsByObjId, new(object.Object), objectIds); err != nil {
+					err := utils.ForEachRow[object.Object](chCtx, db, "id", objectIds, func(o *object.Object) {
+						incidentsByObjId[o.ID.String()].Object = o
+					})
+					if err != nil {
 						return errors.Wrap(err, "cannot restore incidents object")
 					}
 
 					// Restore object ID tags matching the given object ids
-					if err := restoreAttrsFor(chCtx, db, incidentsByObjId, new(object.IdTagRow), objectIds); err != nil {
+					err = utils.ForEachRow[object.IdTagRow](ctx, db, "object_id", objectIds, func(ir *object.IdTagRow) {
+						incidentsByObjId[ir.ObjectId.String()].Object.Tags[ir.Tag] = ir.Value
+					})
+					if err != nil {
 						return errors.Wrap(err, "cannot restore incident object ID tags")
 					}
 
 					// Restore object extra tags matching the given object ids
-					if err := restoreAttrsFor(chCtx, db, incidentsByObjId, new(object.ExtraTagRow), objectIds); err != nil {
-						return errors.Wrap(err, "cannot restore incident object ID tags")
+					err = utils.ForEachRow[object.ExtraTagRow](ctx, db, "object_id", objectIds, func(et *object.ExtraTagRow) {
+						incidentsByObjId[et.ObjectId.String()].Object.ExtraTags[et.Tag] = et.Value
+					})
+					if err != nil {
+						return errors.Wrap(err, "cannot restore incident object extra tags")
 					}
 
 					// Restore all escalation states and incident rules matching the given incident ids.
-					if err := restoreAttrsFor(chCtx, db, incidentsById, new(EscalationState), incidentIds); err != nil {
+					err = utils.ForEachRow[EscalationState](ctx, db, "incident_id", incidentIds, func(state *EscalationState) {
+						i := incidentsById[state.IncidentID]
+						i.EscalationState[state.RuleEscalationID] = state
+
+						// Restore the incident rule matching the current escalation state if any.
+						i.runtimeConfig.RLock()
+						defer i.runtimeConfig.RUnlock()
+
+						escalation := i.runtimeConfig.GetRuleEscalation(state.RuleEscalationID)
+						if escalation != nil {
+							i.Rules[escalation.RuleID] = struct{}{}
+						}
+					})
+					if err != nil {
 						return errors.Wrap(err, "cannot restore incident rule escalation states")
 					}
 
 					// Restore incident recipients matching the given incident ids.
-					if err := restoreAttrsFor(chCtx, db, incidentsById, new(ContactRow), incidentIds); err != nil {
+					err = utils.ForEachRow[ContactRow](ctx, db, "incident_id", incidentIds, func(c *ContactRow) {
+						incidentsById[c.IncidentID].Recipients[c.Key] = &RecipientState{Role: c.Role}
+					})
+					if err != nil {
 						return errors.Wrap(err, "cannot restore incident recipients")
 					}
 
@@ -236,100 +261,4 @@ func ProcessEvent(
 	}
 
 	return currentIncident.ProcessEvent(ctx, ev, created)
-}
-
-// restoreAttrsFor restores the attributes of the given subject type from the database.
-//
-// It bulks SELECT the info from the database via a `SELECT ... WHERE ... IN(?)` query scoped to the specified args.
-// The column name for the where clause is determined automatically based on the provided subject type.
-// Currently, it only supports object.Object, object.IdTagRow, object.ExtraTagRow, EscalationState, ContactRow types
-// and panics if something else is provided.
-//
-// Note, this function accesses/modifies the specified incidents map without obtaining any locks
-// and must be protected against simultaneous write operations if a shared map is used.
-func restoreAttrsFor(ctx context.Context, db *icingadb.DB, incidents map[any]*Incident, subject any, scopes []any) error {
-	var (
-		// Name of the column used to filter the database query by.
-		column string
-		// A callback set based on the specified subject and called after each successful row retrieval.
-		restore func(any)
-		// A factory func set based on the given subject and called before each row scan.
-		factory func() any
-	)
-
-	switch subject.(type) {
-	case *EscalationState:
-		column = "incident_id"
-		factory = func() any { return new(EscalationState) }
-		restore = func(r any) {
-			state := r.(*EscalationState)
-			i := incidents[state.IncidentID]
-			i.EscalationState[state.RuleEscalationID] = state
-
-			// Restore the incident rule matching the current escalation state if any.
-			i.runtimeConfig.RLock()
-			escalation := i.runtimeConfig.GetRuleEscalation(state.RuleEscalationID)
-			if escalation != nil {
-				i.Rules[escalation.RuleID] = struct{}{}
-			}
-			i.runtimeConfig.RUnlock()
-		}
-	case *ContactRow:
-		column = "incident_id"
-		factory = func() any { return new(ContactRow) }
-		restore = func(r any) {
-			c := r.(*ContactRow)
-			incidents[c.IncidentID].Recipients[c.Key] = &RecipientState{Role: c.Role}
-		}
-	case *object.IdTagRow:
-		column = "object_id"
-		factory = func() any { return new(object.IdTagRow) }
-		restore = func(r any) {
-			id := r.(*object.IdTagRow)
-			incidents[id.ObjectId.String()].Object.Tags[id.Tag] = id.Value
-		}
-	case *object.ExtraTagRow:
-		column = "object_id"
-		factory = func() any { return new(object.ExtraTagRow) }
-		restore = func(r any) {
-			extraTag := r.(*object.ExtraTagRow)
-			incidents[extraTag.ObjectId.String()].Object.ExtraTags[extraTag.Tag] = extraTag.Value
-		}
-	case *object.Object:
-		column = "id"
-		ev := &event.Event{Tags: make(map[string]string), ExtraTags: make(map[string]string)}
-		factory = func() any { return object.New(db, ev) }
-		restore = func(r any) {
-			obj := r.(*object.Object)
-			incidents[obj.ID.String()].Object = obj
-		}
-	default: // should never happen!
-		panic(fmt.Sprintf("invalid database subject for incient#restoreAttrsFor() provided: %v", subject))
-	}
-
-	query := fmt.Sprintf("%s WHERE %q IN(?)", db.BuildSelectStmt(subject, subject), column)
-	stmt, args, err := sqlx.In(query, scopes)
-	if err != nil {
-		return errors.Wrapf(err, "cannot build placeholders for %q", query)
-	}
-
-	rows, err := db.QueryxContext(ctx, db.Rebind(stmt), args...)
-	if err != nil {
-		return err
-	}
-	// In case the records in the loop below are successfully traversed, rows is automatically closed and an
-	// error is returned (if any), making this rows#Close() call a no-op. Escaping from this function unexpectedly
-	// means we have a more serious problem, so in either case just discard the error here.
-	defer func() { _ = rows.Close() }()
-
-	for rows.Next() {
-		row := factory()
-		if err = rows.StructScan(row); err != nil {
-			return err
-		}
-
-		restore(row)
-	}
-
-	return rows.Err()
 }
--- a/internal/incident/incidents.go
+++ b/internal/incident/incidents.go
@@ -76,17 +76,17 @@ func LoadOpenIncidents(ctx context.Context, db *icingadb.DB, logger *logging.Log
 
 				g.Go(func() error {
 					chunkLen := len(bulk)
-					objectIds := make([]types.Binary, chunkLen)
-					incidentIds := make([]int64, chunkLen)
+					objectIds := make([]types.Binary, 0, chunkLen)
+					incidentIds := make([]int64, 0, chunkLen)
 					incidentsById := make(map[int64]*Incident, chunkLen)
 					incidentsByObjId := make(map[string]*Incident, chunkLen)
 
-					for k, i := range bulk {
+					for _, i := range bulk {
 						incidentsById[i.Id] = i
 						incidentsByObjId[i.ObjectID.String()] = i
 
-						objectIds[k] = i.ObjectID
-						incidentIds[k] = i.Id
+						objectIds = append(objectIds, i.ObjectID)
+						incidentIds = append(incidentIds, i.Id)
 					}
 
 					// Restore all incident objects matching the given object ids
--- a/internal/incident/incidents.go
+++ b/internal/incident/incidents.go
@@ -10,6 +10,7 @@ import (
 	"github.com/icinga/icingadb/pkg/com"
 	"github.com/icinga/icingadb/pkg/icingadb"
 	"github.com/icinga/icingadb/pkg/logging"
+	"github.com/icinga/icingadb/pkg/types"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
@@ -75,8 +76,8 @@ func LoadOpenIncidents(ctx context.Context, db *icingadb.DB, logger *logging.Log
 
 				g.Go(func() error {
 					chunkLen := len(bulk)
-					objectIds := make([]any, chunkLen)
-					incidentIds := make([]any, chunkLen)
+					objectIds := make([]types.Binary, chunkLen)
+					incidentIds := make([]int64, chunkLen)
 					incidentsById := make(map[int64]*Incident, chunkLen)
 					incidentsByObjId := make(map[string]*Incident, chunkLen)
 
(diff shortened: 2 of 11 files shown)


internal/incident/incidents.go
```diff
@@ -25,34 +23,116 @@ var (
 func LoadOpenIncidents(ctx context.Context, db *icingadb.DB, logger *logging.Logger, runtimeConfig *config.RuntimeConfig) error {
 	logger.Info("Loading all active incidents from database")
 
-	var objectIDs []types.Binary
-	err := db.SelectContext(ctx, &objectIDs, `SELECT object_id FROM incident WHERE "recovered_at" IS NULL`)
-	if err != nil {
-		logger.Errorw("Failed to load active incidents from database", zap.Error(err))
+	incidents := make(chan *Incident, 1)
+	g, _ := errgroup.WithContext(ctx)
```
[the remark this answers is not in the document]
yhabteab: > This one would get canceled if one of the `g.Go(func() error { … })` fails

🤦‍♂️! I was only thinking of the context being cancelled by the syscall signals.

internal/incident/incidents.go
```diff
@@ -25,34 +23,116 @@ var (
...
-	for _, objectID := range objectIDs {
-		obj, err := object.LoadFromDB(ctx, db, objectID)
+		query := `
+			SELECT incident.id, started_at, severity, incident.object_id, object.source_id, object.name, object.url, oit.tag, oit.value, oet.tag, oet.value
+			FROM incident
+				INNER JOIN object on object.id = incident.object_id
+				INNER JOIN public.object_id_tag oit on object.id = oit.object_id
+				LEFT JOIN public.object_extra_tag oet on object.id = oet.object_id
+			WHERE "recovered_at" IS NULL
+			ORDER BY incident.id, object.id`
```
julianbrost, line 42: This will select quite a lot of redundant information. Especially with lots of tags, you get the Cartesian product of ID tags and extra tags, i.e. `len(id_tags) * max(1, len(extra_tags))` rows. We don't have a limit on the number of tags (note in particular that we're not limited to `host` + `service`), so this can get large quite quickly.

So it might be worth it to split this across multiple queries. This would then also allow to get some useful parallelism here. Like do an query like this first:

```sql
SELECT id, object_id, [...] FROM incident WHERE recovered_at IS NULL;
```

And then take the resulting object IDs and perform queries like this with them:

```sql
SELECT [...] FROM object WHERE id IN (?, ?, [...]);
SELECT [...] FROM object_id_tag WHERE object_id IN (?, ?, [...]);
SELECT [...] FROM object_extra_tag WHERE object_id IN (?, ?, [...]);
```

Where those queries could be run in parallel. Also, the set of object IDs has to be partitioned to avoid running into number of placeholders limits, which happens to allow additional parallelism. But don't overdo it, we don't want to overwhelm the database and also keep the code complexity reasonable. So simply partitioning the object IDs and running the three queries in sequence for each subset might be good enough, as this should save a lot of synchronization when merging the results and if there's enough data to be read that actually makes parallelization relevant, enough of it should already happen due to processing different IDs in parallel.

julianbrost (changes requested):

internal/incident/incidents.go
```diff
@@ -181,3 +238,99 @@ func ProcessEvent(
...
+	var (
+		// Name of the column used to filter the database query by.
+		column string
+		// A callback set based on the specified subject and called after each successful row retrieval.
+		restore func(any)
+		// A factory func set based on the given subject and called before each row scan.
+		factory func() any
+	)
+
+	switch subject.(type) {
```
julianbrost, line 261: That can be done without the type switch. I've started a draft where one of the calls of this function is replaced with another function using generics instead: https://github.com/Icinga/icinga-notifications/commit/77f93ce291ee4e0ee6e15074cbb6f96536416416

internal/incident/incidents.go
```diff
@@ -25,38 +26,142 @@ var (
...
+		for {
+			select {
+			case <-chCtx.Done():
+				return chCtx.Err()
+			case bulk, ok := <-bulks:
+				if !ok {
+					return nil
+				}
+
+				g.Go(func() error {
```
julianbrost, line 76: There should be a limit on how many of these are started in parallel.

julianbrost (changes requested):

internal/object/object.go
```diff
@@ -24,28 +25,84 @@ var (
...
+		defer cacheMu.Unlock()
+
+		cache[et.ObjectId.String()].ExtraTags[et.Tag] = et.Value
+	})
+	if err != nil {
+		return errors.Wrap(err, "cannot restore incident object extra tags")
+	}
+
+	return nil
+}
```
julianbrost, line 104: This locks `cacheMu` quite a few times. It would make sense to create a local map first, and move all the entries to the shared `cache` at the end. Not only to avoid extra locking but also that there are no half-initialized objects in `cache` (which may even be left there if the function would return with an error.

internal/incident/incidents.go
```diff
@@ -25,38 +26,122 @@ var (
...
+					chunkLen := len(bulk)
+					objectIds := make([]any, chunkLen)
+					incidentIds := make([]any, chunkLen)
```
julianbrost, line 79: Shouldn't it be possible to use `[]types.Binary` and `[]int64` here as well?

julianbrost (changes requested):

internal/incident/incidents.go
```diff
@@ -25,38 +27,122 @@ var (
...
-// Returns error ony database failure.
+// Returns error on any database failure.
 func LoadOpenIncidents(ctx context.Context, db *icingadb.DB, logger *logging.Logger, runtimeConfig *config.RuntimeConfig) error {
 	logger.Info("Loading all active incidents from database")
 
-	var objectIDs []types.Binary
-	err := db.SelectContext(ctx, &objectIDs, `SELECT object_id FROM incident WHERE "recovered_at" IS NULL`)
-	if err != nil {
-		logger.Errorw("Failed to load active incidents from database", zap.Error(err))
+	g, chCtx := errgroup.WithContext(ctx)
```
julianbrost, line 34: You could just set `ctx` here instead instead of introducing `chCtx`. The rest of the function is supposed to use one of these only anyways, so just avoid the confusion of having two different context variables to choose from.

internal/incident/incidents.go
```diff
@@ -25,38 +27,122 @@ var (
...
+					incidentIds := make([]int64, chunkLen)
+					incidentsById := make(map[int64]*Incident, chunkLen)
+					incidentsByObjId := make(map[string]*Incident, chunkLen)
+
+					for _, i := range bulk {
+						incidentsById[i.ID()] = i
+						incidentsByObjId[i.ObjectID.String()] = i
+
+						objectIds = append(objectIds, i.ObjectID)
+						incidentIds = append(incidentIds, i.ID())
```
julianbrost, line 89: You're initializing the slices to that `len(objectIds) = len(incidentIds) = len(bulk)` and then append to the first two. I.e. those will end up having `len(objectIds) = len(incidentIds) = 2 * len(bulk)` in the end where the first half is filled with zero values.

internal/incident/incidents.go
```diff
@@ -69,37 +155,14 @@ func GetCurrent(
...
-			currentIncident = incident
-		}
-
-		if create && currentIncident == nil {
-			created = true
-			currentIncident = incident
-		}
+		incidentLogger := logger.With(zap.String("object", obj.DisplayName()))
+		currentIncident = NewIncident(db, obj, runtimeConfig, incidentLogger)
+		currentIncident.ObjectID = obj.ID
```
julianbrost, line 163: If the object is already passed to `NewIncident()` already, that function should take care of initializing the `ObjectID` field.

julianbrost (changes requested):

julianbrost:
@oxzi FYI: I don't expect any big structural changes to this PR anymore, so feel free to have a look as well.

internal/incident/incidents.go
yhabteab: 🤦‍♂️ !

internal/incident/incidents.go
```diff
@@ -25,38 +27,122 @@ var (
...
+					incidentsById := make(map[int64]*Incident, chunkLen)
+					incidentsByObjId := make(map[string]*Incident, chunkLen)
+
+					for k, i := range bulk {
+						incidentsById[i.Id] = i
+						incidentsByObjId[i.ObjectID.String()] = i
+
+						objectIds[k] = i.ObjectID
+						incidentIds[k] = i.Id
+					}
```
julianbrost, line 90: Those loop variable are somewhat unexpected/confusing. I'd just go for `for i, incident := range bulk`. (Or you could initialize the slices as `make([]T, 0, chunkLen)` and keep using `append()`.)

julianbrost (changes requested):

julianbrost:
FWIW: if someone else tries to use `/dump-incidents` with this PR and runs into errors, here's a crude patch to work around this:

<details>
<summary>Patch</summary>

```diff
diff --git a/internal/listener/listener.go b/internal/listener/listener.go
index ad626fb..d0f4ad3 100644
--- a/internal/listener/listener.go
+++ b/internal/listener/listener.go
@@ -12,6 +12,7 @@ import (
        "github.com/icinga/icinga-notifications/internal/incident"
        "github.com/icinga/icingadb/pkg/icingadb"
        "github.com/icinga/icingadb/pkg/logging"
+       "github.com/icinga/icingadb/pkg/types"
        "go.uber.org/zap"
        "net/http"
        "time"
@@ -218,6 +219,12 @@ func (l *Listener) DumpIncidents(w http.ResponseWriter, r *http.Request) {
                incident.Lock()
                defer incident.Unlock()
 
+               if incident.RecoveredAt.Time().IsZero() {
+                       savedRecoveredAt := incident.RecoveredAt
+                       incident.RecoveredAt = types.UnixMilli(time.Unix(0, 0))
+                       defer func() { incident.RecoveredAt = savedRecoveredAt }()
+               }
+
                encoded, err := json.Marshal(incident)
                if err != nil {
                        encoded, err = json.Marshal(err.Error())
```
</details>

The actual cause is already fixed by https://github.com/Icinga/icingadb/pull/612 but we're currently branched of from an older version, it's definitely time for #114 :see_no_evil:

oxzi (commented):
The code change looks fine, so far. Furthermore, I have tested the changed version.

yhabteab:
Actually, my just written unit tests were able to catch a serious bug 🐞, `utils#ForEachRow()` acquires a table semaphore, however I forgot to release it when returning and causes the tests to hang endlessly 🙈.

tests/sql/restore_incidents_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+func makeRandomString(t *testing.T) string {
+	buf := make([]byte, 20)
+	_, err := rand.Read(buf)
+	require.NoError(t, err, "failed to generate random string")
+
+	return fmt.Sprintf("%x", buf)
+}
+
+func getDBConfig() (*testConfig, error) {
```
julianbrost, line 200: You could use a function with that signature instead:

```go
func getTestDB(t testing.TB) *icingadb.DB {
	// ...
}
```

That function could then just do the error handling internally and just terminate the test using `t.Fatal()` (or using a corresponding wrapper from `testify`).

Also, it could use `t.Skip()` if no database config was provided in the environment. Allowing it to simply be used more like a standard unit test, i.e. the actual test code be moved to `internal/incident/*_test.go` like it would usually be done with tests in Go.

tests/sql/restore_incidents_test.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	db, err := c.Database.Open(logging.NewLogger(zap.NewNop().Sugar(), time.Hour))
+	require.NoError(t, err, "connecting to database should not fail")
+
+	ctx := context.Background()
+	if c.Schema != "" { // Import schema when requested to do it!
+		_, err = db.ExecContext(ctx, c.Schema)
+		require.NoError(t, err, "schema import should not fail")
+	}
```
julianbrost: I think I'd prefer importing the schema on the outside, i.e. in the GHA workflow definition. With PostgreSQL this is probably fine, but with MySQL, this will likely become annoying (see the workarounds we do in `icinga-testing` as the `mysql` CLI client parses a schema file differently than the MySQL server parses statements sent to it).

tests/sql/restore_incidents_test.go
```diff
@@ -0,0 +1,254 @@
...
+	"go.uber.org/zap"
+	"os"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+)
+
+func TestIncidentLoading(t *testing.T) {
+	t.Parallel()
```
julianbrost, line 28: You should be careful with `t.Parallel()` here. If the test uses a shared database, you might not want other tests to run at the same time.

tests/sql/restore_incidents_test.go
yhabteab: However, when using Github Actions for this, I have to somehow perform `\c DATABASE` command before importing the schema, something I have not yet managed to do, so I've added this fallback.

julianbrost: Shouldn't it just do the right thing it you set `POSTGRES_DB` and put the schema into something like `/docker-entrypoint-initdb.d/schema.sql`?

yhabteab: No, it doesn't seem to be the case!
https://github.com/Icinga/icinga-notifications/actions/runs/8734094424/job/23964046786#step:2:143

julianbrost: > psql:/docker-entrypoint-initdb.d/notifications.sql: error: could not read from input file: Is a directory

Looks like something went wrong with mounting the schema file into the container?

tests/sql/restore_incidents_test.go
yhabteab: > Also, it could use t.Skip() if no database config was provided in the environment.

And how should I exclude the other unit tests from those that require a database connection? Otherwise the test would get mixed up.

tests/sql/restore_incidents_test.go
yhabteab: Didn't get that log to see, but yeah, I wasn't aware that `github.workspace` is just an empty directory until the `checkout` step gets executed.

tests/sql/restore_incidents_test.go
julianbrost: To be honest, I simply wouldn't care if the "simple" (i.e. not interacting with a database) test cases are just run multiple times. We don't have any super expensive tests, so I don't think there's a need to put effort into this. If an expensive test was added in the future, that could simply add another flag to enable it that's then only set in one run of the tests.

yhabteab:
The tests are finally running properly on GitHub as well 😀!

internal/incident/restore_incidents_test.go
```diff
@@ -0,0 +1,238 @@
...
+// invalid configurations are specified.
+func getTestDB(ctx context.Context, t *testing.T) *icingadb.DB {
+	if _, ok := os.LookupEnv("NOTIFICATIONS_TESTS_DB_TYPE"); !ok {
+		t.Skipf("Environment %q not set, skipping test!", "NOTIFICATIONS_TESTS_DB_TYPE")
+	}
+
+	c := &baseConfig.Database{}
+	if err := defaults.Set(c); err != nil {
+		t.Skipf("Skipping test due to an error: %s", err.Error())
+	}
```
julianbrost, line 195: That call should pretty much never fail, thus, I'd just `require.NoErr()` here.

internal/incident/restore_incidents_test.go
```diff
@@ -0,0 +1,238 @@
...
+			c.Host = v
+		case "NOTIFICATIONS_TESTS_DB_PORT":
+			port, err := strconv.Atoi(v)
+			if err != nil {
+				require.NoError(t, err, "invalid port provided")
+			}
+
+			c.Port = port
+		}
+	}
```
julianbrost, line 227: I don't really see a benefit of the loop here and it just repeats the variable names.

I'd just chain a bunch of regular old `if`s:

```go
if v, ok := os.LookupEnv("NOTIFICATIONS_TESTS_DB_TYPE"); ok {
	c.Type = strings.ToLower(v)
} else {
	t.Skip(...)
}
```

That one just combines the one from the top into here, the others won't even need an else branch, so it shouldn't even make the code longer.

internal/incident/restore_incidents_test.go
```diff
@@ -0,0 +1,238 @@
...
+			c.User = v
+		case "NOTIFICATIONS_TESTS_DB_PASSWORD":
+			c.Password = v
+		case "NOTIFICATIONS_TESTS_DB_HOST":
+			c.Host = v
+		case "NOTIFICATIONS_TESTS_DB_PORT":
+			port, err := strconv.Atoi(v)
+			if err != nil {
+				require.NoError(t, err, "invalid port provided")
+			}
```
julianbrost, line 223: `NoError()` already does that `if` for you.

internal/incident/restore_incidents_test.go
```diff
@@ -0,0 +1,238 @@
...
+
+			c.Port = port
+		}
+	}
+
+	if err := c.Validate(); err != nil {
+		require.NoError(t, err, "database config validation should not fail")
+	}
```
julianbrost, line 231: Also here, no need for the `if` with `NoError()`.

internal/incident/restore_incidents_test.go
```diff
@@ -0,0 +1,238 @@
...
+
+	db, err := c.Open(logging.NewLogger(zap.NewNop().Sugar(), time.Hour))
```
julianbrost, line 233: `zap` allows to create a logger for tests, have you considered using it? https://pkg.go.dev/go.uber.org/zap/zaptest#NewLogger

internal/incident/restore_incidents_test.go
```diff
@@ -0,0 +1,238 @@
...
+//
+// This will firstly create and synchronise a new object from a freshly generated dummy event with distinct
+// tags and name, and ensures that no error is returned, otherwise it will cause the entire test suite to fail.
+// Once the object has been successfully synchronised, an incident is created and synced with the database.
+func makeIncident(ctx context.Context, db *icingadb.DB, t *testing.T, recovered bool) *Incident {
+	ev := &event.Event{
+		Time:     time.Time{},
+		SourceId: 1,
+		Name:     makeRandomString(t),
+		Tags: map[string]string{ // Always generate unique object tags not to produce same sha1!
```
julianbrost, line 146: We're not using SHA1, but that's also an irrelevant detail here, just say ID instead.

internal/incident/restore_incidents_test.go
```diff
@@ -0,0 +1,238 @@
...
+			"hostgroup/database-server": "",
+			"servicegroup/webserver":    "",
+		},
+	}
+
+	o, err := object.FromEvent(ctx, db, ev)
+	require.NoError(t, err)
+
+	i := NewIncident(db, o, &config.RuntimeConfig{}, nil)
+	i.StartedAt = types.UnixMilli(time.Unix(time.Now().Add(-2*time.Hour).Unix(), 0))
```
julianbrost, line 160: Why not just use `types.UnixMilli(time.Now().Add(-2*time.Hour))` here?

julianbrost (changes requested):
> Actually, my just written unit tests were able to catch a serious bug 🐞

Well, I'd say they already paid off :sweat_smile:

internal/incident/restore_incidents_test.go
yhabteab: NVM! It's already there as part of `zap` main package.

internal/incident/restore_incidents_test.go
yhabteab: `types.Scan()` uses `utils#FromUnixMilli()`, which in turn uses `time.Unix()`, and `time.Now()` is just a local time, thus the loaded incidents won't have the same timestamp.

julianbrost: What exactly is the difference? Is it timezone information and/or sub-second precision? It might be possible to achieve the same by calling [`UTC()`](https://pkg.go.dev/time#Time.UTC) and/or [`Truncate(time.Second)`](https://pkg.go.dev/time#Time.Truncate) to achieve the same in a more obvious way.

yhabteab:
```go
expected: types.UnixMilli{wall:0xc1805b940a3e40e0, ext:-7199881851166, loc:(*time.Location)(0x100e1ffa0)}
actual  : types.UnixMilli{wall:0xa3140c3, ext:63849033168, loc:(*time.Location)(0x100e1ffa0)}
```

yhabteab: `Truncate(time.Second)` fixes the tests.

internal/incident/restore_incidents_test.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	require.NoError(t, c.Validate(), "database config validation should not fail")
+
+	db, err := c.Open(logging.NewLogger(zaptest.NewLogger(t).Sugar(), time.Hour))
+	require.NoError(t, err, "connecting to database should not fail")
+	require.NoError(t, db.PingContext(ctx), "pinging the database should not fail")
+
+	return db
+}
```
julianbrost, line 223: Please move this somewhere where it can be used by other test cases in the future, this will then become a useful part of #130.

After reading https://github.com/golang/go/issues/12120, the best way to do so seems to create a new package that's only imported from tests (and place it in a regular file, i.e. *not* `*_test.go`). I'd suggest `internal/testutils/`.

internal/incident/restore_incidents_test.go
```diff
@@ -0,0 +1,223 @@
...
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap/zaptest"
+	"os"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+)
+
+func TestIncidentLoading(t *testing.T) {
```
julianbrost, line 25: This basically is a test for the `LoadOpenIncidents()` function, therefor, please name it `TestLoadOpenIncidents()` and also simply name this file `incidents_test.go` so that it corresponds to the file name the tested function is in.

internal/object/object.go
```diff
@@ -105,11 +163,6 @@ func FromEvent(ctx context.Context, db *icingadb.DB, ev *event.Event) (*Object,
 		return nil, fmt.Errorf("can't commit object database transaction: %w", err)
 	}
 
-	object.ExtraTags = ev.ExtraTags
-	object.Tags = ev.Tags
-	object.Name = ev.Name
-	object.URL = ev.URL
```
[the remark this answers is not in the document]
yhabteab: You are right! I should have removed only `object.Tags = ev.Tags` as its change always means a new object.

internal/incident/incidents.go
```diff
@@ -25,38 +27,122 @@ var (
...
+
+						i.RetriggerEscalations(&event.Event{
+							Time:    time.Now(),
+							Type:    event.TypeInternal,
+							Message: "Incident reevaluation at daemon startup",
+						})
+
+						currentIncidentsMu.Lock()
+						currentIncidents[i.Object] = i
+						currentIncidentsMu.Unlock()
```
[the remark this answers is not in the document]
yhabteab: Makes no difference to me, but I'm not going to waste time trying to reason about, so I'll change it!

.github/workflows/compliance.yml
```diff
@@ -31,4 +31,5 @@ jobs:
         # and this very icinga-notifications are licensed under GPL-2.0.
         # https://github.com/Icinga/icingadb/blob/v1.1.1/.github/workflows/compliance/check-licenses.sh
         go-licenses check github.com/icinga/icinga-notifications/... \
-          --allowed_licenses BSD-2-Clause,BSD-3-Clause,GPL-2.0,MIT,MPL-2.0
+          --allowed_licenses BSD-2-Clause,BSD-3-Clause,GPL-2.0,MIT,MPL-2.0 \
+          --ignore github.com/davecgh/go-spew/spew
```
julianbrost, line 35: > Since we now utilise `testing.T, require.**` in a file that does not end with `*_test.go`, the go licence compliance failed due to the spew package, so I had to put this on the ignore list.

Please don't do that, that would prevent us from noticing possible future license changes of that module and also from accidentally using it in non-test code in the future.

If there was indeed a license problem, it would be better to just avoid the package in this case. However, [it's licensed under the ISC license](https://github.com/davecgh/go-spew/blob/master/LICENSE) which [is compatible with the GPL](https://www.gnu.org/licenses/license-list.en.html#ISC). As that's pretty much the only reason for this check, it can simply be added to the allowed licenses.

julianbrost (changes requested):

.github/workflows/compliance.yml
yhabteab: Done!

julianbrost (approved):

PR #132, opened 2023-11-22
Multiple Source in Conf, Listener authentication, Populate source table on startup

oxzi:
*    Multiple Source in Conf, Listener authentication

    This commit brings two connected changes.

    First, the YAML configuration might now holds multiple sources.

    Second, domain knowledge was moved into the listener. Next to some
    generic error logging refactoring, it checks for each submitted event if
    it refers a known source, is allowed to be submitted and, as a further
    change, adds HTTP "Basic" authentication.
*    Populate source table on startup

    Make the manual INSERT statement into the source table obsolete by
    upserting data from the YAML configuration file.


julianbrost (commented):
Basically, this was my first idea for a somewhat common source configuration considering the options needed in #112. However, on second thought, syncing sources from the config file to the database is exactly the opposite of how most other config works. So the consistent way would probably be to the `source` table in the database with all the options needed and load them from there. But then, the web module should probably provide proper config forms, including a special one for Icinga 2 with #112, so we should probably discuss this with @nilmerg next week. (Below are the things I noticed so far, but I'd postpone this PR until we know how we actually want to do it.)

internal/listener/listener.go
```diff
@@ -79,42 +79,59 @@ func (l *Listener) Run(ctx context.Context) error {
...
+	}
+
+	user, pass, _ := req.BasicAuth()
+	userOk := subtle.ConstantTimeCompare([]byte(source.Authentication.User), []byte(user)) == 1
+	passOk := subtle.ConstantTimeCompare([]byte(source.Authentication.Pass), []byte(pass)) == 1
+	if !userOk || !passOk {
+		w.Header().Set("WWW-Authenticate", fmt.Sprintf(`Basic realm="source-%d"`, source.Id))
+		abort(http.StatusUnauthorized, ev, "unauthorized for source %d", source.Id)
+		return
+	}
```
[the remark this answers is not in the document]
oxzi: Yes, my train of thoughts was to keep the possibility of an unauthorized API open. Especially as the body needs to be parsed first to know the Source ID to get the required credentials, I placed the code as one unit.

I have now changed the code to check for HTTP basic authentication before parsing the body and comparing the passed credentials against the source's credentials later on.

oxzi:
@julianbrost: According to your feedback and our off-band communication about this feature, I have reversed this PR's logic to store all required data in the SQL database and load it into the `RuntimeConfig`. Please take another look.

julianbrost (changes requested):

schema/pgsql/schema.sql
```diff
@@ -129,6 +129,14 @@ CREATE TABLE source (
     -- will likely need a distinguishing value for multiple sources of the same type in the future, like for example
     -- the Icinga DB environment ID for Icinga 2 sources
 
+    -- listener_password_hash is required to limit API access for incoming connections to the Listener. The username is
+    -- "source-${id}", allowing an early verification before having to parse the POSTed event.
+    -- The hash is PHP's password_hash with the PASSWORD_DEFAULT algorithm, as already being used in Icinga Web 2.
+    -- https://icinga.com/docs/icinga-web/latest/doc/20-Advanced-Topics/#manual-user-creation-for-database-authentication-backend
```
julianbrost, line 135: Wouldn't it be more helpful to say that this must be a bcrypt hash starting with `$2y$`? I mean not sure how often PHP changes `PASSWORD_DEFAULT`, but when the web module writes to this column, it would be a good idea to explicitly specify bcrypt when hashing the value.

internal/config/runtime.go
```diff
@@ -137,6 +142,51 @@ func (r *RuntimeConfig) GetContact(username string) *recipient.Contact {
...
+// cases a debug logging message, except when something server-side is wrong, e.g., the hash is invalid.
+func (r *RuntimeConfig) GetSourceFromCredentials(user, pass string, logger *logging.Logger) *Source {
+	r.RLock()
+	defer r.RUnlock()
+
+	usernameMatches := getSourceFromCredentialsUsernameRegexp.FindStringSubmatch(user)
+	if usernameMatches == nil {
+		logger.Debugw("Cannot extract source ID from HTTP basic auth username", zap.String("user-input", user))
+		return nil
+	}
```
julianbrost, line 163: [`strings.CutPrefix()`](https://pkg.go.dev/strings#CutPrefix) should be much simpler here. Also, regarding negative numbers: it ends up as a map lookup, that doesn't care if `r.Sources[404]` or `r.Sources[-404]` is not within the map, so that shouldn't need any special treatment.

internal/listener/listener.go
```diff
@@ -79,42 +79,69 @@ func (l *Listener) Run(ctx context.Context) error {
...
-		_, _ = fmt.Fprintf(w, "cannot parse JSON body: %v\n", err)
+		abort(http.StatusBadRequest, nil, "cannot parse JSON body: %v", err)
 		return
 	}
 	ev.Time = time.Now()
 
+	if ev.SourceId != source.ID {
+		abort(http.StatusBadRequest, &ev, "authenticated for source %d with an event for %d", source.ID, ev.SourceId)
+		return
+	}
```
julianbrost, line 124: It's redundant that the client sends the source ID in two places now. This could just do `ev.SourceId = source.ID` here. Ideally, we'd also disallow setting the from JSON so that we know that if it's set, it can be trusted.

julianbrost (changes requested):

schema/pgsql/schema.sql
oxzi: I like this idea. Now it is double checked during DB insertion as well as during comparison by `golang.org/x/crypto/bcrypt`.

internal/listener/listener.go
oxzi: Yes, and due to this redundancy I have added this check to fail hard in case something got mixed up. Otherwise, this would open doors to mitigate the Source ID-related authentication.

Thus, I have now doubled down on using the HTTP authentication and no longer cared for the JSON-deserialized `SourceId` field.

oxzi:
Bumped the `schema/pgsql/upgrades/` file number, as #136 just got approved.

internal/event/event.go
```diff
@@ -13,7 +13,7 @@ import (
 
 type Event struct {
 	Time     time.Time
-	SourceId int64 `json:"source_id"`
+	SourceId int64
```
julianbrost, line 16: IIRC you could still set this as `{"SourceId": 42}` this way, so you probably want to explicitly specify `json:"-"` now.

julianbrost (changes requested):
> Bumped the `schema/pgsql/upgrades/` file number, as #136 just got approved.

Which is now the next number with #105 merged. Please just keep that number as I'd wait with merging #136 as commented there, so this PR here is a prime candidate for being merged soon.

julianbrost (approved):

PR #133, opened 2023-11-22
Consistent logging

### LINKED ISSUE #115: Consistently wrap errors
During the review of #99, I noticed there's already quite a bit of error handling in [`internal/incident/incident.go`](https://github.com/Icinga/icinga-notifications/blame/main/internal/incident/incident.go) that returns entirely new errors instead of wrapping the errors that caused that error condition.

I don't see a reason why we wouldn't want to wrap errors there, but maybe @yhabteab does (pinging you as you wrote large parts of that code). If not, we should go over that code and adapt the error handling (I only noticed it in this file so far, but scrolling over all uses of `errors.New()` and `fmt.Errorf()` without `%w` probably doesn't hurt).

yhabteab:
resolves #115


yhabteab:
@julianbrost do you still want to return wrapped errors everywhere even though they're already logged, or is it fine the way it is? And as for the HTTP response messages, Alvar has already cleaned them up with https://github.com/Icinga/icinga-notifications/pull/132, so I haven't done anything in that regard here.

julianbrost:
In general, I'd try to keep close to typical Go conventions regarding the returned errors. So that would be wrapping the errors even if we basically dismiss the message later on. Note that wrapping the error also isn't just about the message but also allows the proper use of functions like `errors.Is()`.

oxzi:
Maybe we could use this PR to also get rid of contractions. Currently, there is both "can't" as well as "cannot" in the code, and other contraction forms. A (most probably incomplete) `grep -ri "\b[a-z]*'[a-z]*\b"` also shows "doesn't", "it's", and even more.

This might not be part of the referenced issue, but fits the PR title very well.

internal/listener/listener.go
```diff
@@ -122,7 +122,7 @@ func (l *Listener) ProcessEvent(w http.ResponseWriter, req *http.Request) {
 	ctx := context.Background()
 	obj, err := object.FromEvent(ctx, l.db, &ev)
 	if err != nil {
-		l.logger.Errorw("Can't sync object", zap.Error(err))
+		l.logger.Errorw("Cannott sync object", zap.Error(err))
```
oxzi, line 125:
```suggestion
		l.logger.Errorw("Cannot sync object", zap.Error(err))
```

oxzi (commented):
Within code - excluding comments - contractions there is still one _"doesn't"_ in `internal/filter/parser_test.go` on line 144.

Otherwise, next to the two small comments, the diff looks good to me.

internal/incident/incident.go
```diff
@@ -134,9 +131,8 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event, created bo
 	}
 
 	if err = i.AddEvent(ctx, tx, ev); err != nil {
-		i.logger.Errorw("Can't insert incident event to the database", zap.Error(err))
-
-		return errors.New("can't insert incident event to the database")
+		i.logger.Errorw("Cannott insert incident event to the database", zap.Error(err))
```
oxzi, line 134:
```suggestion
		i.logger.Errorw("Cannot insert incident event to the database", zap.Error(err))
```

oxzi (commented):
Sorry, I have missed this typo the other time.

Otherwise, looks good to me.

oxzi (dismissed):

yhabteab:
Sorry! I had to adjust the time period entry a bit.

oxzi (dismissed):

julianbrost:
I'm surprised by the number of times where `return errors.New("[...]")` is replaced by just `return err` without wrapping it with some extra context information. Is that extra error message that was used there always redundant with what's already contained in `err` or in other ways pointless to add?

yhabteab:
> Is that extra error message that was used there always redundant with what's already contained in `err` or in other ways pointless to add?

That extra error message was supposed to be used by the listener as an HTTP response message, but since we don't want to expose such specific internal error messages to the clients, there is no need in keeping those extra error messages as they're always logged with the required additional contexts.

julianbrost (commented):
I don't really see the reason for rewording many of the messages. What rules are you trying to follow here or what are you trying to be consistent with? It's just a lot of places were the meaning of the messages is changed slightly and one has to check if the new message still describes what's happening. (I haven't looked at everything yet as I don't really get the point so far.)

internal/icinga2/client.go
```diff
@@ -240,21 +240,21 @@ func (client *Client) startCatchupWorkers(delay time.Duration) (chan *catchupEve
...
-			client.Logger.Debugw("Catching up the API has finished",
+			client.Logger.Debugw("Finished catching up the API event streams",
 				zap.Stringer("worker", workerId),
 				zap.Duration("duration", time.Since(startTime)))
 		} else if errors.Is(err, context.Canceled) {
 			// The context is either canceled when the Client got canceled or, more likely, when another catch-up-worker
 			// was requested. In the first case, the already sent messages will be discarded as the worker's main loop
 			// was left. In the other case, the message buffers will be reset to an empty state.
-			client.Logger.Debugw("Catching up the API was interrupted",
+			client.Logger.Debugw("Interrupted catching up the API event streams",
```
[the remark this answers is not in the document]
yhabteab: reverted!

yhabteab:
> I don't really see the reason for rewording many of the messages. What rules are you trying to follow here or what are you trying to be consistent with? It's just a lot of places were the meaning of the messages is changed slightly and one has to check if the new message still describes what's happening. (I haven't looked at everything yet as I don't really get the point so far.)

It's used to be some Icinga Web notification rule `state (success/fail) -> action (started|to start) -> subject`, but I've reverted all the changes now, so I don't want to wast that much time in this PR anymore.

oxzi (dismissed):

julianbrost (changes requested):
https://github.com/Icinga/icinga-notifications/pull/133#issuecomment-1849638204:
> but since we don't want to expose such specific internal error messages to the clients

https://github.com/Icinga/icinga-notifications/pull/133#pullrequestreview-1759388937:

> I'm surprised by the number of times where `return errors.New("[...]")` is replaced by just `return err` without wrapping it with some extra context information.

If you do replace these basic errors with the full internal errors, they must not be sent to the HTTP client anymore. Instead, the clients should just receive some generic "internal error, see server log".

internal/listener/listener.go
```diff
@@ -86,13 +85,17 @@ func (l *Listener) ProcessEvent(w http.ResponseWriter, req *http.Request) {
...
 			logger = logger.With(zap.Stringer("event", ev))
 		}
 
+		if statusCode >= http.StatusInternalServerError {
+			msg = event.ErrEventProcessing.Error()
+		}
+
 		http.Error(w, msg, statusCode)
-		logger.Debugw("Abort listener submitted event processing")
+		logger.Error("Failed to successfully process event")
```
julianbrost, line 98: I'd do this redaction of the error outside of this function. Then you don't have to rely on the status code here. Additionally, the way you do the logging now, this would log something like "POST required" at the error level which doesn't sound very useful to me, thus I'd also leave this at debug level here and instead add an error log for that specific error case where this function is called.

internal/event/event.go
```diff
@@ -12,6 +12,9 @@ import (
 	"time"
 )
 
+// ErrEventProcessing is a general event processing error message.
+var ErrEventProcessing = errors.New("event could not be processed successfully, see server logs for details")
```
julianbrost, line 16: This is never actually used as an `error`. For me, this is more similar to the other response messages like "POST required", thus I'd also just put it there in `listener.go`.

julianbrost (changes requested):

PR #134, opened 2023-11-23
channels: Use `incident.severity` instead of `event.severity`

sukhwinder33445:


julianbrost (dismissed):
Setting the severity in the `TypeInternal` types as added in 1acb259bb4b324bf59da05cb04bd444155830d4b should no longe be necessary with this change. Please remove it there and check if the notifications still work as expected.

oxzi (approved):

PR #137, opened 2023-11-28
Enhance notification types

### LINKED ISSUE #106: Additional notifications: notifications without an incident + incident meta notifications
At the moment, pretty much only state notifications are sent. We also have to allow sending other notifications.

For example, in case of Icinga 2, the available notification types are:
 * DowntimeStart
 * DowntimeEnd
 * DowntimeRemoved
 * Custom
 * Acknowledgement
 * Problem
 * Recovery
 * FlappingStart
 * FlappingEnd

Also, there are events within an incident that could be worth an extra notification, for example:

* Incident (un)managed / manager(s) changed
* Incident escalated

Suggestion for moving forward (1. should be pretty straight forward, the others probably need a bit more discussion about the details)

- [x] Just send all of these non-state notifications to all incident recipients. As the very first step without any new configuration options. If there's an open incident, it sounds like a reasonable default to get notified about everything that's related to the incident if someone is involved in the incident.
- [ ] It should also be possible to receive those notifications without an incident. For this, an option would be to have something like second part in the rules next to (on a logical level, not on the representation in the rule editor) the escalations where one can configure which event types should be routed to which recipients.
- [ ] There should probably be the possibility to limit the types from 1., maybe with something like allowing to exclude certain types in the event rule per recipient.

sukhwinder33445:
resolves #106

### DIFF
--- a/internal/event/event.go
+++ b/internal/event/event.go
@@ -34,10 +34,48 @@ type Event struct {
 }
 
 const (
-	TypeState           = "state"
-	TypeAcknowledgement = "acknowledgement"
-	TypeInternal        = "internal"
+	TypeState             = "state"
+	TypeAcknowledgement   = "acknowledgement"
+	TypeCommentAdded      = "commentAdded"
+	TypeCommentRemoved    = "commentRemoved"
+	TypeDowntimeStart     = "downtimeStart"
+	TypeDowntimeEnd       = "downtimeEnd"
+	TypeDowntimeCancelled = "downtimeCancelled"
+	TypeCustom            = "custom"
+	TypeFlappingStart     = "flappingStart"
+	TypeFlappingEnd       = "flappingEnd"
+	TypeInternal          = "internal"
 )
+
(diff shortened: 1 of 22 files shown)


internal/event/event.go
```diff
@@ -34,11 +34,49 @@ type Event struct {
...
+		case TypeAcknowledgement:
+		case TypeCommentAdded:
+		case TypeCommentRemoved:
+		case TypeDowntimeStart:
+		case TypeDowntimeEnd:
+		case TypeDowntimeCancelled:
+		case TypeCustom:
+		case TypeFlappingStart:
+		case TypeFlappingEnd:
+			return nil
```
julianbrost, line 70: There's no automatic fallthrough between cases, so this doesn't really do what you probably expect (or in some way it does, the next statement after `switch` will be the `return nil` at the end of the function, so you won't notice a difference here.

Hint: you can do `case A, B, C:`

yhabteab:
I'll take over this PR from now on!

PR #138, opened 2023-11-29
schema: Drop column `schedule_member.membership_hash`

nilmerg:
It's not needed by Web anymore.


oxzi (approved):
After being introduced in 2e363d7006608da2eb09f29fe1d006ae3eb48447 (#87), it seems to be nowhere referenced within the Go code:
https://github.com/Icinga/icinga-notifications/blob/e89bbdccf3253e136594c0c066f8884b881b0633/internal/recipient/schedule.go#L22

julianbrost:
Yup, that column was intended purely for the web module, similar to `contact.color` for example.

PR #140, opened 2023-12-06
GitHub Actions: Check Go Dependency Licenses

### LINKED ISSUE #139: Track Dependency Licenses
We should also be able to track all dependency licenses, similar to icingadb, <https://github.com/Icinga/icingadb/tree/main/.github/workflows>. As an alternative to an home-brew solution, maybe checkout <https://github.com/google/go-licenses> or something else.

oxzi:
By utilizing the neat go-licenses[^0] tool, scanning the cached Go dependencies against an allow list of licenses, which is currently leaned from Icinga DB, works quite like a charm.

This, however, only includes Go code and produces warnings for (transitive) included Go Assembly code[^1]. If we are planning to include other non-Go artefacts in the future, those also might need to be identified - REUSE[^2] might help there.

Closes #139.

For an example, take a look at:
- f944c7a751efaebfea155f2b7d401ec5de91d630, which was a successfully run or
- e04222074127af44567da8aba2e168bb7eb8b0ba, where `MPL-2.0` was missing in the allowed licenses, resulting in an adequate error of:
  > Not allowed license MPL-2.0 found for library github.com/go-sql-driver/mysql 

[^0]: https://github.com/google/go-licenses
[^1]: https://github.com/google/go-licenses/issues/120
[^2]: https://reuse.software/

### DIFF
--- a/.github/workflows/compliance.yml
+++ b/.github/workflows/compliance.yml
@@ -1,0 +1,34 @@
+name: Compliance
+
+on:
+  push:
+    branches: [ main ]
+  pull_request: {}
+
+permissions:
+  # https://docs.github.com/en/rest/overview/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents
+  contents: read
+
+jobs:
+  compliance:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-go@v5
+      with:
+        go-version: stable
+
+    - name: Download modules to local cache
+      run: go mod download
+
+    - name: Install go-licenses
+      run: go install github.com/google/go-licenses@latest
+
+    - name: Check licenses against allow list
+      run: |
+        # Pass allowed licenses as SPDX Identifiers: https://spdx.org/licenses/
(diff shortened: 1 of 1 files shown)


.github/workflows/compliance.yml
```diff
@@ -0,0 +1,34 @@
...
+  push:
+    branches: [ main ]
+  pull_request: {}
+
+permissions:
+  # https://docs.github.com/en/rest/overview/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents
+  contents: read
+
+jobs:
+  go:
```
julianbrost, line 13: Conflicts with a job name from `go.yml`, which is a problem for branch protection rules on GitHub, see #113.

.github/workflows/compliance.yml
```diff
@@ -0,0 +1,34 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-go@v4
```
julianbrost, line 17: As I just merged #141, please also update this to `actions/setup-go@v4` (otherwise, we'd just get another dependabot PR soon).

julianbrost (changes requested):

julianbrost (approved):

PR #142, opened 2023-12-12
Plugin#ConfigOptions: Remove placeholder and use default instead

sukhwinder33445:


julianbrost (approved):
IMHO somewhat strange to not distinguish between the two, but if web wants it that way, fine fore me.

PR #146, opened 2024-01-08
Send non-state notifications for incident

sukhwinder33445:
Please note that this PR only enables sending non-state notifications for existing incidents and does not interpret the events in any way. The PR diffs appear to be somewhat large due to the changes to the event stream in the icinga2 package, though they're fairly trivial changes that should be relatively easy to fellow/review.


internal/icinga2/api_responses.go
```diff
@@ -91,6 +98,10 @@ type Downtime struct {
 	Service string `json:"service_name"`
 	Author  string `json:"author"`
 	Comment string `json:"comment"`
+
+	// RemoveTime is used to indicate whether a downtime was ended automatically or cancelled prematurely by a user.
+	// It is set to 0 for the former case, otherwise to the timestamp at which time has been cancelled.
```
oxzi, line 103: Still `0` or is it now the zero value?

internal/icinga2/client.go
```diff
@@ -176,19 +177,62 @@ func (client *Client) buildHostServiceEvent(ctx context.Context, result CheckRes
...
+	}
+
+	ev.Username = d.Author
+	ev.Message = d.Comment
+
+	return ev, nil
+}
+
+// buildFlappingEvent from the given fields.
+func (client *Client) buildFlappingEvent(ctx context.Context, host, service string, state, stateType int, isFlapping bool) (*event.Event, error) {
```
oxzi, line 222: Parameters `state` and `stateType` are unused.

oxzi (commented):
The following comments are based on a first scrolling over the diffs.

cmd/channel/rocketchat/main.go
```diff
@@ -23,7 +23,7 @@ func main() {
 
 func (ch *RocketChat) SendNotification(req *plugin.NotificationRequest) error {
 	var output bytes.Buffer
-	_, _ = fmt.Fprintf(&output, "[#%d] %s %s is %s\n\n", req.Incident.Id, req.Event.Type, req.Object.Name, req.Incident.Severity)
+	_, _ = fmt.Fprint(&output, plugin.FormatSubject(req))
 
 	plugin.FormatMessage(&output, req)
```
julianbrost, line 28: That one still needs the newlines in between, so they should be added here.

julianbrost (changes requested):

cmd/channel/rocketchat/main.go
```diff
@@ -23,7 +23,7 @@ func main() {
 
 func (ch *RocketChat) SendNotification(req *plugin.NotificationRequest) error {
 	var output bytes.Buffer
-	_, _ = fmt.Fprintf(&output, "[#%d] %s %s is %s\n\n", req.Incident.Id, req.Event.Type, req.Object.Name, req.Incident.Severity)
+	_, _ = fmt.Fprint(&output, plugin.FormatSubject(req))
```
oxzi, line 26: Thanks for [your changes](https://github.com/Icinga/icinga-notifications/compare/fd02071db871792456d9a5e02a49f2a0bb2d0019..fb863216e971e886b2d23893490a1466cf31af7c). But now this one misses newlines :upside_down_face:

oxzi: Duplicate of <https://github.com/Icinga/icinga-notifications/pull/146#discussion_r1576458904>, @julianbrost was a bit faster.

internal/icinga2/api_responses.go
```diff
@@ -10,7 +10,7 @@ import (
 // UnixFloat is a custom time.Time type for millisecond Unix timestamp, as used in Icinga 2's API.
 type UnixFloat time.Time
 
-// Time returns the time.Time of UnixFloat.
+// Time returns the time.Time of UnixFloat or zero value of time.Time.
```
julianbrost, line 13: That doesn't really add any information for me. After all, this function is only a glorified cast, is returns the zero value if the underlying `Time` is set to the zero value. I think the idea behind this might have been to say that JSON `0` is mapped to `Time{}`, but that would better fit into a comment for the `UnmarshalJSON` method (including a short explanation why it does so).

internal/event/event.go
```diff
@@ -34,11 +34,58 @@ type Event struct {
...
-	TypeState           = "state"
-	TypeAcknowledgement = "acknowledgement"
-	TypeInternal        = "internal"
+	TypeState                  = "state"
+	TypeAcknowledgementSet     = "acknowledgement-set"
+	TypeAcknowledgementCleared = "acknowledgement-cleared"
+	TypeInternal               = "internal"
+	TypeDowntimeRemoved        = "downtime-removed"
+	TypeDowntimeStarted        = "downtime-started"
+	TypeDowntimeTriggered      = "downtime-triggered"
```
julianbrost, line 43: started vs. triggered? There's no such differentiation within the [Icinga 2 notification config](https://icinga.com/docs/icinga-2/latest/doc/09-object-types/#notification). For me, that's just a weird implementation detail within Icinga 2 that happens to be exposed via the event stream, but I'd try to avoid further exposing both.

oxzi (commented):
I went another time about the diff, this time being a bit more pedantic. Furthermore, I have successfully tested the code.

julianbrost (changes requested):

internal/icinga2/api_responses.go
```diff
@@ -15,13 +15,22 @@ func (t *UnixFloat) Time() time.Time {
 	return time.Time(*t)
 }
 
+// UnmarshalJSON decodes the provided timestamp into the current UnixFloat instance.
+// If a JSON value `0/0.0` is provided, it will be mapped to time.Time{} and allows you to
+// subsequently perform a time.Time.IsZero() check on t, which would not be possible otherwise.
```
julianbrost, line 20: The condition mentioned here does not match the implementation.

internal/icinga2/client.go
```diff
@@ -176,19 +177,64 @@ func (client *Client) buildHostServiceEvent(ctx context.Context, result CheckRes
...
 	ev, err := client.buildCommonEvent(ctx, host, service)
 	if err != nil {
 		return nil, err
 	}
 
-	ev.Type = event.TypeAcknowledgement
+	switch evType {
+	case event.TypeAcknowledgementSet, event.TypeAcknowledgementCleared:
+	default:
+		return nil, fmt.Errorf("unsupporeted acknowledgement event type %q", evType)
```
julianbrost, line 189: Typo: unsupporeted

internal/icinga2/client.go
```diff
@@ -176,19 +177,64 @@ func (client *Client) buildHostServiceEvent(ctx context.Context, result CheckRes
...
+	ev.Type = evType
+	if ev.Type == event.TypeDowntimeRemoved && d.RemoveTime.Time().IsZero() {
+		ev.Type = event.TypeDowntimeEnd
+	}
+
+	ev.Username = d.Author
+	ev.Message = d.Comment
+
+	return ev, nil
+}
```
julianbrost, line 221: These two function could take a bool like `buildFlappingEvent()` instead of `evType`. Then you wouldn't need error handling for that parameter and also the IMHO slightly strange mapping of `TypeDowntimeRemoved` to `TypeDowntimeEnd` would look less strange.

julianbrost (changes requested):
One thing seems to be missing completely: at least some of these additional events need a check for missed events during the event stream catch-up phase.

yhabteab:
> One thing seems to be missing completely: at least some of these additional events need a check for missed events during the event stream catch-up phase.

What makes you think we actually need this now? Unlike state and ack events, we don't track which events/ downtimes, flapping have already been processed like Icinga 2 does, catching all checkable active downtimes will inevitably produce huge notification spams after each Icinga 2 or/and notification daemon reload. So I'd rather opt to miss some downtime events than overwhelming the recipients with huge downtime notifications!

julianbrost:
> What makes you think we actually need this now?

Basically that I saw the change within `checkMissedChanges()` in `client_api.go` and thought "shouldn't there be more here".

You're right, as long as we don't have the backing information, just sending notifications again and again makes no sense.

Another thing that just came to my mind where I don't immediately know what to do here: currently, this PR will happily send further state notifications while in downtime/flapping. At the moment, we don't have a mute incident functionality (which these events should probably trigger), so it might be an idea to ignore state events if in downtime/flapping, similar to how soft states are ignored.

Oh and one more thing: please have a look at the logging, especially with the default logging configuration, all you see is the following:
```
2024-04-26T09:33:00.235Z	INFO	incident	Notify contact "Icinga Admin" via "E-Mail" of type "email"	{"object": "master-1!keep-state", "incident": "#9765", "channel_id": 1}
2024-04-26T09:33:00.238Z	INFO	incident	Successfully sent a notification via channel plugin	{"object": "master-1!keep-state", "incident": "#9765", "type": "email", "contact": "Icinga Admin"}
```

At least, these should now also include the event type.

internal/icinga2/client_api.go
```diff
@@ -427,26 +438,56 @@ func (client *Client) listenEventStream() error {
...
+			if !respT.Downtime.IsFixed {
+				continue
+			}
+
+			ev, err = client.buildDowntimeEvent(client.Ctx, respT.Downtime, true)
+			evTime = respT.Timestamp.Time()
+		case *DowntimeTriggered:
+			if respT.Downtime.IsFixed {
+				continue
+			}
```
julianbrost, line 484: A comment on why these two `continue` are here would be highly appreciated, especially given that it's confusing and the reason is outside of this repository.

julianbrost (changes requested):
Two small changes and a question for @nilmerg, apart from that, there are still some limitations to this implementation but fixing this is out of scope for this PR, I've created another issue: #190

internal/event/event.go
```diff
@@ -34,11 +34,52 @@ type Event struct {
 }
 
 const (
-	TypeState           = "state"
-	TypeAcknowledgement = "acknowledgement"
-	TypeInternal        = "internal"
+	TypeState                  = "state"
+	TypeAcknowledgementSet     = "acknowledgement-set"
```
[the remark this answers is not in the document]
yhabteab: > I think there's a quite good chance that Web might be unhappy if acknowledgement changes, at least if that happens without prior announcement. @nilmerg (That's what ends up in the event.type column.)

And why is that? Web literally does nothing with that column!

internal/icinga2/client_api.go
yhabteab: Did you actually read the comments on that field?

internal/event/event.go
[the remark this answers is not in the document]
julianbrost: I haven't checked the source, @nilmerg said so in https://github.com/Icinga/icinga-notifications/issues/162 for exactly this column:

> There may be more added in the future. We need to make sure that there are no such surprises for Web anymore

internal/icinga2/client_api.go
julianbrost: Yes, but indeed not while looking over this file.

I'm be looking for something like "Match the behavior of Icinga 2 which notifies for fixed downtimes only in OnDowntimeStarted: https://github.com/Icinga/icinga2/blob/v2.14.2/lib/icinga/checkable.cpp#L39", that's also missing there.

internal/event/event.go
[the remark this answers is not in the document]
nilmerg: `event.type` is indeed not handled right now. There's no need for it as there are only state and acknowledge events and only the latter have no severity which is how they're differentiated. This is only relevant once this change is done and Web is adjusted accordingly.

julianbrost:
I got very confused when looking at this again last week because of acknowledgments. I think I got it by now but can you please explain how acknowledgements themselves and especially other events during an acknowledgement are handled and why?

yhabteab:
> I think I got it by now but can you please explain how acknowledgements themselves and especially other events during an acknowledgement are handled and why?

Didn't you explicitly tell me to just simply send notifications for everything in this PR and not to interpret events in any way if not already there? That's exactly what this PR is doing. So I don't know exactly what is confusing you, after all, acknowledgement events are not something new introduced with this PR. So far we've only handled the AckSet event, which only does one thing, promote the author of the Ack to Incident Manager, if that's not already the case, and nothing has changed with PR.

julianbrost:
The root of my confusion was whether the following two places should also consider the acknowledgement state of the checkable (as in Icinga 2, it behaves very similar to a downtime):

https://github.com/Icinga/icinga-notifications/blob/8c380a19ff156daab10dbd3dcc35939db885e8e0/internal/icinga2/client_api.go#L233-L247
https://github.com/Icinga/icinga-notifications/blob/8c380a19ff156daab10dbd3dcc35939db885e8e0/internal/icinga2/client_api.go#L441-L453

Ideally, the PR description should state how notifications behave in those different scenarios and how it did change compared to before.

If I got it correctly, the acknowledgements aren't considered in these places as acknowledgement set is already interpreted by the notifications daemon. And this is done with a different effect than it would have in Icinga 2, therefore it's also handled differently here.

yhabteab:
> Ideally, the PR description should state how notifications behave in those different scenarios and how it did change compared to before.

But this PR doesn't change anything to the previous behaviour, except just sending notifications for the new and ack events. I don't mind suppressing any other states/events when a checkable is acknowledged if you want to stick to the Icinga 2 behaviour, but that should be handled/fixed in its own/separate PR as it is not related to the changes made by this PR.

> If I got it correctly, the acknowledgements aren't considered in these places as acknowledgement set is already interpreted by the notifications daemon. And this is done with a different effect than it would have in Icinga 2, therefore it's also handled differently here.

I don't think that's the rationale for doing it that way, as there was no discussion of it at all in #112 and so I think that's a bug that should definitely be fixed.

julianbrost:
> > If I got it correctly, the acknowledgements aren't considered in these places as acknowledgement set is already interpreted by the notifications daemon. And this is done with a different effect than it would have in Icinga 2, therefore it's also handled differently here.
> 
> I don't think that's the rationale for doing it that way, as there was no discussion of it at all in #112 and so I think that's a bug that should definitely be fixed.

That's not coming from #112 but from the very beginning: the idea that an Icinga 2 acknowledgement declares the acknowledging person as the incident manager. But in the current situation, this feels strange, so there's a good chance that this will still change in some way.

internal/icinga2/client_api.go
```diff
@@ -230,6 +230,16 @@ func (client *Client) checkMissedChanges(ctx context.Context, objType string, ca
 			return fmt.Errorf("querying API delivered a wrong object type %q", objQueriesResult.Type)
 		}
 
+		// Downtime suppresses all kind of events on that checkable, so skip it if it's in downtime.
+		if objQueriesResult.Attrs.DowntimeDepth > 0 {
+			continue
```
oxzi, line 235: Maybe add a debug log entry similar to those in the same case within `(client *Client) listenEventStream`.

internal/icinga2/client_api.go
```diff
@@ -427,26 +438,55 @@ func (client *Client) listenEventStream() error {
...
-		// case *DowntimeRemoved:
-		// case *DowntimeStarted:
-		// case *DowntimeTriggered:
+		case *DowntimeRemoved:
+			ev, err = client.buildDowntimeEvent(client.Ctx, respT.Downtime, false)
+			evTime = respT.Timestamp.Time()
+		case *DowntimeStarted:
+			if !respT.Downtime.IsFixed {
+				// This may never happen, but Icinga 2 does the same thing, and we need to ignore the start
+				// event for flexible downtime, as there will definitely be a triggered event for it.
```
oxzi, line 472: As it "may never happen", a debug log message might be a good idea to catch exactly those bugs one thinks they shouldn't be possible to exist.

oxzi (approved):
I went over the diff multiple times and it looks good to me.

Maybe add some debug messages here and there, as I have commented.

internal/icinga2/client_api.go
```diff
@@ -427,26 +438,55 @@ func (client *Client) listenEventStream() error {
...
-			// Only process HARD states
+			// Ignore any state changes while in downtime
+			if respT.DowntimeDepth > 0 {
+				client.Logger.Debugf("Skipping state change, checkable is in downtime, %#v", respT)
+				continue
+			}
+
+			// Only process HARD states.
+			// The checkable might also be in the flapping state, but Icinga 2 doesn't expose "is_flapping" to the
+			// event streams for state change. So, the only thing we can do here is to ignore soft state changes.
```
julianbrost, line 449: I'm fully aware that you added this based on [my earlier suggestion](https://github.com/Icinga/icinga-notifications/pull/146#issuecomment-2079014352):

> Another thing that just came to my mind where I don't immediately know what to do here: currently, this PR will happily send further state notifications while in downtime/flapping. At the moment, we don't have a mute incident functionality (which these events should probably trigger), so it might be an idea to ignore state events if in downtime/flapping, similar to how soft states are ignored.

However, after having played with this implementation, I noticed that this introduces quite some quirks due simply discarding a state change: after a downtime ended, the incident may still have the wrong severity, or it might even not be opened/closed even though it should.

So yes, while sending state notifications in a downtime sounds obviously undesirable, it might actually be the lesser evil until we properly tackle incident muting (#190). At least "downtimes are currently ignored when sending state notifications" is easier to understand than "in these circumstances (that would have to be explained additionally), an incident may go completely out of sync with what's happening in the source".

julianbrost (changes requested):

internal/icinga2/api_responses.go
```diff
@@ -91,6 +100,16 @@ type Downtime struct {
 	Service string `json:"service_name"`
 	Author  string `json:"author"`
 	Comment string `json:"comment"`
+
+	// RemoveTime is used to indicate whether a downtime was ended automatically or cancelled prematurely by a user.
+	// It is set to zero time for the former case, otherwise to the timestamp at which time has been cancelled.
+	RemoveTime UnixFloat `json:"remove_time"`
```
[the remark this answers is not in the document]
yhabteab: > This makes the interface to the outside more obvious

Are you serious, which outside? That field is intended to be used within the package and is never supposed to be used elsewhere.

> (you don't need details from a comment within the struct definition)

I'm not sure what that's supposed to mean.

> (the comment in there states "Neither Icinga 2 nor Icinga DB support negative timestamps for downtime" but the type is elsewhere as well

Are you proposing all of these changes just because the comment mentions downtime? This is just an example of something that I know for sure will cause problems. IMHO, you are focusing on totally non-essential things, after all this code is only temporary as far as I know and we could usefully invest the time elsewhere.

internal/icinga2/client_api.go
yhabteab: Does this also include the catch-up phase or just this one? I am asking this because this is not the only place where the state changes are being ignored.

internal/icinga2/api_responses.go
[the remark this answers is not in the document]
julianbrost: > > This makes the interface to the outside more obvious
> 
> Are you serious, which outside? That field is intended to be used within the package and is never supposed to be used elsewhere.
>
> > (you don't need details from a comment within the struct definition)
> 
> I'm not sure what that's supposed to mean.

Outside of this type. The following line could be `else if d.WasCanceled()` instead, making the information from the comment for the member obvious from the type's interface:

https://github.com/Icinga/icinga-notifications/blob/8c380a19ff156daab10dbd3dcc35939db885e8e0/internal/icinga2/client.go#L208

> > (the comment in there states "Neither Icinga 2 nor Icinga DB support negative timestamps for downtime" but the type is elsewhere as well
> 
> Are you proposing all of these changes just because the comment mentions downtime? This is just an example of something that I know for sure will cause problems. IMHO, you are focusing on totally non-essential things, after all this code is only temporary as far as I know and we could usefully invest the time elsewhere.

I found the special-casing in the `UnixFloat` to be a bit strange from the beginning, i.e. that's not really something you'd expect from such a type:

https://github.com/Icinga/icinga-notifications/blob/8c380a19ff156daab10dbd3dcc35939db885e8e0/internal/icinga2/api_responses.go#L28-L32

And now I've got a better idea how to do this instead. Wasn't the whole point of this so that you can write `d.RemoveTime.Time().IsZero()` instead of `d.RemoveTime.Time().Equal(time.UnixMilli(0))` in the `else if` mentioned above?

Also, `UnixFloat` might be a candidate for being moved to our common Go library like `UnixMilli`.

internal/icinga2/client_api.go
julianbrost: In both places, they should stay consistent. (That's also another reason for postponing it to proper incident muting as currently these are inconsistent with flapping due due missing information in the event stream).

julianbrost (approved):

PR #148, opened 2024-01-09
Load ConfigFile from Environment Variables

oxzi:
Next to populating ConfigFile through default values and by a YAML configuration, environment variable support was added.

The code works by reflecting on the existing nested ConfigFile struct. The environment variable key is an underscore separated string of uppercase struct fields.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,11 @@ git clone https://github.com/Icinga/icinga-notifications.git
 ```
 
 Next, you need to provide a `config.yml` file, similar to the [example config](config.example.yml), for the daemon.
+It is also possible to set environment variables by name instead of or in addition to the configuration file.
+The environment variable key is an underscore separated string of uppercase struct fields. For example
+* `ICINGA_NOTIFICATIONS_LISTEN` sets `ConfigFile.Listen` and
+* `ICINGA_NOTIFICATIONS_DATABASE_HOST` sets `ConfigFile.Database.Host`.
+
 It is required that you have created a new database and imported the [schema](schema/pgsql/schema.sql) file beforehand.
 > **Note**
 > At the moment **PostgreSQL** is the only database backend we support.
--- a/cmd/icinga-notifications-daemon/main.go
+++ b/cmd/icinga-notifications-daemon/main.go
@@ -41,11 +41,6 @@ func main() {
 		return
 	}
 
-	if configPath == "" {
-		_, _ = fmt.Fprintln(os.Stderr, "missing -config flag")
-		os.Exit(1)
-	}
-
 	err := daemon.LoadConfig(configPath)
 	if err != nil {
 		_, _ = fmt.Fprintln(os.Stderr, "cannot load config:", err)
--- a/internal/daemon/config.go
+++ b/internal/daemon/config.go
@@ -2,12 +2,127 @@ package daemon
 
 import (
 	"errors"
+	"fmt"
 	"github.com/creasty/defaults"
 	"github.com/goccy/go-yaml"
 	icingadbConfig "github.com/icinga/icingadb/pkg/config"
+	"io"
 	"os"
+	"reflect"
+	"regexp"
+	"strings"
 )
 
+// populateFromYamlEnvironmentPathStep is the recursive worker function for PopulateFromYamlEnvironment.
+//
+// It performs a linear search along the path with pathNo as the current element except when a wild `,inline` appears,
+// resulting in branching off to allow peeking into the inlined struct.
+func populateFromYamlEnvironmentPathStep(keyPrefix string, cur reflect.Value, path []string, pathNo int, value string) error {
+	notFoundErr := errors.New("cannot resolve path")
+
+	subKey := keyPrefix + "_" + strings.Join(path[:pathNo+1], "_")
+
+	t := cur.Type()
+	for fieldNo := 0; fieldNo < t.NumField(); fieldNo++ {
+		fieldName := t.Field(fieldNo).Tag.Get("yaml")
+		if fieldName == "" {
+			return fmt.Errorf("field %q misses yaml struct tag", subKey)
+		}
+		if strings.Contains(fieldName, "_") {
+			return fmt.Errorf("field %q contains an underscore, the environment key separator, in its yaml struct tag", subKey)
+		}
+
+		if regexp.MustCompile(`^.*(,[a-z]+)*,inline(,[a-z]+)*$`).MatchString(fieldName) {
+			// Peek into the `,inline`d struct but ignore potential failure.
+			err := populateFromYamlEnvironmentPathStep(keyPrefix, reflect.Indirect(cur).Field(fieldNo), path, pathNo, value)
+			if err == nil {
+				return nil
+			} else if !errors.Is(err, notFoundErr) {
+				return err
+			}
+		}
+
+		if strings.ToUpper(fieldName) != path[pathNo] {
+			continue
+		}
+
+		if pathNo < len(path)-1 {
+			return populateFromYamlEnvironmentPathStep(keyPrefix, reflect.Indirect(cur).Field(fieldNo), path, pathNo+1, value)
+		}
+
+		field := cur.Field(fieldNo)
+		tmp := reflect.New(field.Type()).Interface()
+		err := yaml.NewDecoder(strings.NewReader(value)).Decode(tmp)
+		if err != nil {
+			return fmt.Errorf("cannot unmarshal into %q: %w", subKey, err)
+		}
+		field.Set(reflect.ValueOf(tmp).Elem())
+		return nil
+	}
+
+	return fmt.Errorf("%w %q", notFoundErr, subKey)
+}
+
+// PopulateFromYamlEnvironment populates a struct with "yaml" struct tags based on environment variables.
+//
+// To write into targetElem, it must be passed as a pointer reference.
+//
+// Environment variables of the form ${KEY_PREFIX}_${KEY_0}_${KEY_i}_${KEY_n}=${VALUE} will be translated to a YAML path
+// from the struct field with the "yaml" struct tag ${KEY_0} across all further nested fields ${KEY_i} up to the last
+// ${KEY_n}. The ${VALUE} will be YAML decoded into the referenced field of the targetElem.
+//
+// Next to addressing fields through keys, elementary `,inline` flags are also being supported. This allows referring an
+// inline struct's field as it would be a field of the parent.
+//
+// Consider the following struct:
+//
+//	type Example struct {
+//		Outer struct {
+//			Inner int `yaml:"inner"`
+//		} `yaml:"outer"`
+//	}
+//
+// The Inner field can get populated through:
+//
+//	PopulateFromYamlEnvironment("EXAMPLE", &example, []string{"EXAMPLE_OUTER_INNER=23"})
+func PopulateFromYamlEnvironment(keyPrefix string, targetElem any, environ []string) error {
+	matcher, err := regexp.Compile(`(?s)\A` + keyPrefix + `_([A-Z0-9_-]+)=(.*)\z`)
+	if err != nil {
+		return err
+	}
+
+	if reflect.ValueOf(targetElem).Type().Kind() != reflect.Ptr {
+		return errors.New("targetElem is required to be a pointer")
+	}
+
+	for _, env := range environ {
+		match := matcher.FindStringSubmatch(env)
+		if match == nil {
+			continue
+		}
+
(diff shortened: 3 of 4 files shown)


oxzi:
This PR's motivation is to address this comment <https://github.com/Icinga/icinga-notifications/pull/144#issuecomment-1877369284> which might be a precondition for the integration tests.

julianbrost:
> The code works by reflecting on the existing nested ConfigFile struct.

If it already uses reflection on the config struct, why did you choose to generate a YAML file which is then later parsed into that struct rather than setting the values using reflection?

Also, it would be great if the code would be reusable so that Icinga DB could be adapted to use this as well in the future. If I'm not missing something, this would only require making the environment variable prefix and destination type function parameters.

oxzi:
> > The code works by reflecting on the existing nested ConfigFile struct.
> 
> If it already uses reflection on the config struct, why did you choose to generate a YAML file which is then later parsed into that struct rather than setting the values using reflection?

Because I wanted to achieve the exact same behavior as by passing a YAML configuration.

This starts with detecting the right native type for each field and trying to convert the environment variable's value - which is at first always a string - to it and ends at custom parsing for everything from different `*Unmarshaler` interfaces to, e.g.,  `time.Duration`[^go-yaml-duration]. Thus, I had to re-write or import lots of already existing code to mimic a library's behavior which is already included.

With this architecture, a small amount of code translates environment variables to the already known configuration format.

> Also, it would be great if the code would be reusable so that Icinga DB could be adapted to use this as well in the future. If I'm not missing something, this would only require making the environment variable prefix and destination type function parameters.

Yes, the current code only requires that `yaml` struct tags are available, which is afaik also the case for Icinga DB. Thus, when making the environment variable prefix configurable, it should be portable.

[^go-yaml-duration]: https://github.com/goccy/go-yaml/blob/master/decode.go#L1099-L1125

julianbrost:
The basic problem here is taking a `string` and putting it into `any` (representing some scalar type), isn't it? Unfortunately, [`TextUnmarshaler`](https://pkg.go.dev/encoding#TextUnmarshaler) isn't implemented universally, but if it was, I'd say that would be the way to go.

Your code already contains some special handling for integers (probably inspired by the Icinga DB container entrypoint):
https://github.com/Icinga/icinga-notifications/blob/83ce59ae181ce831126b8ccd1126e5bc533ecf89/internal/daemon/config.go#L123-L128

Which is basically guessing based on the value, if the destination type would be taken into account, the code could make a better informed decision about which conversion would be needed.

Icinga DB already happens to need and have a [`string` to `any` function](https://github.com/Icinga/icingadb/blob/0e9810c1e7d5b619fec48849698ff14a0526176f/pkg/structify/structify.go#L110-L144) function anyways, which will make it's way to [`icinga-go-library`](https://github.com/Icinga/icinga-go-library/blob/5f0078f60fb15d3d9c315d846d03635fd8f2ba8e/structify/structify.go#L105-L139), could be exported and extended with the missing few types and also be used here.

Note: I couldn't test if the current code would handle `bool` correctly, the only ones available in the config are somewhere that uses `yaml:",inline"` which isn't handled by the code currently and also, the environment variable names are derived from the struct member names, not the yaml tags. Currently, the enable TLS option is (attempted to be) used from `ICINGA_NOTIFICATIONS_DATABASE_TLSOPTIONS_ENABLE=true` whereas in the config file YAML it's just `database.tls`. Also, the generated YAML doesn't work in that situation:

```yaml
database:
  ",inline":
    tls: "true"
```

oxzi:
> The basic problem here is taking a `string` and putting it into `any` (representing some scalar type), isn't it? Unfortunately, [`TextUnmarshaler`](https://pkg.go.dev/encoding#TextUnmarshaler) isn't implemented universally, but if it was, I'd say that would be the way to go.

Exactly.

Unfortunately, even some (more or less) scalar types from the standard library do not implement the `TextUnmarshaler`, e.g., `time.Time` or `time.Duration`. As it seems, those two are not going to implement it - <https://github.com/golang/go/issues/16039>.

At this point, I would like to link again to go-yaml's additional type check:
<https://github.com/goccy/go-yaml/blob/0640a158e99023787bc503fd4476e472d804c3a3/decode.go#L674-L692>

Furthermore, there are all possible compound data types (structs, arrays and slices, maps) and interfaces.

> 
> Your code already contains some special handling for integers (probably inspired by the Icinga DB container entrypoint):
> 
> https://github.com/Icinga/icinga-notifications/blob/83ce59ae181ce831126b8ccd1126e5bc533ecf89/internal/daemon/config.go#L123-L128
> 
> Which is basically guessing based on the value, if the destination type would be taken into account, the code could make a better informed decision about which conversion would be needed.

This is, honestly, just ~~an ugly hack~~ a very intelligent solution as unquoted text might still get parsed as a string - even when the string only consist of number literals -, but a quoted number cannot be parsed as a numeric value. For all values in the added tests, this was the only corner case I ran into (maybe because I ignored the `, inline` you mentioned later on).

> Icinga DB already happens to need and have a [`string` to `any` function](https://github.com/Icinga/icingadb/blob/0e9810c1e7d5b619fec48849698ff14a0526176f/pkg/structify/structify.go#L110-L144) function anyways, which will make it's way to [`icinga-go-library`](https://github.com/Icinga/icinga-go-library/blob/5f0078f60fb15d3d9c315d846d03635fd8f2ba8e/structify/structify.go#L105-L139), could be exported and extended with the missing few types and also be used here.

As I stated earlier, this would require more testing to ensure that this code behaves like go-yaml's code. Furthermore, when there are changes to either the future icinga-go-library or go-yaml code, things might break.

IMO, when using a (partial) YAML configuration as an intermediate representation, the logical coupling is smaller as when try to mimic go-yaml's behavior, especially where they made decisions in the absence of standards.

> Note: I couldn't test if the current code would handle `bool` correctly, the only ones available in the config are somewhere that uses `yaml:",inline"` which isn't handled by the code currently and also, the environment variable names are derived from the struct member names, not the yaml tags. Currently, the enable TLS option is (attempted to be) used from `ICINGA_NOTIFICATIONS_DATABASE_TLSOPTIONS_ENABLE=true` whereas in the config file YAML it's just `database.tls`. Also, the generated YAML doesn't work in that situation:
> 
> ```yaml
> database:
>   ",inline":
>     tls: "true"
> ```

The field names were derived from the Go fields as, by naming convention, Go does not use underscores as separators but camel case. I was expecting to find underscores in the `yaml` struct fields as separators, but, as I just tried to verified that, there are none, but dashes.
My assumption was that an underscore in a `yaml` struct tag would break the environment variable separator. As long as none are going to appear, this should work. (Let's try to not mention the sword of Damocles hanging over this "parser".)

I'll try to generalize the code as mentioned in <https://github.com/Icinga/icinga-notifications/pull/148#issuecomment-1887517075>, make some tests for other types, e.g., booleans, derive the names from the `yaml` struct tag with an additional safety check, and try to address the ` ,inline` possibility.

oxzi:
After multiple iterations, both the code as well as the logic has now changed quite a bit.

I reflected on my prior statement that creating a temporary YAML configuration would be a good idea and now used the available runtime type information to decode each value into a given reference to the struct directly. Furthermore, I added rudimentary support for `,inline`d fields and created tests both on a green field as well as for the `Database.TlsOptions`.

Please feel free to take another look, @julianbrost.

lippserd:
Have you considered existing libraries such as https://github.com/caarlos0/env?

oxzi:
> Have you considered existing libraries such as https://github.com/caarlos0/env?

Yes. IIRC, I have even tried this library or at least some other Go library for this purpose.

The problem is, however, it requires a fixed `env` struct tag on all fields. Since this also affects fields of types within Icinga DB, that I don't have code write access to, this turns out to be difficult. On first glance, there is no built-in mechanism to handle missing struct tags.

Furthermore, I aimed to get both the same names as well as the same decoding behavior as with the already used YAML library. As outlined in earlier comments, there is already some highly opinionated behavior and having to mimic another library's decisions - also with future or forward compatibility - just sounds tedious.

julianbrost:
> The problem is, however, it requires a fixed `env` struct tag on all fields. Since this also affects fields of types within Icinga DB, that I don't have code write access to, this turns out to be difficult.

That shouldn't be a general issue. This code would probably be destined for icinga-go-library where those database config struct will also end up. (It would probably be a problem if the library wouldn't allow a custom variable prefix per project.)

> On first glance, there is no built-in mechanism to handle missing struct tags.

Is that a general issue or is it just saying that because of that, it's not possible to build a workaround that wouldn't require adding tags to the structs?

> Furthermore, I aimed to get both the same names as well as the same decoding behavior as with the already used YAML library. As outlined in earlier comments, there is already some highly opinionated behavior and having to mimic another library's decisions - also with future or forward compatibility - just sounds tedious.

In general, please elaborate more on which libraries you looked into and what limitations made you rule them out in particular. We might find that some can easily be worked around, otherwise we have a good picture of why implementing it on our own is the better choice.

oxzi:
> In general, please elaborate more on which libraries you looked into and what limitations made you rule them out in particular. We might find that some can easily be worked around, otherwise we have a good picture of why implementing it on our own is the better choice.

I have tried two different libraries.

My first attempt was using <https://github.com/sethvargo/go-envconfig>. Unfortunately, my testing code is gone as I have not committed it.

However, with @lippserd's suggestion I achieved more, as outlined below. (But GitHub just ate my half-written comparison of what worked and what did not.)

Then, I tried <https://github.com/caarlos0/env>.

- It is possible to change the checked struct tag from `env` to, e.g., `yaml` through the [`Options`](https://pkg.go.dev/github.com/caarlos0/env/v10#Options).
  Even when names like `ICINGA_NOTIFICATIONS_icingaweb2-url` might not be beautiful to the eye, this generally works.
  However, as there are some additional comma-separated attributes in the go-yaml struct tags which are not known to env, the parser errors with something like "_env: tag option "inline" not supported_". With ignoring the errors, this will still work, but ignoring parser errors might not be the best move.
- As the previous error shows, it is not possible to directly mimic go-yaml's struct `inline`ing. This would result in manual code to achieve feature or naming parity.
- Instead of using struct tags, it is possible to let the library derive names from their field names via `UseFieldNameByDefault`.
  This mostly works, but in some cases the [`toEnvName` function](https://github.com/caarlos0/env/blob/v10.0.0/env.go#L301) generates unguessable names, e.g., `Icingaweb2URL` will be `ICINGAWEB_2U_R_L`.
- When using nested structs, the additional `envPrefix` struct tag becomes necessary to resolve sub-struct fields. Take a look at this [example](https://github.com/caarlos0/env#prefixes).
- Contrary to prior assumptions, unmarshalling a `time.Duration` worked as expected with go-yaml.
  For example, `ICINGA_NOTIFICATIONS_interval=9001h` correctly populated the `Logging.Interval` field.

In a nutshell, with some tweaks it would be possible to use the <https://github.com/caarlos0/env> library.

However, in the meantime since creating this PR, the premise has changed.
First, I wanted to alter only code in `icinga-notifications` and thus used some (more or less hacky) workarounds to achieve this. Now, with planning to move the referenced code from `icingadb` to the `icinga-go-library` anyway, we can then also populate it with correct `env` struct tags.

Thus, would it be possible to (re-)define the concrete scope of this PR, which was initially created to address a [comment](https://github.com/Icinga/icinga-notifications/pull/144#issuecomment-1877369284) of @julianbrost, allowing to configure Icinga Notifications through environment variables for integration tests. We have moved a considerable distance away from the initial motivation, inflating the dependencies before getting anywhere near the tests.

Now, to something completely different: password and other credentials in environment variables.

I am not quite sure if this is desirable from a security perspective. First, at least the same, or a more privileged, user can access the environment through `/proc` (on Linux) or `ps`. It seems like there are (or at least were) some systems with a [`setuid`ed `ps`](https://stackoverflow.com/questions/12461484/is-it-secure-to-store-passwords-as-environment-variables-rather-than-as-plain-t#comment62701956_12461680), allowing any user to inspect each process' environment. Furthermore, as listed in this [blog post](https://blog.diogomonica.com//2017/03/27/why-you-shouldnt-use-env-variables-for-secret-data/), the environment will be passed to child processes - which we are currently creating - and is often included in crash logs.

One can argue that currently the subprocesses are having the same privileges as the daemon anyway and we have control over logs. But, however, this seems like something which might get forgotten along the path and that traditional file permissions are an easier concept.

Al2Klimov:
> Now, to something completely different: password and other credentials in environment variables.
> 
> I am not quite sure if this is desirable from a security perspective. (...) the environment will be passed to child processes - which we are currently creating - and is often included in crash logs.

Not if we unset such first. https://github.com/caarlos0/env?tab=readme-ov-file#unset-environment-variable-after-reading-it

PR #149, opened 2024-01-12
Webhook Channel

oxzi:
This PR contains two changes:

* plugin: Document ConfigOption's Help field
* Introduce Webhook Channel
  In its initial form, the webhook channel can be used to send notifications against a HTTP/HTTPS web server with a dynamic configurable URL and request body message.

This addresses #144 where @julianbrost [suggested to implement a web hook channel](https://github.com/Icinga/icinga-notifications/pull/144#issuecomment-1877245251).

I have tested this channel locally against <https://github.com/mendhak/docker-http-https-echo> for debugging. Next, I am going to rework #144 to use this channel instead of the file-like channel.

### DIFF
--- a/cmd/channel/webhook/main.go
+++ b/cmd/channel/webhook/main.go
@@ -1,0 +1,170 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"github.com/icinga/icinga-notifications/internal"
+	"github.com/icinga/icinga-notifications/pkg/plugin"
+	"io"
+	"net/http"
+	"slices"
+	"strconv"
+	"strings"
+	"text/template"
+)
+
+type Webhook struct {
+	Method              string `json:"method"`
+	URLTemplate         string `json:"url_template"`
+	RequestBodyTemplate string `json:"request_body_template"`
+	ResponseStatusCodes string `json:"response_status_codes"`
+
+	tmplUrl         *template.Template
+	tmplRequestBody *template.Template
+
+	respStatusCodes []int
+}
+
+func (ch *Webhook) GetInfo() *plugin.Info {
+	elements := []*plugin.ConfigOption{
+		{
+			Name: "method",
+			Type: "string",
+			Label: map[string]string{
+				"en_US": "HTTP Method",
+				"de_DE": "HTTP-Methode",
+			},
+			Help: map[string]string{
+				"en_US": "HTTP request method used for the web request.",
+				"de_DE": "HTTP-Methode für die Anfrage.",
+			},
+			Default:  "POST",
+			Required: true,
+		},
+		{
+			Name: "url_template",
+			Type: "string",
+			Label: map[string]string{
+				"en_US": "URL Template",
+				"de_DE": "URL-Template",
+			},
+			Help: map[string]string{
+				"en_US": "URL, optionally as a Go template over the current plugin.NotificationRequest.",
+				"de_DE": "URL, optional als Go-Template über das zu verarbeitende plugin.NotificationRequest.",
+			},
+			Required: true,
+		},
+		{
+			Name: "request_body_template",
+			Type: "string",
+			Label: map[string]string{
+				"en_US": "Request Body Template",
+				"de_DE": "Anfragedaten-Template",
+			},
+			Help: map[string]string{
+				"en_US": "Go template applied to the current plugin.NotificationRequest to create an request body.",
+				"de_DE": "Go-Template über das zu verarbeitende plugin.NotificationRequest zum Erzeugen der mitgesendeten Anfragedaten.",
+			},
+			Default: `{{json .}}`,
+		},
+		{
+			Name: "response_status_codes",
+			Type: "string",
+			Label: map[string]string{
+				"en_US": "Response Status Codes",
+				"de_DE": "Antwort-Status-Codes",
+			},
+			Help: map[string]string{
+				"en_US": "Comma separated list of expected HTTP response status code, e.g., 200,201,202,208,418",
+				"de_DE": "Kommaseparierte Liste erwarteter Status-Code der HTTP-Antwort, z.B.: 200,201,202,208,418",
+			},
+			Default:  "200",
+			Required: true,
+		},
+	}
+
+	configAttrs, err := json.Marshal(elements)
+	if err != nil {
+		panic(err)
+	}
+
+	return &plugin.Info{
+		Name:             "Webhook",
+		Version:          internal.Version.Version,
+		Author:           "Icinga GmbH",
+		ConfigAttributes: configAttrs,
+	}
+}
+
+func (ch *Webhook) SetConfig(jsonStr json.RawMessage) error {
+	err := json.Unmarshal(jsonStr, ch)
+	if err != nil {
+		return err
+	}
+
+	tmplFuncs := template.FuncMap{
+		"json": func(a any) (string, error) {
+			data, err := json.Marshal(a)
+			if err != nil {
+				return "", err
+			}
+			return string(data), nil
+
+		},
+	}
+
+	ch.tmplUrl, err = template.New("url").Funcs(tmplFuncs).Parse(ch.URLTemplate)
+	if err != nil {
+		return fmt.Errorf("cannot parse URL template: %w", err)
+	}
+
+	ch.tmplRequestBody, err = template.New("request_body").Funcs(tmplFuncs).Parse(ch.RequestBodyTemplate)
+	if err != nil {
+		return fmt.Errorf("cannot parse Request Body template: %w", err)
+	}
+
+	respStatusCodes := strings.Split(ch.ResponseStatusCodes, ",")
+	ch.respStatusCodes = make([]int, len(respStatusCodes))
+	for i, respStatusCodeStr := range respStatusCodes {
+		respStatusCode, err := strconv.Atoi(respStatusCodeStr)
+		if err != nil {
+			return fmt.Errorf("cannot convert status code %q to int: %w", respStatusCodeStr, err)
+		}
+		ch.respStatusCodes[i] = respStatusCode
+	}
+
+	return nil
+}
+
+func (ch *Webhook) SendNotification(req *plugin.NotificationRequest) error {
+	var urlBuff, reqBodyBuff bytes.Buffer
+	if err := ch.tmplUrl.Execute(&urlBuff, req); err != nil {
+		return fmt.Errorf("cannot execute URL template: %w", err)
+	}
+	if err := ch.tmplRequestBody.Execute(&reqBodyBuff, req); err != nil {
+		return fmt.Errorf("cannot execute Request Body template: %w", err)
+	}
+
+	httpReq, err := http.NewRequest(ch.Method, urlBuff.String(), &reqBodyBuff)
+	if err != nil {
+		return err
+	}
+	httpResp, err := http.DefaultClient.Do(httpReq)
+	if err != nil {
+		return err
+	}
+	_, _ = io.Copy(io.Discard, httpResp.Body)
+	_ = httpResp.Body.Close()
+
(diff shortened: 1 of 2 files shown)


cmd/channel/webhook/main.go
```diff
@@ -0,0 +1,138 @@
...
+			Label: map[string]string{
+				"en_US": "Request Body Template",
+				"de_DE": "Anfragedaten-Template",
+			},
+			Help: map[string]string{
+				"en_US": "Go template applied to the current plugin.NotificationRequest to create an request body.",
+				"de_DE": "Go-Template über das zu verarbeitende plugin.NotificationRequest zum Erzeugen der mitgesendeten Anfragedaten.",
+			},
+			// Default as a minimal working example to send a JSON object.
+			Default: `{"id": {{.Incident.Id}}, "type": "{{.Event.Type}}", "name": "{{.Object.Name}}", "severity": "{{.Incident.Severity}}", "url": "{{.Incident.Url}}"}`,
```
julianbrost, line 76: As far as I know, there's no escaping magic in `text/template` like there is in `html/template`.

So if this is the intended use, there at least has to be a function to properly JSON-encode values so that you can do something like `"name": {{json .Object.Name}}` so that you don't run into quoting issues.

I'm unsure if the built-in [`js` function](https://pkg.go.dev/text/template#hdr-Functions) would be suitable for this, at the very least it would named confusingly if it was.

oxzi: Next to the `js` or maybe `urlquery` function, we could add extra escaping functions (like PHP's infamous `htmlentities` or `mysql_escape_string` functions), but in the end this approach would still offer an open template format string to the user.

From a security perspective, this is questionable, but otherwise I have no idea how to make a webhook channel configurable enough. I mean, there is even a `call` function… But still, the current PR state is quite strict and has, e.g., no additional header or authentication support.

Or would just suggest to add escaping function calls to the default value?

julianbrost: Sorry, I totally missed that comment.

From a security perspective, I currently see three points:
1. Designing things in a way that that the correct/secure way to use them is the most intuitive one, like SQL prepared statements vs. string concatenation and manual escaping (as you already mentioned). If we want to do this here, we would need something more advances then a basic template engine, either one that know understands what it's generating (like Go's `html/template` is for HTML) or something completely different (maybe like Jsonnet, which is probably a bit more than just a template language). The addition of a `json` function was merely the probably least invasive change needed to the (at that time current state of the) PR to allow using it in a safe way for JSON.
2. Doing bad things to the process to the process evaluating the template: I guess that's what you mean when you mention the `call` function, where that might allow to call undesired functions, in the worst case resulting in remote code execution in the process evaluating the template. With Go's `text/template`, that should only be the case if you expose such functions, so we should probably be safe as it just receives a struct that contains values deserialized from JSON. But of course something we should make sure.
3. Abusing the intended functionality to do unintended things: in this example, an attacker could who gained write access to the database could use this to send requests to systems otherwise not accessible to them. Here, it's the question of finding the balance between convenience and security.

And this is the story of how the quick thought of "a generic webhook channel implementation would probably be nice" turns into not so simple questions.

We can take something like this as a starting point. Properly addressing all of this feels like to much to ask when the actual goal that brought us here is to implement tests.

oxzi: > Sorry, I totally missed that comment.

No problem.

> From a security perspective, I currently see three points:
> 
>1. Designing things in a way that that the correct/secure way to use them is the most intuitive one, like SQL prepared statements vs. string concatenation and manual escaping (as you already mentioned). If we want to do this here, we would need something more advances then a basic template engine, either one that know understands what it's generating (like Go's `html/template` is for HTML) or something completely different (maybe like Jsonnet, which is probably a bit more than just a template language). The addition of a `json` function was merely the probably least invasive change needed to the (at that time current state of the) PR to allow using it in a safe way for JSON.

For a generic HTTP-based channel, it depends on the context what should be escaped. Are we dealing with creating some HTML and need to be aware of XSS or are we creating some query in SQL or even a command to be executed?

The problem lies in the fact that both a naive user as well as an attacker can adjust the template at runtime. No prepared statement or anything similar would help in this case.

Thus, regarding this bullet point I would see the responsibility on the backend's site. Otherwise, an adversary could just attack the vulnerable webhook backend directly (if it is directly reachable) and does not have to infiltrate some malicious payload.

> 2. Doing bad things to the process to the process evaluating the template: I guess that's what you mean when you mention the `call` function, where that might allow to call undesired functions, in the worst case resulting in remote code execution in the process evaluating the template. With Go's `text/template`, that should only be the case if you expose such functions, so we should probably be safe as it just receives a struct that contains values deserialized from JSON. But of course something we should make sure.

I was afraid of the possibilities of a format string attack on steroids through a rich template language. Due to the fact that each channel is being executed with the same OS permissions as the icinga-notifications daemon itself, this would be bad.

> 3. Abusing the intended functionality to do unintended things: in this example, an attacker could who gained write access to the database could use this to send requests to systems otherwise not accessible to them. Here, it's the question of finding the balance between convenience and security.

Honestly, if an adversary can write to the database, we have lost with the current architecture. It would allow them to disable notifications, get (hopefully limited) credentials for the Icinga 2 API and more. But maybe this is out of this PR's scope, sorry.

> And this is the story of how the quick thought of "a generic webhook channel implementation would probably be nice" turns into not so simple questions.
> 
> We can take something like this as a starting point. Properly addressing all of this feels like to much to ask when the actual goal that brought us here is to implement tests.

cmd/channel/webhook/main.go
```diff
@@ -0,0 +1,148 @@
...
+	err := json.Unmarshal(jsonStr, ch)
+	if err != nil {
+		return err
+	}
+
+	tmplFuncs := template.FuncMap{
+		"json": func(a any) string {
+			var buff bytes.Buffer
+			if err := json.NewEncoder(&buff).Encode(a); err != nil {
+				panic(err)
```
julianbrost, line 103: [`FuncMap`](https://pkg.go.dev/text/template#FuncMap) explicitly allows to return errors:

> or two return values of which the second has type error. In that case, if the second (error) return value evaluates to non-nil during execution, execution terminates and Execute returns that error.

So there seems to be no need to panic here.

cmd/channel/webhook/main.go
```diff
@@ -0,0 +1,148 @@
...
+			Label: map[string]string{
+				"en_US": "Request Body Template",
+				"de_DE": "Anfragedaten-Template",
+			},
+			Help: map[string]string{
+				"en_US": "Go template applied to the current plugin.NotificationRequest to create an request body.",
+				"de_DE": "Go-Template über das zu verarbeitende plugin.NotificationRequest zum Erzeugen der mitgesendeten Anfragedaten.",
+			},
+			// Default as a minimal working example to send a JSON object.
+			Default: `{"id": {{.Incident.Id}}, "type": "{{.Event.Type}}", "name": "{{.Object.Name}}", "severity": "{{.Incident.Severity}}", "url": "{{.Incident.Url}}"}`,
```
julianbrost, line 76: That default value should use the `json` encode function that's now exposed to the templates to insert values into the JSON.

cmd/channel/webhook/main.go
```diff
@@ -0,0 +1,148 @@
...
+func (ch *Webhook) SetConfig(jsonStr json.RawMessage) error {
+	err := json.Unmarshal(jsonStr, ch)
+	if err != nil {
+		return err
+	}
+
+	tmplFuncs := template.FuncMap{
+		"json": func(a any) string {
+			var buff bytes.Buffer
+			if err := json.NewEncoder(&buff).Encode(a); err != nil {
```
julianbrost, line 102: Is there a specific reason to explicitly create an encode here instead of just using `json.Marshal()`?

julianbrost (changes requested):

cmd/channel/webhook/main.go
oxzi: I cannot remember why, but there doesn't seem to be any reason against `json.Marshal`.

cmd/channel/webhook/main.go
```diff
@@ -0,0 +1,149 @@
...
+				"GET":     "GET",
+				"HEAD":    "HEAD",
+				"POST":    "POST",
+				"PUT":     "PUT",
+				"DELETE":  "DELETE",
+				"CONNECT": "CONNECT",
+				"OPTIONS": "OPTIONS",
+				"TRACE":   "TRACE",
+				"PATCH":   "PATCH",
+			},
```
julianbrost, line 49: That's an interesting list for sure. I guess the idea behind providing all of them was so that there are no arbitrary restrictions? But on the other hand, this will now provide quite some strange suggestions.

Another option could be to just use an arbitrary text input with default `POST`, that will probably be the value of choice in almost all cases anyways.

If there was a way to express "or a custom value" with `Type: "option"`, I think that would be the obvious choice, but at the moment, that's not possible. So I'm not entirely sure what the best option is at the moment.

cmd/channel/webhook/main.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	httpResp, err := http.DefaultClient.Do(httpReq)
+	if err != nil {
+		return err
+	}
+
+	_, _ = io.Copy(io.Discard, httpResp.Body)
+	_ = httpResp.Body.Close()
+
+	return nil
```
julianbrost, line 144: This should probably perform some check on the response, at least that the status code was 2xx or something like that.

oxzi: When your endpoint honors HTTP, then yes. But in our sad reality, this mustn't be the case. Unless one wants to exclude someone's potential use case, then this must be configurable.

However, this opens a wide field of optional configurations: additional or changed HTTP headers, HTTP versions, TLS configuration, response parsing and whatnot. At day's end, one might end up with something like the good old [check_http plugin](https://www.monitoring-plugins.org/doc/man/check_http.html) - can do variants of everything, is hard to test and maybe one just wanna use a wrapper around curl anyway.

I don't want to create a straw man against performing a small change, but want to underline both that I considered the idea and that making a channel (or check plugin) as flexible as a protocol like HTTP is maybe too much. And then there are even further cases, e.g., `PUT`ing a new Notification into a very REST-ful backend and `DELETE`ing when it's resolved (_inb4: just create a custom channel_).

cmd/channel/webhook/main.go
oxzi: I guess this boils down to the question if this channel should cover HTTP in all its glory and with all its features or to cover _just enough_, which will eventually also lead to feature creep in the long run.

When I first started this, I did enough for the integration tests with some additional knobs. But, as I also wrote in my [other comment](https://github.com/Icinga/icinga-notifications/pull/149#discussion_r1568933901), I don't think one can develop a maintainable channel for everything which might be covered under the name _webhook_.

On the one hand, I would argue to limit this channel to the integration test's necessaries and later move it into a testing sub-directory. However, on the other hand, real users are going to request something HTTP-speaking as most external notification/messaging services are having an HTTP API.

cmd/channel/webhook/main.go
julianbrost: > When your endpoint honors HTTP, then yes. But in our sad reality, this mustn't be the case. Unless one wants to exclude someone's potential use case, then this must be configurable.

What else should they speak but HTTP? I mean this isn't about allowing for every possible scenario like always sending 200 OK but then including an error message in the response body. With the current state of the PR, there is no reasonable way for the remote to signal an error, like they would have to close the connection without sending a response to do so.

oxzi: Speaking enough HTTP to make a client happy and honoring the HTTP specifications are two different things, I'd say. But yes, I am going to add an expected status code, defaulting to 200.

I mainly wanted to express my fear of this simple channel degenerating into some untestable state monster that honors every use case while being horrible to use.

julianbrost (approved):

PR #169, opened 2024-04-10
Rename `icinga-notifications-daemon` -> `icinga-notifications`

yhabteab:
It's pointless to append `-daemon` to the product name, and we don't do it that way in Icinga DB btw.


julianbrost:
I wouldn't say it's completely pointless. With Icinga DB, the thing is, if you say just that, it's not clear if you referring to its Go daemon in particular or the whole thing, i.e. including Icinga DB Web and the Icinga 2 feature. So keeping the daemon suffix would give it a distinct name. But yes, that would be inconsistent with Icinga DB, so I won't insist on this, but I'd like to hear a few more opinions before deciding.

julianbrost (changes requested):
As there were no name suggestions that were so convincing that it would justify inconsistent naming with other Icinga projects, we'll do it as suggested by this PR. It just got conflicts in the meantime and thus needs a rebase.

PR #171, opened 2024-04-12
Commit transaction for acknowledgements

julianbrost:
The early return from `ProcessEvent()` for acknowledgements meant that the commit of the database transaction later in the function would never happen. This was introduced in 4044b188a9f9423be532d11fc0c70b313fc0ca39 which moved the transaction handling into `ProcessEvent()`


yhabteab (approved):
🤦‍♂️

PR #175, opened 2024-04-15
Distinct "internal" Event Types

### LINKED ISSUE #162: Use type to reflect the actual type of internal event
At the moment, internal events use the type name `internal` which is rather useless for Web.

The type should use specific names to indicate what the event is about.

It's currently used in [two cases](https://github.com/Icinga/icinga-notifications/blob/fdb0808af446928c3127b8441078718792e2ac18/internal/event/event.go#L39):

* Incident age thresholds
* Incident reevaluation at startup

There may be more added in the future. We need to make sure that there are no such surprises for Web anymore, @julianbrost suggested to use an enum for `event.type` to ensure that.

oxzi:
The "internal" event type was used both for scheduled reevaluations and for initial reevaluations during startup. This type is currently being used at two different points for two distinct topics: scheduled reevaluations and initial reevaluations during startup.

This event type was now split into "reevaluation-scheduled" and "reevaluation-startup" to distinct those types.

To ensure database-level stability, the column type was changed from text to a custom enum.

Closes #162.

### DIFF
--- a/internal/event/event.go
+++ b/internal/event/event.go
@@ -37,17 +37,19 @@ type Event struct {
 	ID int64 `json:"-"`
 }
 
+// Please keep the following types in alphabetically order and, even more important, make sure that the database type
+// event_type reflects the same values.
 const (
-	TypeState                  = "state"
-	TypeAcknowledgementSet     = "acknowledgement-set"
 	TypeAcknowledgementCleared = "acknowledgement-cleared"
-	TypeInternal               = "internal"
+	TypeAcknowledgementSet     = "acknowledgement-set"
+	TypeCustom                 = "custom"
+	TypeDowntimeEnd            = "downtime-end"
 	TypeDowntimeRemoved        = "downtime-removed"
 	TypeDowntimeStart          = "downtime-start"
-	TypeDowntimeEnd            = "downtime-end"
-	TypeCustom                 = "custom"
-	TypeFlappingStart          = "flapping-start"
 	TypeFlappingEnd            = "flapping-end"
+	TypeFlappingStart          = "flapping-start"
+	TypeIncidentAge            = "incident-age"
+	TypeState                  = "state"
 )
 
 // Validate validates the current event state.
@@ -68,16 +70,17 @@ func (e *Event) Validate() error {
 	switch e.Type {
 	case "":
 		return fmt.Errorf("invalid event: 'type' must not be empty")
-	case TypeState,
-		TypeAcknowledgementSet,
+	case
 		TypeAcknowledgementCleared,
-		TypeInternal,
+		TypeAcknowledgementSet,
+		TypeCustom,
+		TypeDowntimeEnd,
 		TypeDowntimeRemoved,
 		TypeDowntimeStart,
-		TypeDowntimeEnd,
-		TypeCustom,
+		TypeFlappingEnd,
 		TypeFlappingStart,
-		TypeFlappingEnd:
+		TypeIncidentAge,
+		TypeState:
 		return nil
 	default:
 		return fmt.Errorf("invalid event: unsupported event type %q", e.Type)
--- a/internal/incident/incident.go
+++ b/internal/incident/incident.go
@@ -455,8 +455,8 @@ func (i *Incident) evaluateEscalations(eventTime time.Time) ([]*rule.Escalation,
 			i.logger.Info("Reevaluating escalations")
 
 			i.RetriggerEscalations(&event.Event{
-				Type:    event.TypeInternal,
 				Time:    nextEvalAt,
+				Type:    event.TypeIncidentAge,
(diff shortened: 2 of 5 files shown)


internal/event/event.go
```diff
@@ -34,9 +34,10 @@ type Event struct {
 }
 
 const (
-	TypeState           = "state"
-	TypeAcknowledgement = "acknowledgement"
-	TypeInternal        = "internal"
+	TypeState                 = "state"
+	TypeAcknowledgement       = "acknowledgement"
+	TypeReevaluationScheduled = "reevaluation-scheduled"
```
julianbrost, line 39: For this one, the relevant thing that happened, is that the incident reached a certain age, so I'd say that should be reflected in the type.

internal/event/event.go
```diff
@@ -34,9 +34,10 @@ type Event struct {
...
+	TypeReevaluationStartup   = "reevaluation-startup"
```
julianbrost, line 40: Now that I think of this, how much of a different event is this? I mean currently, any follow-up action is still due to the incident reaching a certain age, this simply additionally says that the processing might have been delayed due to the daemon not running at the time it happened.

@nilmerg You opened the corresponding issue, so that probably means that you'd want to display this differently. Is that correct and how would you display it?

(Something I've thought about in the past and that could serve as an alternative here: store two timestamps with each event: when it happened and when it was processed, that could then be used to show a generic "processing was delayed" indicator if there's a significant difference.)

julianbrost (commented):
General question: do we want to distinguish internal/external events in an obvious way, like keeping an `internal-` prefix? My intuitive choice would have been to keep some common prefix, but no strong opinion there. @nilmerg Any thoughts on this?

internal/event/event.go
oxzi: I have renamed it to reflect it being scheduled due to the age.

internal/event/event.go
nilmerg: I'd have no objections about both using the same type. We'd display an icon and short textual phrase, in addition to the event's message. (e.g. :hourglass: Incident reached age)

nilmerg:
> General question

No, no idea. I wouldn't parse it as such either, so if there's a prefix, it wouldn't have any meaning to Web.

internal/event/event.go
oxzi: I have pushed a new version, effectively renaming `internal` to `reevaluation`. If future _internal_ events emerge, we could then introduce further event types.

nilmerg:
The source icon will be removed from the history alltogether (https://github.com/Icinga/icinga-notifications-web/issues/161) and the internal events won't have one either once https://github.com/Icinga/icinga-notifications-web/issues/157#issuecomment-1976805707 is done.

internal/event/event.go
nilmerg: Is this now final? Can Web rely on these identifiers now?

julianbrost: No, at least I'm not yet convinced enough that `reevaluation` isn't just too generic as well. Intuitively, I'd have named it more like `incident_age`, but I'm als not sure if this isn't too specific and how well we can isolate the actual reason (once there are more) that happened while the daemon was stopped and thus caused some action to happen right after daemon startup.

I'd give it a try to stay as specific as we can, worst case, we could still add an generic "we reevaluated the rule, now it matched, we don't really know the event that caused this" event type in the future, but at the same time, this sounds like something that should be avoided.

julianbrost: No other comments/opinions/feedback on this? :sweat_smile: 

> We'd display an icon and short textual phrase, in addition to the event's message. (e.g. ⌛ Incident reached age)

If that's the stuff you want to display for that event type, it sounds more like you'd also expect something more like `incident_age`, at least "⌛ Incident reached age" wouldn't really fit for `reevaluation`.

julianbrost: I discussed this with @nilmerg yesterday and it looks like there was some confusion due to the following messages being very different:
https://github.com/Icinga/icinga-notifications/blob/62e47fba02c7b19af7230b742bafc387ba855f9f/internal/incident/incidents.go#L129-L133
https://github.com/Icinga/icinga-notifications/blob/62e47fba02c7b19af7230b742bafc387ba855f9f/internal/incident/incident.go#L457-L461

However, in the end both are supposed to say something very similar: some trigger point is now or in the past. Thus, the suggestion would be to use `incident_age` and use more similar messages while still clearly pointing out that the event was delayed, something like "Incident reached age 3h14m (daemon was restarted)" (i.e. 3h14m being the current age, not the trigger time, so that event might trigger escalations for `incident_age>=2h` and `incident_age>=3h`)

oxzi: Please take another look.

julianbrost (approved):

PR #176, opened 2024-04-15
Drop incident_history.caused_by_incident_history_id

### LINKED ISSUE #163: Drop `incident_history.caused_by_incident_history_id`
Since the introduction of internal events, there's no need for incident events to link to each other anymore.

## What needs to be done?

Drop `incident_history.caused_by_incident_history_id` and remove code to populate it. `incident_history.event_id` should in turn be not nullable.

## How is Web affected?

The column isn't actively used, but it also needs to be dropped from the corresponding model.

In the (near) future Web should still try to visualize corresponding history entries. As they share the same `event.id` it can use this to detect how to group events. But this is not planned yet. Though, it would be beneficial to already ensure that `incident_history.time` is used to indicate order. (Even if it's just a millisecond difference)

oxzi:
Next to the schema change, the field and its usage was removed from the Go code. This affects continuous changes to signatures.

refs #163.

### DIFF
--- a/internal/incident/incident.go
+++ b/internal/incident/incident.go
@@ -153,9 +153,8 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event, created bo
 		return nil
 	}
 
-	var causedBy types.Int
 	if !created {
-		causedBy, err = i.processSeverityChangedEvent(ctx, tx, ev)
+		err := i.processSeverityChangedEvent(ctx, tx, ev)
 		if err != nil {
 			return err
 		}
@@ -163,7 +162,7 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event, created bo
 
 	// Check if any (additional) rules match this object. Filters of rules that already have a state don't have
 	// to be checked again, these rules already matched and stay effective for the ongoing incident.
-	causedBy, err = i.evaluateRules(ctx, tx, ev.ID, causedBy)
+	err = i.evaluateRules(ctx, tx, ev.ID)
 	if err != nil {
 		return err
 	}
@@ -174,11 +173,11 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event, created bo
 		return err
 	}
 
-	if err := i.triggerEscalations(ctx, tx, ev, causedBy, escalations); err != nil {
+	if err := i.triggerEscalations(ctx, tx, ev, escalations); err != nil {
 		return err
 	}
 
-	notifications, err := i.addPendingNotifications(ctx, tx, ev, i.getRecipientsChannel(ev.Time), causedBy)
+	notifications, err := i.addPendingNotifications(ctx, tx, ev, i.getRecipientsChannel(ev.Time))
 	if err != nil {
 		return err
 	}
@@ -233,7 +232,7 @@ func (i *Incident) RetriggerEscalations(ev *event.Event) {
 			return fmt.Errorf("can't insert incident event to the database: %w", err)
 		}
 
-		if err = i.triggerEscalations(ctx, tx, ev, types.Int{}, escalations); err != nil {
+		if err = i.triggerEscalations(ctx, tx, ev, escalations); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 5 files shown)


oxzi:
_NOTE: The schema upgrade is 024.sql as I expect #175 to be merged first._

internal/incident/incident.go
```diff
@@ -278,15 +276,13 @@ func (i *Incident) processSeverityChangedEvent(ctx context.Context, tx *sqlx.Tx,
 		Message:     utils.ToDBString(ev.Message),
 	}
 
-	historyId, err := i.AddHistory(ctx, tx, history, true)
+	_, err := i.AddHistory(ctx, tx, history, true)
```
yhabteab, line 279: Since we no longer need to insert and retrieve the history ID here, please change the last arg to `false`.

internal/incident/incident.go
```diff
@@ -381,30 +377,25 @@ func (i *Incident) evaluateRules(ctx context.Context, tx *sqlx.Tx, eventID int64
...
-				RuleID:                    utils.ToDBInt(r.ID),
-				Type:                      RuleMatched,
-				CausedByIncidentHistoryID: causedBy,
+				Time:    types.UnixMilli(time.Now()),
+				EventID: utils.ToDBInt(eventID),
+				RuleID:  utils.ToDBInt(r.ID),
+				Type:    RuleMatched,
 			}
-			insertedID, err := i.AddHistory(ctx, tx, history, true)
+			_, err = i.AddHistory(ctx, tx, history, true)
```
yhabteab, line 389: same here!

schema/pgsql/upgrades/024.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE incident_history
+    ALTER COLUMN event_id SET NOT NULL,
```
yhabteab, line 2: Didn't verify this myself if it's actually the case, but wouldn't that cause problems with existing setups and have `event_id` set to NULL in the history?

oxzi: Well.. this might be the case. But which existing setups?

yhabteab: > Well.. this might be the case. But which existing setups?

Existing setups of our web colleges 😄! I don't think they will be willingly wiping out their history entries, but if @nilmerg says otherwise, then I don't mind.

oxzi: But what would be a sane fallback as it `REFERENCES event(id)`. I could create a placeholder `event` with its placeholder `object` on the first available `source`, but this would "corrupt" the database even more, IMO.

Thus, I would prefer the schema upgrade to fail and let the users (our web colleges) handle with it.

nilmerg: This isn't just an issue during migration of existing setups. I missed that Web creates history entries on its own, when changing recipient roles. In that case, it would also need to emit its own internal events? Eh!

oxzi: @nilmerg: I added the `NOT NULL` clause on your request in #163. Thus, should I remove it again?

nilmerg: I know. :wink: I wanted to say that this needs to be solved first, before finishing this. The `event_id` being non nullable is and should be fine. 

To answer Yonas' argument: Web could handle this on its own, though a `DELETE FROM incident_history where event_id IS NULL` in the upgrade file wouldn't hurt either.

oxzi: K. I have added the `DELETE` into the schema upgrade file.

yhabteab (approved):

yhabteab (dismissed):

schema/pgsql/upgrades/024.sql
julianbrost: So if I understand correctly, merging this as-is would break the (un)manage/subscribe buttons in Web due to this `not null` constraint, wouldn't it?

@nilmerg Should this wait for something in Web therefor?

I think there will be another PR anyways which will add a new type for Web to use in the `event_type` enum introduced in #175, so adding the constraint could also be postponed to that PR until Web is ready for it?

nilmerg: +1 for postponing the constraint change

julianbrost (changes requested):
Please remove the schema changes for `event_id` from this PR as per https://github.com/Icinga/icinga-notifications/pull/176#discussion_r1568815124 and https://github.com/Icinga/icinga-notifications/pull/176#discussion_r1570743030.

julianbrost (approved):
I've changed the issue reference from "closes" to "refs" as that ticket does describe the `event_id NOT NULL`, so the PR does not fully resolve that issue (as we decided).

PR #181, opened 2024-04-18
Use standard library instead of x/exp/slices

julianbrost:
That package is part of the standard library by now, so just use it instead.

`golang.org/x/exp` seems to be still used indirectly by some of our dependencies, so it just moves within `go.mod` instead of being removed.


oxzi (approved):
The three usages of `slice` are identical between the old `x/exp/slices` and the `slices` package.

PR #186, opened 2024-04-24
utils.IterateOrderedMap for plugin.FormatMessage

### LINKED ISSUE #177: plugin.FormatMessage: Sort `req.Object.{,Extra}Tags`
The `FormatMessage` function within the `plugin` package creates a textual representation of a `NotificationRequest`. This includes `map` fields - like the mentioned `.Object.{,Extra}Tags` - which are being accessed by `range`, resulting in a non-deterministic order and ~non-comparable~ harder to compare outputs.

I would suggest sorting the output alphabetically.

As an example, the following two outputs refer to the exact same `NotificationRequest`.

```
Info: Life is to you a dashing and bold adventure.

When: 2024-04-16 08:11:19 UTC

Object: http://localhost/icingaweb2/icingadb/service?name=random%20fortune&host.name=dummy-485

Tags:
host: dummy-485
service: random fortune

Extra Tags:
servicegroup/location-london: 
hostgroup/app-swift: 
hostgroup/department-dev: 
hostgroup/env-qa: 
hostgroup/location-rome: 
servicegroup/app-xrp: 
servicegroup/department-dev: 
servicegroup/env-prod: 

Incident: http://localhost/icingaweb2/notifications/incident?id=1018
```

```
Info: Life is to you a dashing and bold adventure.

When: 2024-04-16 08:11:19 UTC

Object: http://localhost/icingaweb2/icingadb/service?name=random%20fortune&host.name=dummy-485

Tags:
service: random fortune
host: dummy-485

Extra Tags:
hostgroup/department-dev: 
hostgroup/env-qa: 
hostgroup/location-rome: 
servicegroup/app-xrp: 
servicegroup/department-dev: 
servicegroup/env-prod: 
servicegroup/location-london: 
hostgroup/app-swift: 

Incident: http://localhost/icingaweb2/notifications/incident?id=1018
```

oxzi:
The "Tags" and "Extra Tags" printed in the `plugin.FormatMessage` were directly read from their map, thus having no order. This resulted in the same `NotificationRequest` being represented by different messages due to the unordered map.

This change was the result of investigating Go's new [rangefunc experiment](https://go.dev/wiki/RangefuncExperiment). The utilization of this novel language feature - which can also be indirectly used in the absence of `GOEXPERIMENT=rangefunc` - ensures that the map is traversed in key order.

Closes #177.

### DIFF
--- a/internal/utils/utils_test.go
+++ b/internal/utils/utils_test.go
@@ -26,3 +26,52 @@ func TestRemoveNils(t *testing.T) {
 		})
 	}
 }
+
+func TestIterateOrderedMap(t *testing.T) {
+	tests := []struct {
+		name    string
+		in      map[int]string
+		outKeys []int
+	}{
+		{"empty", map[int]string{}, nil},
+		{"single", map[int]string{1: "foo"}, []int{1}},
+		{"few-numbers", map[int]string{1: "a", 2: "b", 3: "c"}, []int{1, 2, 3}},
+		{
+			"1k-numbers",
+			func() map[int]string {
+				m := make(map[int]string)
+				for i := 0; i < 1000; i++ {
+					m[i] = "foo"
+				}
+				return m
+			}(),
+			func() []int {
+				keys := make([]int, 1000)
+				for i := 0; i < 1000; i++ {
+					keys[i] = i
+				}
+				return keys
+			}(),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var outKeys []int
+
+			// Either run with GOEXPERIMENT=rangefunc or wait for rangefuncs to land in the next Go release.
+			// for k, _ := range IterateOrderedMap(tt.in) {
+			// 	outKeys = append(outKeys, k)
+			// }
+
+			// In the meantime, it can be invoked as follows.
+			IterateOrderedMap(tt.in)(func(k int, v string) bool {
+				assert.Equal(t, tt.in[k], v)
+				outKeys = append(outKeys, k)
+				return true
+			})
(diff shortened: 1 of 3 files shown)


internal/utils/utils_test.go
```diff
@@ -26,3 +26,51 @@ func TestRemoveNils(t *testing.T) {
...
+			// Either run with GOEXPERIMENT=rangefunc or wait for rangefuncs to land in the next Go release.
+			// for k, _ := range IterateOrderedMap(tt.in) {
+			// 	outKeys = append(outKeys, k)
+			// }
+
+			// In the meantime, it can be invoked as follows.
+			IterateOrderedMap(tt.in)(func(k int, _ string) bool {
+				outKeys = append(outKeys, k)
+				return true
+			})
```
julianbrost, line 71: One-line addition to additionally check that the function passes the values correctly:
```suggestion
			IterateOrderedMap(tt.in)(func(k int, v string) bool {
				assert.Equal(t, )
				outKeys = append(outKeys, tt.in[k], v)
				return true
			})
```

internal/utils/utils_test.go
```diff
@@ -26,3 +26,51 @@ func TestRemoveNils(t *testing.T) {
...
+
+			assert.Equal(t, outKeys, tt.outKeys)
```
julianbrost, line 73: The two values should be swapped, currently expected/actual is mixed up.
```go
func Equal(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) bool
```

julianbrost (changes requested):

julianbrost (approved):

PR #193, opened 2024-05-06
schema: Change as needed for the new rotations

nilmerg:
**rotation**

Holds the configuration of a specific rotation.

*priority*: As the comment states, the lower the more important. Meaning: The shifts described by a rotation with a higher priority (i.e. lower integer) overrides others with a lower one:

image

weekends: 2
anytime: 1
dayshift: 0

*mode*: The configuration type. It describes the pattern the timeperiod entries, associated to each member, have.

**rotation_member**

Maps a member of a rotation to a specific contact or contactgroup.

*position*: Defines the order the member placed at inside the rotation.

image

Though, this is probably not of interest to the daemon. As *timeperiod_entry* also describes it.

**timeperiod_entry**

Every member of a rotation has at least one entry associated with it inside the timeperiod of a rotation.

The order of a rotation's members is incorporated in the definition of their entries here.

The daemon just has to interpret them, based on the priority of their rotation, to know who's on duty. Look at the first screenshot again, as the *Result* is what the intended effect should be.

### DIFF
--- a/schema/pgsql/schema.sql
+++ b/schema/pgsql/schema.sql
@@ -123,20 +123,6 @@ CREATE TABLE timeperiod (
     CONSTRAINT pk_timeperiod PRIMARY KEY (id)
 );
 
-CREATE TABLE timeperiod_entry (
-    id bigserial,
-    timeperiod_id bigint NOT NULL REFERENCES timeperiod(id),
-    rotation_member_id bigint REFERENCES rotation_member(id), -- nullable for future standalone timeperiods
-    start_time bigint NOT NULL,
-    end_time bigint NOT NULL,
-    -- Is needed by icinga-notifications-web to prefilter entries, which matches until this time and should be ignored by the daemon.
-    until_time bigint,
-    timezone text NOT NULL, -- e.g. 'Europe/Berlin', relevant for evaluating rrule (DST changes differ between zones)
-    rrule text, -- recurrence rule (RFC5545)
-
-    CONSTRAINT pk_timeperiod_entry PRIMARY KEY (id)
-);
-
 CREATE TABLE rotation_member (
     id bigserial,
     rotation_id bigint NOT NULL REFERENCES rotation(id),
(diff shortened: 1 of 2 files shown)


julianbrost:
> as needed for the new rotations

Is there an issue in the web project that explains how those are supposed to work that you can reference here?

How is all of this supposed to be interpreted? What needs more explanation: `rotation.priority`[^1], `rotation.mode` and `rotation_member.position` (especially: how do we know the current position?)

[^1]: I can imagine it based on mock-ups I was shown before, it should be written down nonetheless.

julianbrost:
> _mode_: The configuration type. It describes the pattern the timeperiod entries, associated to each member, have.

Is `rotation.mode` in any way relevant to derive the final schedule? Or is all of that information (also) contained in the individual entries?

Is it correct as a very high-level summary that with this, a new schedule is now a list of what previously was a schedule (and is now called rotation)? So to resolve the schedule to recipients at a specific time, evaluate each rotation like an old schedule was evaluated. Do this for each rotation in the order given by `rotation.priority` until the one gives a result.

nilmerg:
> Or is all of that information (also) contained in the individual entries?

Yes. Think of a `24/7` rotation with a handoff every 2 weeks and two members, to be associated with entries that have a duration of 2 weeks, each with a start date, both 2 weeks apart and a recurrence frequency of 4 weeks.

> Is it correct as a very high-level summary...

I'd rephrase

> in the order given by `rotation.priority`

with

> in ascending order based on `rotation.priority`

julianbrost:
As just discussed with @nilmerg: this change makes the `timeperiod` table irrelevant for the daemon for the moment (it can select everything needed following the `schedule` -> `rotation` -> `rotation_member` -> `timeperiod_entry` relations). It's kept though for possible future use (having a holiday time period that can be used to restrict per-user availability for example) and also as it seems to allow Web to perform some more efficient updates.

schema/pgsql/schema.sql
```diff
@@ -90,45 +90,66 @@ CREATE TABLE schedule (
...
-    frequency frequency_type,
-    description text,
 
     CONSTRAINT pk_timeperiod_entry PRIMARY KEY (id)
 );
 
-CREATE TABLE schedule_member (
-    schedule_id bigint NOT NULL REFERENCES schedule(id),
-    timeperiod_id bigint NOT NULL REFERENCES timeperiod(id),
+CREATE TABLE rotation_member (
```
sukhwinder33445, line 135: Table `rotation_member` must be created before `timeperiod_entry.`
Same in upgrade file.

sukhwinder33445 (changes requested):

schema/pgsql/schema.sql
```diff
@@ -90,45 +90,71 @@ CREATE TABLE schedule (
...
+    options text NOT NULL,
+
+    -- A date in the format 'YYYY-MM-DD' when the first handoff should happen.
+    -- It is a string as handoffs are restricted to happen only once per day
+    first_handoff date NOT NULL,
+
+    -- Set in case the first_handoff was in the past during creation of the rotation.
+    -- It is essentially the creation time of the rotation.
+    -- Used by Web to avoid showing shifts that never happened
+    actual_handoff bigint,
```
julianbrost, line 111: > It is essentially the creation time of the rotation.

Even if `first_handoff` is set to be in the future when creating the rotation?

nilmerg: No, in that case it's NULL

julianbrost: Can you always set it? That would actually make it useful as it would allow to easily determine the currently active version of the rotation (the one with the largest value of `actual_handoff`, that is at or before the current time).

julianbrost: Only if the switchover to the newer version happens at a fixed time like 00:00 UTC. Otherwise, it misses timezone and/or time of day information.

nilmerg: Changed it, by the way

schema/pgsql/schema.sql
nilmerg: fixed

PR #198, opened 2024-05-16
Code enhancements for future use

yhabteab:
This PR proposes the following improvements in preparation for #188.

- 9502b20958f6cbd085952a41e8c05b771096ebda decouples `contactChannels` from the `incident` package, since we're going to use this recipient channels map elsewhere outside the `incident` package in the future and it seems reasonable to me to move it here to the `rule` package.
- d238b0bf33cf0781a2965745f27ccc88fa20d6a4 implements [`zapcore.ObjectMarshaler`](https://pkg.go.dev/go.uber.org/zap/zapcore#ObjectMarshaler) for the `rule.Escalation` and `recipient.Key` types and allows us to use `zap.Inline(escalation)` or `zap.Object("escalation", escalation)`wherever more logging context is needed without having to add all the individual fields ourselves each time.
- ce8732e7e5de9739a11621293da9e743e1e43233 moves the `ErrSuperfluousStateChange` error type to the `event` package in order to mitigate a cyclic import error with the upcoming changes to non-state event processing support. It kind of also fit into this package.
- 706ed389af7c054545662405965900a00839fae1 drops `incident#AddHistory()` method as it was already quite annoying having to return `types.Int` every single time this method was used. This commit embeds the functionality for `insert and fetch ID` directly into the `HistoryRow` type and nowhere needs to be returned.
- 7ef9aa25f3b82bc8076bca4f5ec483661f0c66ae drops the `create` param from `incident#ProcessEvent()` and `incident#GetCurrent()`, since we can easily determine whether an incident is new based on its `StartedAt` timestamp, and thus, we don't need to pass this info around methods. (Added @sukhwinder33445 as a co-author as the original [idea was from him](https://github.com/Icinga/icinga-notifications/pull/137/commits/2eb858b1994f10bfad74552fa64848126fb01a71), but I've implemented it in a slightly different way).

### DIFF
--- a/internal/config/rule.go
+++ b/internal/config/rule.go
@@ -57,13 +57,7 @@ func (r *RuntimeConfig) fetchRules(ctx context.Context, tx *sqlx.Tx) error {
 
 	escalationsByID := make(map[int64]*rule.Escalation)
 	for _, escalation := range escalations {
-		escalationLogger := r.logger.With(
-			zap.Int64("id", escalation.ID),
-			zap.Int64("rule_id", escalation.RuleID),
-			zap.String("condition", escalation.ConditionExpr.String),
-			zap.String("name", escalation.NameRaw.String),
-			zap.Int64("fallback_for", escalation.FallbackForID.Int64),
-		)
+		escalationLogger := r.logger.With(zap.Inline(escalation))
 
 		rule := rulesByID[escalation.RuleID]
 		if rule == nil {
@@ -85,10 +79,6 @@ func (r *RuntimeConfig) fetchRules(ctx context.Context, tx *sqlx.Tx) error {
 			// TODO: implement fallbacks (needs extra validation: mismatching rule_id, cycles)
 			escalationLogger.Warnw("ignoring fallback escalation (not yet implemented)")
 			continue
-		}
-
-		if escalation.NameRaw.Valid {
-			escalation.Name = escalation.NameRaw.String
 		}
 
 		rule.Escalations[escalation.ID] = escalation
@@ -165,11 +155,11 @@ func (r *RuntimeConfig) applyPendingRules() {
 					recipientLogger := r.logger.With(
 						zap.Int64("id", recipient.ID),
 						zap.Int64("escalation_id", recipient.EscalationID),
-						zap.Int64("channel_id", recipient.ChannelID.Int64))
+						zap.Int64("channel_id", recipient.ChannelID.Int64),
+						zap.Inline(recipient.Key))
 
 					if recipient.ContactID.Valid {
 						id := recipient.ContactID.Int64
-						recipientLogger = recipientLogger.With(zap.Int64("contact_id", id))
 						if c := r.Contacts[id]; c != nil {
 							recipient.Recipient = c
 						} else {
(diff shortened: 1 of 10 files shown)


internal/rule/rule.go
```diff
@@ -16,3 +18,40 @@ type Rule struct {
...
+func (ch ContactChannels) LoadFromEscalationRecipients(escalation *Escalation, t time.Time, isNotifiable func(recipient.Key) bool) {
+	for _, escalationRecipient := range escalation.Recipients {
+		ch.LoadRecipientChannel(escalationRecipient, t, isNotifiable)
+	}
+}
+
+// LoadRecipientChannel loads recipient channel to the current map.
+// You can provide this method a callback to control whether the channel of a specific contact should
+// be loaded, and it will skip those for whom the callback returns false. Pass IsNotifiable for default actions.
+func (ch ContactChannels) LoadRecipientChannel(recipients *EscalationRecipient, t time.Time, isNotifiable func(recipient.Key) bool) {
```
julianbrost, line 37: It's just one `recipient`, so the parameter should be named like that.

internal/config/rule.go
```diff
@@ -165,11 +159,10 @@ func (r *RuntimeConfig) applyPendingRules() {
 					recipientLogger := r.logger.With(
 						zap.Int64("id", recipient.ID),
 						zap.Int64("escalation_id", recipient.EscalationID),
-						zap.Int64("channel_id", recipient.ChannelID.Int64))
+						zap.Int64("channel_id", recipient.ChannelID.Int64), zap.Inline(recipient.Key))
```
julianbrost, line 162: It's one line per parameter here so far. Please keep it that way.

internal/rule/escalation.go
```diff
@@ -21,6 +22,30 @@ type Escalation struct {
...
+// https://pkg.go.dev/go.uber.org/zap/zapcore#ObjectMarshaler
+func (e *Escalation) MarshalLogObject(encoder zapcore.ObjectEncoder) error {
+	encoder.AddInt64("id", e.ID)
+	encoder.AddInt64("rule_id", e.RuleID)
+
+	if e.NameRaw.Valid && e.NameRaw.String != "" {
+		encoder.AddString("name", e.NameRaw.String)
+	} else {
+		encoder.AddString("name", e.DisplayName())
+	}
```
julianbrost, line 38: Why does this need to access `NameRaw` here instead of just relying on `DisplayName()`?

internal/incident/db_types.go
```diff
@@ -85,6 +89,26 @@ func (h *HistoryRow) TableName() string {
...
+		}
+
+		h.ID = historyId
+
+		return nil
+	}
+
+	_, err := tx.NamedExecContext(ctx, stmt, h)
+	return err
+}
```
julianbrost, line 110: I'd just assume `fetchID = true` here and remove the parameter. It's not like that would be much more expensive for the database, so this mostly just increases code complexity here.

julianbrost (changes requested):

internal/rule/escalation.go
yhabteab: Because if `zap.Inline(escalation)` is used from within the runtime config, `escalation.Name` could still be empty, while `NameRaw` is not. 
https://github.com/Icinga/icinga-notifications/blob/454502ef05c86e944f14705f896710ab03e1fc44/internal/config/rule.go#L90-L92

julianbrost: The `Name` attribute is only used in `DisplayName()` and where it's initialized, so it really isn't needed (for where the name is needed, `DisplayName()` should be used). Thus, I'd just remove the current `Name` member, and move what's done here into `DisplayName()`.

yhabteab: Done!

julianbrost (approved):

PR #199, opened 2024-05-23
Code enhancements for future use v2

yhabteab:
This PR is a follow-up to/basically the same as https://github.com/Icinga/icinga-notifications/pull/198 and proposes the following changes.

- 7c42f14a47c75d38de208d3d8e196334b6dcfdea Implements the [`zapcore.ObjectMarshaler`](https://pkg.go.dev/go.uber.org/zap/zapcore#ObjectMarshaler) interface also for the `rule.Rule` type and have adjusted many of its usages already.
- 349c9940c1c10f14e954dffb1d726af6e0943242 Introduces a new method `Eval()` for the `rule.Rule` and `rule.Escalation` types, which will be responsible for evaluating their respective filters/conditions, and the caller will not have to check beforehand whether a particular rule/escalation has a condition/filter that needs to be evaluated.


oxzi (approved):

internal/rule/rule.go
```diff
@@ -19,6 +20,33 @@ type Rule struct {
 	Escalations      map[int64]*Escalation
 }
 
+// MarshalLogObject implements the zapcore.ObjectMarshaler interface.
+func (r *Rule) MarshalLogObject(encoder zapcore.ObjectEncoder) error {
+	encoder.AddInt64("id", r.ID)
+	encoder.AddString("name", r.Name)
+	// TODO: uncomment this once #188 is merged
+	// encoder.AddString("type", r.Type.String())
```
julianbrost, line 28: If #188 adds a field that should be logged, that should be added as part of this PR, no need to add a TODO before. I doubt anyone searches the full code for TODO comments referring the current PR when reviewing it anyways.

internal/incident/incident.go
```diff
@@ -429,24 +427,15 @@ func (i *Incident) evaluateEscalations(eventTime time.Time) ([]*rule.Escalation,
...
+					i.logger.Warnw(
+						"Failed to evaluate escalation condition", zap.Object("rule", r),
+						zap.Object("escalation", escalation), zap.Error(err),
+					)
+				} else if !matched {
+					incidentAgeFilter := filterContext.ReevaluateAfter(escalation.Condition)
+					retryAfter = min(retryAfter, incidentAgeFilter)
 				}
 
 				if matched {
```
julianbrost, line 441: The existing `if` below can be combined with the if before. That's also a bit nicer because then `matched` will simply be ignored if an error was returned.

julianbrost (changes requested):

oxzi (approved):

PR #200, opened 2024-05-24
Incident muting & notification suppression

### LINKED ISSUE #190: Incident muting (for downtimes, flapping, acknowledgements)
Reviewing the code of #146 and testing it showed some issues that where the proper solution requires some kind of incident muting mechanism.

1. As we don't track any kind of in-downtime state, downtime notifications get lost if we're not connected to Icinga 2 while the event happens as the event stream catch-up code has no last known state that it could compare the current state from the Icinga 2 API with.
2. The current implementation does not properly open incidents after a downtime ended as it just completely dismisses the state change event. Doing this properly requires processing the event but not sending notifications, which is what this issue asks for: muting.
3. State change events from the event stream API don't include whether the checkable is flapping so this is currently not handled. That could be solved by flapping just muting the incident (or the object, as it might flap between OK and something else, yet another question).

yhabteab:
resolves #190


internal/icinga2/api_responses.go
```diff
@@ -176,40 +181,29 @@ type StateChange struct {
 	CheckResult     CheckResult `json:"check_result"`
 	DowntimeDepth   int         `json:"downtime_depth"`
 	Acknowledgement bool        `json:"acknowledgement"`
+	IsFlapping      bool        `json:"-"` // TODO: Need a workaround for this!! Sending extra API query is ridiculous.
```
yhabteab, line 184: I'm quite unsure about this one as the Icinga 2 event streams for `StateChange` don't contain that bit information, so I'm just clueless as to how to proceed here. Apart from that and the handling of the `ACK' events, this PR should just work smoothly.

internal/incident/history_event_type.go
```diff
@@ -11,6 +11,8 @@ const (
 	HistoryEventTypeNull HistoryEventType = iota
 
 	Opened
+	Muted
+	UnMuted
```
oxzi, line 15: How about this?

```suggestion
	Unmuted
```

internal/incident/incident.go
```diff
@@ -147,6 +151,11 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event) error {
 		return errors.New("can't insert incident event to the database")
 	}
 
+	if err := i.handleMuteUnMute(ctx, tx, ev); err != nil {
+		i.logger.Errorw("Cannot insert incident muted history", zap.String("event", ev.String()), zap.Error(err))
+		return errors.New("cannot insert incident muted history")
```
oxzi, line 156: Why not wrap the original error?

internal/incident/incident.go
```diff
@@ -173,13 +182,22 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event) error {
 		}
 	case event.TypeAcknowledgementSet:
 		if err := i.processAcknowledgementEvent(ctx, tx, ev); err != nil {
+			if errors.Is(err, errSuperfluousAckEvent) {
+				// That ack error type indicates that the acknowledgement author was already a manager,
+				// thus we can safely that event and return without even committing the DB transaction.
+				return nil
```
oxzi, line 188: Maybe add at least a debug log message?

internal/incident/incident.go
```diff
@@ -346,6 +367,31 @@ func (i *Incident) processIncidentOpenedEvent(ctx context.Context, tx *sqlx.Tx,
 	return nil
 }
 
+// handleMuteUnMute generates an incident Muted or UnMuted history based on the Object state.
+// Returns an error if fails to insert the generated history to the database.
+func (i *Incident) handleMuteUnMute(ctx context.Context, tx *sqlx.Tx, ev *event.Event) error {
```
oxzi, line 372:
```suggestion
// handleMuteUnmute generates an incident Muted or Unmuted history based on the Object state.
// Returns an error if fails to insert the generated history to the database.
func (i *Incident) handleMuteUnmute(ctx context.Context, tx *sqlx.Tx, ev *event.Event) error {
```

oxzi (commented):
This was a first inspection of the change. Please feel free to comment if you find something unfitting.

internal/event/event.go
```diff
@@ -84,6 +90,11 @@ func (e *Event) Validate() error {
 	}
 }
 
+// SetMute sets the event Mute field to the given value.
+func (e *Event) SetMute(val bool) {
+	e.Mute = types.Bool{Valid: true, Bool: val}
+}
```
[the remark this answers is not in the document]
yhabteab: The event `MuteReason` can also be set while `Mute` isn't, so your first suggestion looks nice 👍 .

internal/icinga2/client.go
```diff
@@ -163,36 +207,62 @@ func (client *Client) buildHostServiceEvent(ctx context.Context, result CheckRes
...
+	ev, err := client.buildCommonEvent(ctx, ack.Host, ack.Service)
 	if err != nil {
 		return nil, err
 	}
 
-	if clearEvent {
+	ev.Username = ack.Author
+	if ack.EventType == typeAcknowledgementCleared {
 		ev.Type = event.TypeAcknowledgementCleared
+		ev.MuteReason = "Acknowledgement cleared"
```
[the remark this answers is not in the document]
yhabteab: An `Acknowledgement-cleared` event does not mute an object, but unmutes it if possible, which is set to `false` just a few lines below if the if condition matches.

internal/icinga2/client_api.go
```diff
@@ -309,9 +320,19 @@ func (client *Client) connectEventStream(esTypes []string) (io.ReadCloser, error
...
+		// so that our filter expressions such as ‘&’ don't get replaced by some UTF-8 code points.
+		encoder.SetEscapeHTML(false)
+
+		err := encoder.Encode(map[string]any{
 			"queue": fmt.Sprintf("icinga-notifications-%x", queueNameRndBuff),
 			"types": esTypes,
+			"filter": fmt.Sprintf(
+				"(event.type!=%q && event.type!=%q) || event.object_type==%q || event.object_type==%q",
+				typeObjectCreated, typeObjectDeleted, "Host", "Service",
+			),
```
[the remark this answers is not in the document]
yhabteab: > Object created for all Icinga 2 objects.

This is a quote from the Icinga 2 event stream docs for the `ObjectCreated` event and means that this event is triggered for every single object of Icinga 2 of any type, but I don't really care about any other types than `Host` and `Service`. Therefore, I want to check if `event.type` is not `ObjectCreated` and not `ObjectDeleted` (this means that the event type could be any other type listed in `esTypes`), otherwise (this means that `event.type` is either `ObjectCreated` or `ObjectDeleted`), I want to additionally filter on the `object_type` and match only on the `Host` or `Service` types.

internal/incident/incident.go
yhabteab: The original error is logged with `zap.Error()` and we shouldn't be returning database-related errors to the clients, but #188 may make this more consistent.

internal/incident/incident.go
yhabteab: Done!

internal/icinga2/client.go
```diff
@@ -163,36 +207,62 @@ func (client *Client) buildHostServiceEvent(ctx context.Context, result CheckRes
...
+		if !result.IsFlapping && result.Acknowledgement <= 0 && result.DowntimeDepth <= 0 {
+			// Only unmute the object if there are no other reasons to keep it muted.
+			ev.SetMute(false)
+		}
 	} else {
 		ev.Type = event.TypeAcknowledgementSet
+		ev.SetMute(true)
+		ev.Message = ack.Comment
+		ev.MuteReason = fmt.Sprintf("Checkable acknowledged by %q: %s", ack.Author, ack.Comment)
 	}
```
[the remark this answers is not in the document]
yhabteab: Honestly, the `EventType` is not something manual but is being decoded from Icinga 2 API streams and I wouldn't expect it to be set to anything else.
https://github.com/Icinga/icinga-notifications/blob/7b542e34488cf9c1ce29231c8eab6e5063403750/internal/icinga2/api_responses.go#L315-L317

Adding error handling here and some other similar places is IMHO pointless as we don't actually need to.

internal/icinga2/client.go
```diff
@@ -64,6 +65,12 @@ type Client struct {
 	eventDispatcherEventStream chan *eventMsg
 	// catchupPhaseRequest requests the main worker to switch to the catch-up-phase to query the API for missed events.
 	catchupPhaseRequest chan struct{}
+
+	// eventExtraTags is used to cache Checkable groups once they have been fetched from the Icinga 2 API so that they
+	// don't have to be fetched over again with each ongoing event. Host/Service groups are never supposed to change at
+	// runtime, so this cache is being refreshed once in a while when Icinga 2 dispatches an object created/deleted
+	// event and thus should not overload the Icinga 2 API in a large environment with numerous Checkables.
+	eventExtraTags map[string]map[string]string
```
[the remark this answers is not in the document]
yhabteab: Like discussed offline, I'm not entirely sure what your implementation of this cache looked like, but unless I'm overlooking the 🐘, with the current implementation, once an object's extra tags are cached, it should should never submit an API request again to retrieve that object's groups, I don't have any metrics to showcase though.

yhabteab:
Have just rebased the PR to resolve the conflicts resulting from #114.

internal/icinga2/client.go
```diff
@@ -441,6 +555,14 @@ func (client *Client) Process() {
 	client.eventDispatcherEventStream = make(chan *eventMsg)
 	client.catchupPhaseRequest = make(chan struct{})
 
+	cache, err := lru.New[string, map[string]string](int(math.Exp2(17)))
```
oxzi, line 558:
```suggestion
	cache, err := lru.New[string, map[string]string](1 << 17)
```

Doing so will also get rid of the `math` import.

internal/icinga2/client.go
```diff
@@ -132,17 +138,53 @@ func (client *Client) buildCommonEvent(ctx context.Context, host, service string
...
+			queryResult, err := client.fetchCheckableCurrentState(ctx, result.GetHostName(), result.GetServiceName())
+			if err != nil {
+				return nil, err
+			}
+			for _, serviceGroup := range queryResult.Attrs.Groups {
+				extraTags["servicegroup/"+serviceGroup] = ""
+			}
+		}
+
+		client.eventExtraTagsCache.Add(result.ObjectName, extraTags)
```
oxzi, line 173: Shouldn't a `return extraTags, nil` be present at this point?

yhabteab: 🐘

internal/incident/incident.go
```diff
@@ -173,13 +182,22 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event) error {
...
+
 			return err
 		}
 	}
 
-	notifications, err := i.addPendingNotifications(ctx, tx, ev, i.getRecipientsChannel(ev.Time))
-	if err != nil {
-		return err
+	var notifications []*NotificationEntry
+	if ev.Type != event.TypeDowntimeStart && ev.Type != event.TypeAcknowledgementSet && ev.Type != event.TypeFlappingStart {
```
oxzi, line 196: This intentionally suppresses sending the event suppressing further events, e.g., the Downtime Start. However, the Downtime End notification will then be sent. I would expect to be notified that further notifications are stopped as Icinga 2 internal notification for starting downtimes.

oxzi (dismissed):
I have successfully tested the PR!
@julianbrost, please take another look.

yhabteab:
I have just rebased it and committed new changes separately 246f0f0.

oxzi (dismissed):

yhabteab:
Once again, I have found a bug where pending notifications, e.g. triggered by a `downtime-start` event, are not being sent ab7e482.

internal/incident/incident.go
```diff
@@ -173,13 +182,30 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event) error {
...
+	// Otherwise, we're going to generate suppressed/pending notifications history for all the remaining event types.
+	switch {
+	case !i.isMuted || !i.Object.IsMuted():
+		fallthrough
+	case ev.Type != event.TypeDowntimeStart && ev.Type != event.TypeAcknowledgementSet && ev.Type != event.TypeFlappingStart:
+		notifications, err = i.generateNotifications(ctx, tx, ev, i.getRecipientsChannel(ev.Time))
+		if err != nil {
+			return err
+		}
 	}
```
julianbrost, line 209: The comment primarily repeats the code but doesn't say why it does what it does, especially why some event types are treated specially.

Also, I don't really understand the overall picture. This reads like this `switch` is supposed to determine whether notifications are sent, on the other hand, `generateNotifications()` determines internally whether notifications are suppressed?

internal/event/event.go
```diff
@@ -34,6 +34,9 @@ type Event struct {
 	Username string   `json:"username"`
 	Message  string   `json:"message"`
 
+	Mute       types.Bool `json:"mute"`
+	MuteReason string     `json:"mute_reason"`
```
julianbrost, line 38: I'd expect these to appear in the database as well.

julianbrost (changes requested):

internal/icinga2/client_api.go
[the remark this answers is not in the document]
yhabteab: > Something like the following should also work as a filter:

Did you test it? Otherwise you should have already noticed that this doesn't work.

> I think this would be more readable if it was a little more verbose if expression.

This is a simple API filter expression, I'm not sure what exactly you find it to be *unreadable*, and apart from that I just didn't want to use the literal `ObjectCreated` while there is a constant `typeObjectCreated` for it, and **no**, you can't just use `event.object_type == "Host" || event.object_type == "Service"` or `(event.type == "ObjectCreated" || event.type == "ObjectDeleted")` as a filter value, you have to use quoted values `"\"FOO\""`.

internal/icinga2/client_api.go
```diff
@@ -329,9 +340,19 @@ func (client *Client) connectEventStream(esTypes []string) (io.ReadCloser, error
 		queueNameRndBuff := make([]byte, 16)
 		_, _ = rand.Read(queueNameRndBuff)
 
-		reqBody, err := json.Marshal(map[string]any{
+		reqBody := new(bytes.Buffer)
+		encoder := json.NewEncoder(reqBody)
+		// Since our encoded JSON isn't going to be used somewhere in a browser, we've to disable this feature
+		// so that our filter expressions such as ‘&’ don't get replaced by some UTF-8 code points.
+		encoder.SetEscapeHTML(false)
```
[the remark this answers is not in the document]
yhabteab: > Did this cause actual problems?

Yes, this does cause problems, or does Icinga 2 html-unescape the request body?

internal/icinga2/api_responses.go
```diff
@@ -264,13 +263,52 @@ type DowntimeTriggered struct {
...
+
+// GetServiceName returns the service of this current response result.
+// Returns empty string if the current result type is of type `Host`.
+func (c *CheckableCreatedDeleted) GetServiceName() string {
+	if c.ObjectType == "Service" {
+		return strings.Split(c.ObjectName, "!")[1]
+	}
+
+	return ""
 }
```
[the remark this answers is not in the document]
yhabteab: > This implementation relies on these events being filtered to hosts and services only.

There are only two types of Checkable objects, `Host` and `Service`, and that type is named after these, why would you want to use it with anything else?

schema/pgsql/schema.sql
```diff
@@ -4,6 +4,8 @@ CREATE TYPE boolenum AS ENUM ( 'n', 'y' );
 CREATE TYPE incident_history_event_type AS ENUM (
     -- Order to be honored for events with identical millisecond timestamps.
     'opened',
+    'muted',
+    'unmuted',
     'incident_severity_changed',
     'rule_matched',
     'escalation_triggered',
```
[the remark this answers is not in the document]
yhabteab: Because this is their logical order in which these histories are generated in the code.

internal/incident/incident.go
yhabteab: > This reads like this switch is supposed to determine whether notifications are sent, on the other hand, generateNotifications() determines internally whether notifications are suppressed?

No, this only determines whether `suppressed` or `pending` notifications should be generated at all, and `generateNotifications()` determines on its own whether the generated notifications should be of type `suppressed`.

internal/icinga2/client_api.go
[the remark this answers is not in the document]
julianbrost: > > Something like the following should also work as a filter:
> 
> Did you test it? Otherwise you should have already noticed that this doesn't work.

I didn't test it within the code here, I just verified in `icinga2 console` that `if` expressions indeed evaluate to a value so I'd expect it to work here as well:

```
Icinga 2 (version: v2.14.2)
Type $help to view available commands.
<1> => if (true) { 23 } else { 42 }
23.000000
<2> => if (false) { 23 } else { 42 }
42.000000
```

> > I think this would be more readable if it was a little more verbose if expression.
> 
> This is a simple API filter expression, I'm not sure what exactly you find it to be _unreadable_, and apart from that I just didn't want to use the literal `ObjectCreated` while there is a constant `typeObjectCreated` for it

That comment was primarily about the structure of the condition. Isn't the purpose of the filter to perform a check for some event types and leave the other ones unfiltered? I think such an `if` would convey this in a more straight-forward way.

> and **no**, you can't just use `event.object_type == "Host" || event.object_type == "Service"` or `(event.type == "ObjectCreated" || event.type == "ObjectDeleted")` as a filter value, you have to use quoted values `"\"FOO\""`.

Yes, you can, you can alternatively enclose string literals in ``` `` ``` and then don't need to escape double quotes.

internal/icinga2/client_api.go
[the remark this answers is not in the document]
julianbrost: The escaping is not HTML-specific, Go just replaces characters that might be unsafe to use in HTML with their `\uXXXX` representation by default. This is not required by JSON but still results in perfectly valid JSON. Icinga 2 can also parse this:

```
Icinga 2 (version: v2.14.2)
Type $help to view available commands.
<1> => Json.decode({{{"foo\u0026bar"}}})
"foo&bar"
```

internal/icinga2/api_responses.go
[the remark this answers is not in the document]
julianbrost: There are no other types of checkables, but other types of objects. This PR assumes that the `ObjectCreated` and `ObjectDeleted` event stream messages won't be used for anything but hosts and services:
 
https://github.com/Icinga/icinga-notifications/blob/ab7e482b635236aa147c4aa30a97cda1f8535988/internal/icinga2/api_responses.go#L362-L363

https://github.com/Icinga/icinga-notifications/blob/ab7e482b635236aa147c4aa30a97cda1f8535988/internal/icinga2/client_api.go#L352-L355

If the use of these event streams was extended to other types, this would result in strange behavior without closely checking all the code again (`GetHostName()` would assume it was a service, `GetServiceName()` would assume it was a host). On the other hand, if the event would just represent a generic ObjectCreated/Deleted event and the code consuming these events just checked if it's a host or service, this would be more robust.

internal/incident/incident.go
```diff
@@ -173,13 +182,31 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event) error {
...
+		// On the other hand, if the current incident and its object are currently muted, and we do not want to
+		// generate any notifications if the type of the ongoing event is either Downtime/Flapping-Start or ACK-set,
+		// but simply ignore them silently. For all other remaining event types, such as state changes etc., we are
+		// going to record "suppressed" notifications in the database, though not actually send them.
+	case ev.Type != event.TypeDowntimeStart && ev.Type != event.TypeAcknowledgementSet && ev.Type != event.TypeFlappingStart:
+		notifications, err = i.generateNotifications(ctx, tx, ev, i.getRecipientsChannel(ev.Time))
+		if err != nil {
+			return err
+		}
 	}
```
julianbrost, line 210: So, if I understand correctly, `i.isMuted != i.Object.IsMuted()` if and only if the current event mutes or unmutes the object. Thus, the first `case` triggers notifications if the object is being muted/unmuted by the current event, or if the object is not muted (and the current event doesn't change that).

The second `case` then only affects "nested" muting start events (we don't track nested muting, this is just saying that another "start muting" event happens while already muted). I presume the asymmetry that it checks for the start but not for the end events is because end events are only generated when all muting reasons are gone, i.e. there won't be a downtime end event while still being acknowledged?

Did I get that correctly so far?

internal/event/event.go
```diff
@@ -143,13 +155,15 @@ func (e *Event) Sync(ctx context.Context, tx *sqlx.Tx, db *database.DB, objectId
...
-	Message  types.String    `db:"message"`
+	ID         int64           `db:"id"`
+	Time       types.UnixMilli `db:"time"`
+	ObjectID   types.Binary    `db:"object_id"`
+	Type       types.String    `db:"type"`
+	Severity   Severity        `db:"severity"`
+	Username   types.String    `db:"username"`
+	Message    types.String    `db:"message"`
+	Mute       types.Bool      `json:"mute"`
+	MuteReason types.String    `json:"mute_reason"`
```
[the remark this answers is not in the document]
yhabteab: That's an oversight, in fact I don't even need to set the db tag since #114.

internal/object/object.go
```diff
@@ -25,10 +25,12 @@ var (
...
 type Object struct {
-	ID       types.Binary `db:"id"`
-	SourceID int64        `db:"source_id"`
-	Name     string       `db:"name"`
-	URL      types.String `db:"url"`
+	ID          types.Binary `db:"id"`
+	SourceID    int64        `db:"source_id"`
+	Name        string       `db:"name"`
+	URL         types.String `db:"url"`
+	Muted       types.Bool   `db:"is_muted"`
```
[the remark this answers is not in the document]
yhabteab: Go's `bool` type doesn't map `y, n` to boolean types, so I still have to use this.

internal/incident/incident.go
yhabteab: > Did I get that correctly so far?

Yes!

yhabteab: Was that a general question of clarification or do you want me to extend the existing comments?

julianbrost: So far only general questions. I'm trying to understand it so that I can see if I can come up with a way to simplify this.

> because end events are only generated when all muting reasons are gone

But now I'm confused about that part after looking at the code again:

https://github.com/Icinga/icinga-notifications/blob/a07add3c95d5a8a63d6885a24dff1f799de2e324/internal/icinga2/client.go#L347-L353

This looks like a flapping stop event would still be emitted even while still muted by another reason (code for ack clear and downtime end looks very similar), it just doesn't say unmute but keeps the current muting state. Is there any other place in the code that would stop the processing of these events?

However, emitting and processing all (including nested) {downtime,flapping,ack} {start,end} events sounds like a good idea, after all, we want to be able to notify for them in the future.

What would happen if the check here was something like this:

```go
if !i.isMuted || !i.Object.IsMuted() || ev.Type == event.TypeState {
	generateNotifications()
}
```

Shouldn't this behave very similar? I.e. notifications outside of any muting are sent as well as for the events actually muting oder unmuting. While being muted, no notifications are sent and only state events generate suppressed history events.

What would happen if `generateNotifications()` was just called here unconditionally and internally decides if a notification is sent or suppressed (just like it already does in the current state of the PR)? Shouldn't this result in a history like the following?

- Ack set (sent)
- Downtime start (suppressed, because already muted by ack)
- Downtime start (suppressed, because already muted)
- State changed / ack cleared (both suppressed, not entirely sure about the exact order, but unless sticky, the ack will be cleared due to the state change, but still in downtime, so still suppressed)
- Downtime end (suppressed, still in the other downtime)
- Downtime end (sent, because this removed the last remaining muting reason)

That would look perfectly fine for me now. And it might even be a good starting point for allowing sending some of these (for now) suppressed notifications.

schema/pgsql/upgrades/026.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TYPE incident_history_event_type ADD VALUE 'muted' AFTER 'opened';
+ALTER TYPE incident_history_event_type ADD VALUE 'unmuted' AFTER 'muted';
+ALTER TYPE notification_state_type ADD VALUE 'suppressed' BEFORE 'pending';
+
+ALTER TABLE object
+    ADD COLUMN is_muted boolenum NOT NULL DEFAULT 'n',
```
julianbrost, line 6: This leaves behind a default value for the column which is inconsistent with the full schema file.

internal/incident/incident.go
yhabteab: > Is there any other place in the code that would stop the processing of these events?

Why would you want to stop processing these? These are regular Icinga 2 events and not virtually generated ones, and as such they should exist somewhere in the database (in the `event` table) even if we aren't going to trigger notifications for them.

> What would happen if generateNotifications() was just called here unconditionally and internally decides if a notification is sent or suppressed (just like it already does in the current state of the PR)? Shouldn't this result in a history like the following?

In its initial version of this PR, that was exactly the case, and I did asked you explicitly about this (how we should handle two consecutive events), and you told me not even to generate a suppressed notification for these events, but rather to silently ignore them if the object was already muted by previous events. Yet, you are suggesting to restore this behaviour?

julianbrost: > > Is there any other place in the code that would stop the processing of these events?
> 
> Why would you want to stop processing these? These are regular Icinga 2 events and not virtually generated ones, and as such they should exist somewhere in the database (in the `event` table) even if we aren't going to trigger notifications for them.

I didn't want to suggest stopping to process these, but I was under the assumption that this was the case and the reason for why `case ev.Type != event.TypeDowntimeStart && ev.Type != event.TypeAcknowledgementSet && ev.Type != event.TypeFlappingStart` only checks for the start but not the end events.

> > What would happen if generateNotifications() was just called here unconditionally and internally decides if a notification is sent or suppressed (just like it already does in the current state of the PR)? Shouldn't this result in a history like the following?
> 
> In its initial version of this PR, that was exactly the case, and I did asked you explicitly about this (how we should handle two consecutive events), and you told me not even to generate a suppressed notification for these events, but rather to silently ignore them if the object was already muted by previous events. Yet, you are suggesting to restore this behaviour?

Sounds like there was a misunderstanding. I was under the impression that these would be generic mute/unmute events. Then, processing a second mute event wouldn't do something useful and I wouldn't have wanted to introduce another "mute reason changed" event. However, when those events still have distinct and useful types, I don't see a reason why they shouldn't be processed.

internal/incident/sync.go
```diff
@@ -150,23 +154,28 @@ func (i *Incident) addPendingNotifications(
...
+				return nil, errors.New("cannot insert incident notification history")
 			}
 
-			notifications = append(notifications, &NotificationEntry{
-				HistoryRowID: hr.ID,
-				ContactID:    contact.ID,
-				State:        NotificationStatePending,
-				ChannelID:    chID,
-			})
+			if !i.isMuted || !i.Object.IsMuted() {
```
julianbrost, line 171: I'd either do a `suppress := i.isMuted && i.Object.IsMuted()` at the beging of the function and use it for both `if`s, or use `hr.State == NotificationStatePending` as the condition for the last `if`.

internal/object/object.go
```diff
@@ -120,11 +124,28 @@ func FromEvent(ctx context.Context, db *database.DB, ev *event.Event) (*Object,
...
+		newObject.Name = ev.Name
+		newObject.URL = utils.ToDBString(ev.URL)
+
+		if ev.Mute.Valid && (!ev.Mute.Bool || !newObject.IsMuted()) {
+			// This behaviour is explicitly requested so that two successive events with `mute` set
+			// to true don't influence each other, even if their `mute_reason` differs. We should
+			// only care about the first mute event.
+			newObject.Muted = ev.Mute
+			newObject.MuteReason = utils.ToDBString(ev.MuteReason)
+		}
```
julianbrost, line 148: > This behaviour is explicitly requested

What would you do here if it wasn't?

Maybe another misunderstanding/miscommunication here. I didn't want to say "`mute_reason` may not be updated under any circumstances" but rather something like "I don't think we want to put the effort into providing useful events/notifications like 'the mute reason changed from being in downtime to being in downtime and acknowledged' as that would require keeping more detailed track of the mute reasons and sending a notification like 'the muting reason changed from downtime to acknowledgement' would be pointless if the downtime still continues to exist".

internal/icinga2/client.go
```diff
@@ -128,31 +129,81 @@ func (client *Client) buildCommonEvent(ctx context.Context, host, service string
...
+// it will fetch the groups from the Icinga 2 API, map them to the event extra tags, add them to the client
+// Cache store and return them.
+//
+// Returns an error if it fails to successfully fetch the host/service groups from the API.
+func (client *Client) fetchExtraTagsFor(ctx context.Context, host, service string) (map[string]string, error) {
+	eventName := host
+	if service != "" {
+		eventName = host + "!" + service
+	}
+	if extraTags, ok := client.eventExtraTagsCache.Get(eventName); ok {
```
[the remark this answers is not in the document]
yhabteab: Because that cache is storing *event* extra tags and not object.

internal/incident/incident.go
```diff
@@ -173,11 +182,18 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event) error {
 		}
 	case event.TypeAcknowledgementSet:
 		if err := i.processAcknowledgementEvent(ctx, tx, ev); err != nil {
+			if errors.Is(err, errSuperfluousAckEvent) {
+				// That ack error type indicates that the acknowledgement author was already a manager, thus
+				// we can safely ignore that event and return without even committing the DB transaction.
+				return nil
+			}
```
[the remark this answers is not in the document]
yhabteab: This is not something I've introduced with this PR, it was already mapped to `nil`, so I just left it as is.

internal/icinga2/client.go
```diff
@@ -173,36 +224,57 @@ func (client *Client) buildHostServiceEvent(ctx context.Context, result CheckRes
...
+	// Downtime suppresses all kind of events, so always generate a muted event if it's in downtime.
+	if sc.DowntimeDepth > 0 {
+		ev.SetMute(true, "Checkable is in downtime")
+	} else if sc.IsFlapping {
+		// If the Checkable is not in downtime but in flapping state, all other states/events become irrelevant as well.
+		ev.SetMute(true, "Checkable is in flapping state")
+	} else if sc.Acknowledgement {
+		// If the Checkable is acked, we've to always generate a muted event as well.
+		ev.SetMute(true, "Checkable problem state is acknowledged")
+	}
```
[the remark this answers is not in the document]
yhabteab: Are you referring to just the `IsFlapping` part or all of them? Yes, currently `IsFlapping` is set for the catch-up phase only, but the others are not, and I want to have them centralised here as `buildHostServiceEvent()` is used for all state change events, so I don't quite get what you're asking for.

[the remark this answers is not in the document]
julianbrost: I'm referring to all of them. These seem to be redundant with the handling of downtime start/flapping start/ack set events, so I was wondering if this is actually just for the catch-up phase where no special handling for these is done besides for acknowledgements.

If this was necessary (just) for the catch-up phase, I'd wonder about two more things: why is there no unmute happening otherwise and it looks like you might lose those mute events if no state change happened.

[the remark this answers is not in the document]
yhabteab: > so I was wondering if this is actually just for the catch-up phase where no special handling for these is done besides for acknowledgements

As I said, this is not exclusive to the catch-up phase, but applies to all state change events and ensures that no notifications are triggered if, for example, we somehow miss the actual downtime start event. In this case, it should not trigger any notifications as it is clearly handled by the user. Yes, I know we don't have access to the `is_flapping` information with every state change event, but it's still better than nothing.

julianbrost (changes requested):

internal/icinga2/client.go
```diff
@@ -213,30 +274,77 @@ func (client *Client) buildDowntimeEvent(ctx context.Context, d Downtime, startE
...
+			flapping.CurrentFlapping, flapping.ThresholdLow,
+		)
+
+		if queryResult == nil {
+			// If queryResult is nil, then the Checkable is still muted due to other reasons, so no need to unmute it.
+			ev.Message = reason
+		} else {
+			ev.Message = queryResult.Attrs.LastCheckResult.Output
+			ev.SetMute(false, reason)
+		}
```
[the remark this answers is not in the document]
yhabteab: > because I missed that you're using the response for more than checking if it's still muted.

That's exactly what I'm doing here too, and to be honest, the previous version was perfect enough for me, I just didn't want to waste time discussing why your suggested solution wasn't better than it was in anyway. We shouldn't over engineer things too much just to *deduplicate* a simple if statement, which by the way wasn't duplicated code at all. These new functions are just useless and overcomplicate things unnecessarily.

internal/icinga2/client_api.go
```diff
@@ -256,46 +271,74 @@ func (client *Client) checkMissedChanges(ctx context.Context, objType string, ca
...
-			ackComment.Author, ackComment.Text, false)
-		if err != nil {
-			return fmt.Errorf("failed to construct Event from Acknowledgement response, %w", err)
+			fakeEv.Type = event.TypeMute
+			if attrs.DowntimeDepth > 0 {
+				fakeEv.SetMute(true, "Checkable is in downtime, but we missed the Icinga 2 DowntimeStart event")
+			} else {
+				fakeEv.SetMute(true, "Checkable is flapping, but we missed the Icinga 2 FlappingStart event")
+			}
+		} else if attrs.Acknowledgement > AcknowledgementNone {
```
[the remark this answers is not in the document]
yhabteab: > Also, I'd just use `!=` for the comparison. I don't see a need why we should have to rely on the exact underlying integer values.

I have absolutely no idea what you're talking about, since when does the `>` check rely on the exact integer values? It's clearly documented what values [`Acknowledgement`](https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#event-stream-type-acknowledgementset) can have, so I don't really know what to say.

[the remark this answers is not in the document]
julianbrost: From the functionality, it's exactly the same, but:

- `attrs.Acknowledgement != AcknowledgementNone` simply says "anything except no acknowledgement"
- `attrs.Acknowledgement > AcknowledgementNone` raises the question what's the value of `AcknowledgementNone` and how does it compare to other possible values, which you have to look up and in the end conclude "anything except no acknowledgement"

julianbrost:
For the record: the recent changes pushed here should address something I discussed with @yhabteab yesterday: so far, the handling events like a downtime start without an incident was that if it was received from the Icinga 2 event stream, it would mute the object, which is also persisted in the database, but the event itself wouldn't make it to the database and result in an error. When catching up missed events, it wouldn't even mute the object if there was no incident.

internal/object/objects.go
```diff
@@ -0,0 +1,97 @@
...
+
+	delete(cache, id.String())
+}
+
+// RestoreMutedObjects restores all muted objects and their extra / ID tags from the database.
+// Note, this function only retrieves muted objects without non-recovered incident.
+//
+// Returns an error on any database failure and panics when trying to cache an object that's already in the cache store.
+func RestoreMutedObjects(ctx context.Context, db *database.DB) error {
+	clause := `WHERE mute_reason IS NOT NULL AND NOT EXISTS((SELECT 1 FROM incident WHERE object_id = object.id AND recovered_at IS NULL))`
```
julianbrost, line 31: Isn't this almost the same function as `RestoreObjects()`, just with a different `where` clause?

Also note that the code that you moved from that function into `restoreObjectsTags()` implicitly relies on this following:

https://github.com/Icinga/icinga-notifications/blob/fb1264c826742d4d98ff09f37044a3be81af6412/internal/incident/incidents.go#L63

Your code will likely run into query size limits with many muted objects there.

I've been thinking about this for a while now and didn't come up with a great solution so far. As `RestoreObjects()` uses `utils.ForEachRow()` internally and relies on the external chunking, it's not extended easily. One option might be to just select the IDs here with the `where` clause from above, chunk them and then simply feed them into the existing `RestoreObjects()` function.

yhabteab: > One option might be to just select the IDs here with the where clause from above, chunk them and then simply feed them into the existing RestoreObjects() function.

Wouldn't that be a bit tedious to just SELECT the object IDs so that we can use them for a where clause and retrieve the same set of info? If in doubt, I would start two goroutines, similar to how it's done for incidents!

julianbrost: As I said, I'm not completely happy with that suggestion, but at least it would allow reusing the existing code instead of duplicating `RestoreObjects()` as `RestoreMutedObjects()` where you than have inlined `utils.ForEachRow()` so that you could adapt the `where` clause and then moved the remaining common code into another function.

Other idea: split `utils.ForEachRow()` in the middle so that it becomes two function, one for generating a `SELECT ... WHERE ... IN (...)` query and another one for executing a query and iterating over the result scanned into instances of a specific type. Other existing users could then get a function that combines the two like it currently does. For the restore object functions, this would allow building a common function like `RestoreObjectsFromQuery()` that could then be used with two different queries.

yhabteab: >  For the restore object functions, this would allow building a common function like RestoreObjectsFromQuery() that could then be used with two different queries.

I do not see how this is supposed to solve the specific problem you mentioned above "Your code will likely run into query size limits with many muted objects there." This is just another way of deduplicating existing codes, but does not address the shortcomings of the current state of this PR.

yhabteab: > If in doubt, I would start two goroutines, similar to how it's done for incidents!

I've implemented this option!

julianbrost: It doesn't. These functions would still have to handle this properly.

There's a second thing here that I think the code structure can be improved: for one, there's some duplication that might not be necessary with more helpful utility functions (hence the suggestion so split `utils.ForEachRow()`). Also the `restoreObjectsTags()` function feels a bit strange to me, like to me it looks like the only reason it exists in the way it does is because you had that common code in two places. Why does it have parameters `objects map[string]*Object` and `ids []types.Binary` where `ids` should contain the keys from `objects`? Probably because that's what existed as local variables. Why does it fetch the tags and then insert the object into the cache? Probably because that was the common code in two functions. I don't think you would come up with that structure when asking questions more like "is this a useful function?" (in the sense that you can describe the purpose of the function not just by "does the steps it does" but also "this is a useful task as a whole").

yhabteab: > It doesn't. These functions would still have to handle this properly.

Which functions are you talking about? Also, what do you expect from *handle this correctly*? Should the functions simply fail if they exceed the query limit? If you're suggesting to rely entirely on `utils.ForEachRow` in both cases regardless if it's split or not, how should it *correctly* handle query limits.

yhabteab: >  Why does it have parameters `objects map[string]*Object` and `ids []types.Binary` where `ids` should contain the keys from `objects`? Probably because that's what existed as local variables.

Existed as local variables? That's because the objects are only added to the global cache once they have been fully restored and thus are passed on to functions where they are needed.

> Why does it fetch the tags and then insert the object into the cache? Probably because that was the common code in two functions.

No. But this is where the objects get fully restored. I don't know what exactly you want to achieve with this, but for me this is just a simple (internal) helper function that does what it is supposed to do.

julianbrost: The current version of your `RestoreMutedObjects()` function can easily be extended into one that can be used to restore objects by ID as well. Have a look at the following code. `restoreObjectsFromQuery()` is basically that that function with `query` and `args` exposed as function arguments. The other two functions then call it with different queries.

For the `RestoreObjectsById()` case - assuming it's never called with more than `MaxPlaceholdersPerStatement` IDs as it's currently the case - the bulker won't do much, it will just generate a single chunk, but having it doesn't hurt and allows using the very same code.

```go
// RestoreMutedObjects restores all muted objects and their extra / ID tags from the database.
// Note, this function only retrieves muted objects without non-recovered incident.
//
// Returns an error on any database failure and panics when trying to cache an object that's already in the cache store.
func RestoreMutedObjects(ctx context.Context, db *database.DB) error {
	query := db.BuildSelectStmt(new(Object), new(Object)) + " WHERE mute_reason IS NOT NULL " +
		"AND NOT EXISTS((SELECT 1 FROM incident WHERE object_id = object.id AND recovered_at IS NULL))"
	return restoreObjectsFromQuery(ctx, db, query)
}

// RestoreObjectsById restores all objects and their (extra)tags matching the given IDs from the database.
// Returns error on any database failures and panics when trying to cache an object that's already in the cache store.
func RestoreObjectsById(ctx context.Context, db *database.DB, ids []types.Binary) error {
	var obj *Object
	query, args, err := sqlx.In(db.BuildSelectStmt(obj, obj)+" WHERE id IN (?)", ids)
	if err != nil {
		return err
	}

	return restoreObjectsFromQuery(ctx, db, query, args...)
}

// restoreObjectsFromQuery takes a query that returns rows of the object table, executes it and loads the returned
// objects into the local cache.
//
// Returns an error on any database failure and panics when trying to cache an object that's already in the cache store.
func restoreObjectsFromQuery(ctx context.Context, db *database.DB, query string, args ...any) error {
	objects := make(chan *Object)
	g, ctx := errgroup.WithContext(ctx)
	g.Go(func() error {
		defer close(objects)

		err := utils.ExecAndApply[Object](ctx, db, query, args, func(o *Object) {
			o.db = db
			o.Tags = map[string]string{}
			o.ExtraTags = map[string]string{}

			select {
			case objects <- o:
			case <-ctx.Done(): // The ctx error is going to be handled by utils.ExecAndApply.
			}
		})

		return errors.Wrap(err, "cannot restore muted objects without an active incident")
	})

	g.Go(func() error {
		bulks := com.Bulk(ctx, objects, db.Options.MaxPlaceholdersPerStatement, com.NeverSplit[*Object])

		for {
			select {
			case <-ctx.Done():
				return ctx.Err()
			case bulk, ok := <-bulks:
				if !ok {
					return nil
				}

				g.Go(func() error {
					var ids []types.Binary
					objectsMap := map[string]*Object{}
					for _, obj := range bulk {
						objectsMap[obj.ID.String()] = obj
						ids = append(ids, obj.ID)
					}

					// Restore object ID tags matching the given object ids
					err := utils.ForEachRow[IdTagRow](ctx, db, "object_id", ids, func(ir *IdTagRow) {
						objectsMap[ir.ObjectId.String()].Tags[ir.Tag] = ir.Value
					})
					if err != nil {
						return errors.Wrap(err, "cannot restore objects ID tags")
					}

					// Restore object extra tags matching the given object ids
					err = utils.ForEachRow[ExtraTagRow](ctx, db, "object_id", ids, func(et *ExtraTagRow) {
						objectsMap[et.ObjectId.String()].ExtraTags[et.Tag] = et.Value
					})
					if err != nil {
						return errors.Wrap(err, "cannot restore objects extra tags")
					}

					addObjectsToCache(objectsMap)

					return nil
				})
			}
		}
	})

	return g.Wait()
}
```

internal/object/objects.go
```diff
@@ -0,0 +1,114 @@
...
+			case <-ctx.Done():
+				return ctx.Err()
+			case bulk, ok := <-bulks:
+				if !ok {
+					return nil
+				}
+
+				g.Go(func() error {
+					var ids []types.Binary
+					objectsMap := map[string]*Object{}
```
julianbrost, line 68: The final size of these two variables is already known (`len(bulk)`), so they can be initialized to that capacity using `make()`.

internal/object/objects_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	var sourceID int64
+	err := utils.RunInTx(ctx, db, func(tx *sqlx.Tx) error {
+		args := map[string]interface{}{
+			"type":     "notifications",
+			"name":     "Icinga Notifications",
+			"insecure": "n",
+		}
+		// We can't use config.Source here unfortunately due to cyclic import error!
+		id, err := utils.InsertAndFetchId(ctx, tx, `INSERT INTO source (type, name, icinga2_insecure_tls) VALUES (:type, :name, :insecure)`, args)
```
julianbrost, line 29: You shouldn't have to specify `icinga2_insecure_tls` here, that column has a default value.

julianbrost:
@oxzi FYI: I don't expect further large changes to this PR, so feel free to have another look.

oxzi:
Otherwise, I have successfully tested the PR and have not seen anything suspicious in the diff.

julianbrost (commented):
I'm happy with the current state now, I'm just not yet clicking the approve button because I still want you to address [@oxzi's recent review](https://github.com/Icinga/icinga-notifications/pull/200#pullrequestreview-2135115495).

oxzi (approved):

Issue #202, opened 2024-05-27
Schema: name all constraints, not just PKs (before v1.0)


Al2Klimov:
## Advantages

* Schema upgrades affecting constraints can be crafted by just reading the current schema
* Schema upgrades don't have to rely on auto-generated constraint names don't change(?) across RDBMS versions
* Schema upgrades don't have to rely on auto-generated constraint names don't differ(?) in MySQL/MariaDB

## See also

* _The Pitfalls of Using System-Named Constraints in SQL Server_ https://www.madeiradata.com/post/the-pitfalls-of-using-system-named-constraints-in-sql-server
* _Beware of the System Generated Constraint Name_ https://www.databasejournal.com/ms-sql/beware-of-the-system-generated-constraint-name/
* _The Importance of Naming Constraints_ https://alembic.sqlalchemy.org/en/latest/naming.html

Al2Klimov:
* Context: Currently I'm adding a #154 and some constraints explicitly say "constraint may be relaxed in the future" while others are complicated enough CHECKs to assume the same possibility.

Al2Klimov:
Fun fact: some are already named, especially in incident_contact.

Al2Klimov:
Obviously, the auto generated constraint names differ across RDBMS:

```
MariaDB [idb]> show create table rotation_member\G
*************************** 1. row ***************************
       Table: rotation_member
Create Table: CREATE TABLE `rotation_member` (
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `rotation_id` bigint(20) NOT NULL,
  `contact_id` bigint(20) DEFAULT NULL,
  `contactgroup_id` bigint(20) DEFAULT NULL,
  `position` int(11) NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `rotation_id` (`rotation_id`,`position`),
  UNIQUE KEY `rotation_id_2` (`rotation_id`,`contact_id`),
  UNIQUE KEY `rotation_id_3` (`rotation_id`,`contactgroup_id`),
  CONSTRAINT `CONSTRAINT_1` CHECK (if(`contact_id` is null,0,1) + if(`contactgroup_id` is null,0,1) = 1)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
1 row in set (0,001 sec)

MariaDB [idb]>
```

```
Column      |  Type   | Collation | Nullable |                   Default
[...]

PR #203, opened 2024-05-28
Port schema to MySQL 5.7 / MariaDB 10.2

### LINKED ISSUE #154: MySQL Schema?
Hi all,
I know, it's early stage. But could you provide a schema for MySQL / MariaDB ? Would love to test it and give feedback :)

Thanks and cheers,
Marcus

Al2Klimov:
fixes #154

### Required

* [x] https://github.com/Icinga/icinga-notifications-web/issues/194
* [x] https://github.com/Icinga/icinga-notifications-web/issues/195

### Preferred

* [x] #202
* [x] https://github.com/Icinga/icinga-notifications-web/pull/221

### Optional

* [ ] #239
* [ ] https://github.com/Icinga/icinga-notifications-web/issues/196

### DIFF
--- a/.github/workflows/tests_with_database.yml
+++ b/.github/workflows/tests_with_database.yml
@@ -71,7 +71,12 @@ jobs:
 
       - name: Run tests
         timeout-minutes: 10
-        run: go test -v -timeout 5m ./...
+        # By default, multiple packages may be tested in parallel in different
+        # processes. Passing -p 1 reduces this to one process to prevent test
+        # cases in different packages from accessing the same database. Note
+        # that t.Parallel() only affects parallelism within one process, i.e.
+        # within the tests of one package.
+        run: go test -v -timeout 5m -p 1 ./...
 
   postgresql:
     name: PostgreSQL ${{ matrix.version }}
@@ -125,4 +130,9 @@ jobs:
 
       - name: Run tests
         timeout-minutes: 10
-        run: go test -v -timeout 5m ./...
+        # By default, multiple packages may be tested in parallel in different
+        # processes. Passing -p 1 reduces this to one process to prevent test
+        # cases in different packages from accessing the same database. Note
+        # that t.Parallel() only affects parallelism within one process, i.e.
+        # within the tests of one package.
+        run: go test -v -timeout 5m -p 1 ./...
--- a/.github/workflows/tests_with_database.yml
+++ b/.github/workflows/tests_with_database.yml
@@ -7,6 +7,72 @@ on:
   pull_request: {}
 
 jobs:
+  mysql:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        image:
+          - mysql:5.7
+          - mysql:8.0
+          - mysql:8
+          - mysql:latest
+          - mariadb:10.2
+          - mariadb:10.3
+          - mariadb:10.4
+          - mariadb:10.5
+          - mariadb:10.6
+          - mariadb:10
+          - mariadb:11.0
+          - mariadb:11.1
+          - mariadb:11.2
+          - mariadb:latest
+
+    env:
+      NOTIFICATIONS_TESTS_DB_TYPE: mysql
+      NOTIFICATIONS_TESTS_DB: notifications
+      NOTIFICATIONS_TESTS_DB_USER: root
+      NOTIFICATIONS_TESTS_DB_PASSWORD: notifications
+      NOTIFICATIONS_TESTS_DB_HOST: 127.0.0.1
+      NOTIFICATIONS_TESTS_DB_PORT: 3306
+
+    services:
+      mysql:
+        image: ${{ matrix.image }}
+        env:
+          MYSQL_ROOT_PASSWORD: ${{ env.NOTIFICATIONS_TESTS_DB_PASSWORD }}
+          MYSQL_DATABASE: ${{ env.NOTIFICATIONS_TESTS_DB }}
+        # Wait until MySQL becomes ready
+        options: >-
+          --health-cmd "${{ (startsWith(matrix.image, 'mysql:') || startsWith(matrix.image, 'mariadb:10')) && 'mysqladmin ping' || 'healthcheck.sh --connect --innodb_initialized' }}"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 10
+        ports:
+          - 3306:3306
+
+    steps:
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: stable
+
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Importing Schema
+        run: >
+          mysql -h127.0.0.1 -u$NOTIFICATIONS_TESTS_DB_USER -p$NOTIFICATIONS_TESTS_DB_PASSWORD
+          $NOTIFICATIONS_TESTS_DB < ${{ github.workspace }}/schema/mysql/schema.sql
+
+      - name: Download dependencies
+        run: go get -v -t -d ./...
+
+      - name: Run tests
+        timeout-minutes: 10
+        run: go test -v -timeout 5m ./...
+
   postgresql:
     name: PostgreSQL ${{ matrix.version }}
     runs-on: ubuntu-latest
--- a/doc/02-Installation.md
+++ b/doc/02-Installation.md
@@ -21,19 +21,11 @@ or install [from source](02-Installation.md.d/From-Source.md).
 
 ## Setting up the Database
 
-A MySQL (≥5.5), MariaDB (≥10.1), or PostgreSQL (≥9.6) database is required to run Icinga Notifications.
+A MySQL (≥5.7.9), MariaDB (≥10.2.2), or PostgreSQL (≥9.6) database is required to run Icinga Notifications.
 Please follow the steps listed for your target database,
 which guide you through setting up the database and user and importing the schema.
 
 ### Setting up a MySQL or MariaDB Database
-
-If you use a version of MySQL < 5.7 or MariaDB < 10.2, the following server options must be set:
-
-```
-innodb_file_format=barracuda
-innodb_file_per_table=1
-innodb_large_prefix=1
-```
 
 Set up a MySQL database for Icinga Notifications:
 
--- a/internal/event/event.go
+++ b/internal/event/event.go
@@ -68,6 +68,20 @@ func (e *Event) Validate() error {
 		return fmt.Errorf("invalid event: tags must not be empty")
 	}
 
+	for tag := range e.Tags {
+		if len(tag) > 255 {
+			return fmt.Errorf("invalid event: tag %q is too long, at most 255 chars allowed, %d given", tag, len(tag))
+		}
+	}
+
+	for tag := range e.ExtraTags {
+		if len(tag) > 255 {
+			return fmt.Errorf(
+				"invalid event: extra tag %q is too long, at most 255 chars allowed, %d given", tag, len(tag),
+			)
+		}
+	}
+
 	if e.SourceId == 0 {
 		return fmt.Errorf("invalid event: source ID must not be empty")
 	}
--- a/schema/mysql/schema.sql
+++ b/schema/mysql/schema.sql
@@ -1,0 +1,458 @@
+CREATE TABLE available_channel_type (
+    type varchar(255) NOT NULL,
+    name text NOT NULL,
+    version text NOT NULL,
+    author text NOT NULL,
+    config_attrs mediumtext NOT NULL,
+
+    CONSTRAINT pk_available_channel_type PRIMARY KEY (type)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE TABLE channel (
+    id bigint NOT NULL AUTO_INCREMENT,
+    name text NOT NULL COLLATE utf8mb4_unicode_ci,
+    type varchar(255) NOT NULL, -- 'email', 'sms', ...
+    config mediumtext, -- JSON with channel-specific attributes
+    -- for now type determines the implementation, in the future, this will need a reference to a concrete
+    -- implementation to allow multiple implementations of a sms channel for example, probably even user-provided ones
+
+    changed_at bigint NOT NULL,
+    deleted enum('n', 'y') NOT NULL DEFAULT 'n',
+
+    CONSTRAINT pk_channel PRIMARY KEY (id),
+    CONSTRAINT fk_channel_available_channel_type FOREIGN KEY (type) REFERENCES available_channel_type(type)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE INDEX idx_channel_changed_at ON channel(changed_at);
+
+CREATE TABLE contact (
+    id bigint NOT NULL AUTO_INCREMENT,
+    full_name text NOT NULL COLLATE utf8mb4_unicode_ci,
+    username varchar(254) COLLATE utf8mb4_unicode_ci, -- reference to web user
+    default_channel_id bigint NOT NULL,
+
+    changed_at bigint NOT NULL,
+    deleted enum('n', 'y') NOT NULL DEFAULT 'n',
+
+    CONSTRAINT pk_contact PRIMARY KEY (id),
+
+    -- As the username is unique, it must be NULLed for deletion via "deleted = 'y'"
+    CONSTRAINT uk_contact_username UNIQUE (username),
+
+    CONSTRAINT fk_contact_channel FOREIGN KEY (default_channel_id) REFERENCES channel(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE INDEX idx_contact_changed_at ON contact(changed_at);
+
+CREATE TABLE contact_address (
+    id bigint NOT NULL AUTO_INCREMENT,
+    contact_id bigint NOT NULL,
+    type varchar(255) NOT NULL, -- 'phone', 'email', ...
+    address text NOT NULL, -- phone number, email address, ...
+
+    changed_at bigint NOT NULL,
+    deleted enum('n', 'y') NOT NULL DEFAULT 'n',
+
+    CONSTRAINT pk_contact_address PRIMARY KEY (id),
+    CONSTRAINT fk_contact_address_contact FOREIGN KEY (contact_id) REFERENCES contact(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE INDEX idx_contact_address_changed_at ON contact_address(changed_at);
+
+CREATE TABLE contactgroup (
+    id bigint NOT NULL AUTO_INCREMENT,
+    name text NOT NULL COLLATE utf8mb4_unicode_ci,
+
+    changed_at bigint NOT NULL,
+    deleted enum('n', 'y') NOT NULL DEFAULT 'n',
+
+    CONSTRAINT pk_contactgroup PRIMARY KEY (id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE INDEX idx_contactgroup_changed_at ON contactgroup(changed_at);
+
+CREATE TABLE contactgroup_member (
+    contactgroup_id bigint NOT NULL,
+    contact_id bigint NOT NULL,
+
+    changed_at bigint NOT NULL,
+    deleted enum('n', 'y') NOT NULL DEFAULT 'n',
+
+    CONSTRAINT pk_contactgroup_member PRIMARY KEY (contactgroup_id, contact_id),
+    CONSTRAINT fk_contactgroup_member_contactgroup FOREIGN KEY (contactgroup_id) REFERENCES contactgroup(id),
+    CONSTRAINT fk_contactgroup_member_contact FOREIGN KEY (contact_id) REFERENCES contact(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE INDEX idx_contactgroup_member_changed_at ON contactgroup_member(changed_at);
+
+CREATE TABLE schedule (
+    id bigint NOT NULL AUTO_INCREMENT,
+    name text NOT NULL COLLATE utf8mb4_unicode_ci,
+
+    changed_at bigint NOT NULL,
+    deleted enum('n', 'y') NOT NULL DEFAULT 'n',
+
+    CONSTRAINT pk_schedule PRIMARY KEY (id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE INDEX idx_schedule_changed_at ON schedule(changed_at);
+
+CREATE TABLE rotation (
+    id bigint NOT NULL AUTO_INCREMENT,
+    schedule_id bigint NOT NULL,
+    -- the lower the more important, starting at 0, avoids the need to re-index upon addition
+    priority integer,
+    name text NOT NULL,
+    mode enum('24-7', 'partial', 'multi'), -- NOT NULL is enforced via CHECK not to default to '24-7'
+    -- JSON with rotation-specific attributes
+    -- Needed exclusively by Web to simplify editing and visualisation
+    options mediumtext NOT NULL,
+
+    -- A date in the format 'YYYY-MM-DD' when the first handoff should happen.
+    -- It is a string as handoffs are restricted to happen only once per day
+    first_handoff date,
+
+    -- Set to the actual time of the first handoff.
+    -- If this is in the past during creation of the rotation, it is set to the creation time.
+    -- Used by Web to avoid showing shifts that never happened
+    actual_handoff bigint NOT NULL,
+
+    changed_at bigint NOT NULL,
+    deleted enum('n', 'y') NOT NULL DEFAULT 'n',
+
+    CONSTRAINT pk_rotation PRIMARY KEY (id),
+
+    -- Each schedule can only have one rotation with a given priority starting at a given date.
+    -- Columns schedule_id, priority, first_handoff must be NULLed for deletion via "deleted = 'y'".
+    CONSTRAINT uk_rotation_schedule_id_priority_first_handoff UNIQUE (schedule_id, priority, first_handoff),
+    CONSTRAINT ck_rotation_non_deleted_needs_priority_first_handoff CHECK (deleted = 'y' OR priority IS NOT NULL AND first_handoff IS NOT NULL),
+
+    CONSTRAINT ck_rotation_mode_notnull CHECK (mode IS NOT NULL),
+    CONSTRAINT fk_rotation_schedule FOREIGN KEY (schedule_id) REFERENCES schedule(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE INDEX idx_rotation_changed_at ON rotation(changed_at);
+
+CREATE TABLE timeperiod (
+    id bigint NOT NULL AUTO_INCREMENT,
+    owned_by_rotation_id bigint, -- nullable for future standalone timeperiods
+
+    changed_at bigint NOT NULL,
+    deleted enum('n', 'y') NOT NULL DEFAULT 'n',
+
+    CONSTRAINT pk_timeperiod PRIMARY KEY (id),
+    CONSTRAINT fk_timeperiod_rotation FOREIGN KEY (owned_by_rotation_id) REFERENCES rotation(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE INDEX idx_timeperiod_changed_at ON timeperiod(changed_at);
+
+CREATE TABLE rotation_member (
+    id bigint NOT NULL AUTO_INCREMENT,
+    rotation_id bigint NOT NULL,
+    contact_id bigint,
+    contactgroup_id bigint,
+    position integer,
+
+    changed_at bigint NOT NULL,
+    deleted enum('n', 'y') NOT NULL DEFAULT 'n',
+
+    -- Each position in a rotation can only be used once.
+    -- Column position must be NULLed for deletion via "deleted = 'y'".
+    CONSTRAINT uk_rotation_member_rotation_id_position UNIQUE (rotation_id, position),
+
+    -- Two UNIQUE constraints prevent duplicate memberships of the same contact or contactgroup in a single rotation.
+    -- Multiple NULLs are not considered to be duplicates, so rows with a contact_id but no contactgroup_id are
+    -- basically ignored in the UNIQUE constraint over contactgroup_id and vice versa. The CHECK constraint below
+    -- ensures that each row has only non-NULL values in one of these constraints.
+    CONSTRAINT uk_rotation_member_rotation_id_contact_id UNIQUE (rotation_id, contact_id),
+    CONSTRAINT uk_rotation_member_rotation_id_contactgroup_id UNIQUE (rotation_id, contactgroup_id),
+
+    CONSTRAINT ck_rotation_member_either_contact_id_or_contactgroup_id CHECK (if(contact_id IS NULL, 0, 1) + if(contactgroup_id IS NULL, 0, 1) = 1),
+    CONSTRAINT ck_rotation_member_non_deleted_needs_position CHECK (deleted = 'y' OR position IS NOT NULL),
+
+    CONSTRAINT pk_rotation_member PRIMARY KEY (id),
+    CONSTRAINT fk_rotation_member_rotation FOREIGN KEY (rotation_id) REFERENCES rotation(id),
+    CONSTRAINT fk_rotation_member_contact FOREIGN KEY (contact_id) REFERENCES contact(id),
+    CONSTRAINT fk_rotation_member_contactgroup FOREIGN KEY (contactgroup_id) REFERENCES contactgroup(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE INDEX idx_rotation_member_changed_at ON rotation_member(changed_at);
+
+CREATE TABLE timeperiod_entry (
+    id bigint NOT NULL AUTO_INCREMENT,
+    timeperiod_id bigint NOT NULL,
+    rotation_member_id bigint, -- nullable for future standalone timeperiods
+    start_time bigint NOT NULL,
+    end_time bigint NOT NULL,
+    -- Is needed by icinga-notifications-web to prefilter entries, which matches until this time and should be ignored by the daemon.
+    until_time bigint,
+    timezone text NOT NULL, -- e.g. 'Europe/Berlin', relevant for evaluating rrule (DST changes differ between zones)
+    rrule text, -- recurrence rule (RFC5545)
+
+    changed_at bigint NOT NULL,
+    deleted enum('n', 'y') NOT NULL DEFAULT 'n',
+
+    CONSTRAINT pk_timeperiod_entry PRIMARY KEY (id),
+    CONSTRAINT fk_timeperiod_entry_timeperiod FOREIGN KEY (timeperiod_id) REFERENCES timeperiod(id),
+    CONSTRAINT fk_timeperiod_entry_rotation_member FOREIGN KEY (rotation_member_id) REFERENCES rotation_member(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE INDEX idx_timeperiod_entry_changed_at ON timeperiod_entry(changed_at);
+
+CREATE TABLE source (
+    id bigint NOT NULL AUTO_INCREMENT,
+    -- The type "icinga2" is special and requires (at least some of) the icinga2_ prefixed columns.
+    type text NOT NULL,
+    name text NOT NULL COLLATE utf8mb4_unicode_ci,
+    -- will likely need a distinguishing value for multiple sources of the same type in the future, like for example
+    -- the Icinga DB environment ID for Icinga 2 sources
+
+    -- The column listener_password_hash is type-dependent.
+    -- If type is not "icinga2", listener_password_hash is required to limit API access for incoming connections
+    -- to the Listener. The username will be "source-${id}", allowing early verification.
+    listener_password_hash text,
+
+    -- Following columns are for the "icinga2" type.
+    -- At least icinga2_base_url, icinga2_auth_user, and icinga2_auth_pass are required - see CHECK below.
+    icinga2_base_url text,
+    icinga2_auth_user text,
+    icinga2_auth_pass text,
+    -- icinga2_ca_pem specifies a custom CA to be used in the PEM format, if not NULL.
+    icinga2_ca_pem text,
+    -- icinga2_common_name requires Icinga 2's certificate to hold this Common Name if not NULL. This allows using a
+    -- differing Common Name - maybe an Icinga 2 Endpoint object name - from the FQDN within icinga2_base_url.
+    icinga2_common_name text,
+    icinga2_insecure_tls enum('n', 'y') NOT NULL DEFAULT 'n',
+
+    changed_at bigint NOT NULL,
+    deleted enum('n', 'y') NOT NULL DEFAULT 'n',
+
+    -- The hash is a PHP password_hash with PASSWORD_DEFAULT algorithm, defaulting to bcrypt. This check roughly ensures
+    -- that listener_password_hash can only be populated with bcrypt hashes.
+    -- https://icinga.com/docs/icinga-web/latest/doc/20-Advanced-Topics/#manual-user-creation-for-database-authentication-backend
+    CONSTRAINT ck_source_bcrypt_listener_password_hash CHECK (listener_password_hash IS NULL OR listener_password_hash LIKE '$2y$%'),
+    CONSTRAINT ck_source_icinga2_has_config CHECK (type != 'icinga2' OR (icinga2_base_url IS NOT NULL AND icinga2_auth_user IS NOT NULL AND icinga2_auth_pass IS NOT NULL)),
+
+    CONSTRAINT pk_source PRIMARY KEY (id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE INDEX idx_source_changed_at ON source(changed_at);
+
+CREATE TABLE object (
+    id binary(32) NOT NULL, -- SHA256 of identifying tags and the source.id
+    source_id bigint NOT NULL,
+    name text NOT NULL,
+
+    url text,
+    -- mute_reason indicates whether an object is currently muted by its source, and its non-zero value is mapped to true.
+    mute_reason mediumtext,
+
+    CONSTRAINT pk_object PRIMARY KEY (id),
+    CONSTRAINT fk_object_source FOREIGN KEY (source_id) REFERENCES source(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE TABLE object_id_tag (
+    object_id binary(32) NOT NULL,
+    tag varchar(255) NOT NULL,
+    value text NOT NULL,
+
+    CONSTRAINT pk_object_id_tag PRIMARY KEY (object_id, tag),
+    CONSTRAINT fk_object_id_tag_object FOREIGN KEY (object_id) REFERENCES object(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE TABLE object_extra_tag (
+    object_id binary(32) NOT NULL,
+    tag varchar(255) NOT NULL,
+    value text NOT NULL,
+
+    CONSTRAINT pk_object_extra_tag PRIMARY KEY (object_id, tag),
+    CONSTRAINT fk_object_extra_tag_object FOREIGN KEY (object_id) REFERENCES object(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE TABLE event (
+    id bigint NOT NULL AUTO_INCREMENT,
+    time bigint NOT NULL,
+    object_id binary(32) NOT NULL,
+    -- NOT NULL is enforced via CHECK not to default to 'acknowledgement-cleared'
+    type enum('acknowledgement-cleared', 'acknowledgement-set', 'custom', 'downtime-end', 'downtime-removed', 'downtime-start', 'flapping-end', 'flapping-start', 'incident-age', 'mute', 'state', 'unmute'),
+    severity enum('ok', 'debug', 'info', 'notice', 'warning', 'err', 'crit', 'alert', 'emerg'),
+    message mediumtext,
+    username text COLLATE utf8mb4_unicode_ci,
+    mute enum('n', 'y'),
+    mute_reason mediumtext,
+
+    CONSTRAINT pk_event PRIMARY KEY (id),
+    CONSTRAINT ck_event_type_notnull CHECK (type IS NOT NULL),
+    CONSTRAINT fk_event_object FOREIGN KEY (object_id) REFERENCES object(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE TABLE rule (
+    id bigint NOT NULL AUTO_INCREMENT,
+    name text NOT NULL COLLATE utf8mb4_unicode_ci,
+    timeperiod_id bigint,
+    object_filter text,
+
+    changed_at bigint NOT NULL,
+    deleted enum('n', 'y') NOT NULL DEFAULT 'n',
+
+    CONSTRAINT pk_rule PRIMARY KEY (id),
+    CONSTRAINT fk_rule_timeperiod FOREIGN KEY (timeperiod_id) REFERENCES timeperiod(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE INDEX idx_rule_changed_at ON rule(changed_at);
+
+CREATE TABLE rule_escalation (
+    id bigint NOT NULL AUTO_INCREMENT,
+    rule_id bigint NOT NULL,
+    position integer,
+    `condition` text,
+    name text COLLATE utf8mb4_unicode_ci, -- if not set, recipients are used as a fallback for display purposes
+    fallback_for bigint,
+
+    changed_at bigint NOT NULL,
+    deleted enum('n', 'y') NOT NULL DEFAULT 'n',
+
+    CONSTRAINT pk_rule_escalation PRIMARY KEY (id),
+
+    -- Each position in an escalation can only be used once.
+    -- Column position must be NULLed for deletion via "deleted = 'y'"
+    CONSTRAINT uk_rule_escalation_rule_id_position UNIQUE (rule_id, position),
+
+    CONSTRAINT ck_rule_escalation_not_both_condition_and_fallback_for CHECK (NOT (`condition` IS NOT NULL AND fallback_for IS NOT NULL)),
+    CONSTRAINT ck_rule_escalation_non_deleted_needs_position CHECK (deleted = 'y' OR position IS NOT NULL),
+    CONSTRAINT fk_rule_escalation_rule FOREIGN KEY (rule_id) REFERENCES rule(id),
+    CONSTRAINT fk_rule_escalation_rule_escalation FOREIGN KEY (fallback_for) REFERENCES rule_escalation(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE INDEX idx_rule_escalation_changed_at ON rule_escalation(changed_at);
+
+CREATE TABLE rule_escalation_recipient (
+    id bigint NOT NULL AUTO_INCREMENT,
+    rule_escalation_id bigint NOT NULL,
+    contact_id bigint,
+    contactgroup_id bigint,
+    schedule_id bigint,
+    channel_id bigint,
+
+    changed_at bigint NOT NULL,
+    deleted enum('n', 'y') NOT NULL DEFAULT 'n',
+
+    CONSTRAINT pk_rule_escalation_recipient PRIMARY KEY (id),
+    CONSTRAINT ck_rule_escalation_recipient_has_exactly_one_recipient CHECK (if(contact_id IS NULL, 0, 1) + if(contactgroup_id IS NULL, 0, 1) + if(schedule_id IS NULL, 0, 1) = 1),
+    CONSTRAINT fk_rule_escalation_recipient_rule_escalation FOREIGN KEY (rule_escalation_id) REFERENCES rule_escalation(id),
+    CONSTRAINT fk_rule_escalation_recipient_contact FOREIGN KEY (contact_id) REFERENCES contact(id),
+    CONSTRAINT fk_rule_escalation_recipient_contactgroup FOREIGN KEY (contactgroup_id) REFERENCES contactgroup(id),
+    CONSTRAINT fk_rule_escalation_recipient_schedule FOREIGN KEY (schedule_id) REFERENCES schedule(id),
+    CONSTRAINT fk_rule_escalation_recipient_channel FOREIGN KEY (channel_id) REFERENCES channel(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE INDEX idx_rule_escalation_recipient_changed_at ON rule_escalation_recipient(changed_at);
+
+CREATE TABLE incident (
+    id bigint NOT NULL AUTO_INCREMENT,
+    object_id binary(32) NOT NULL,
+    started_at bigint NOT NULL,
+    recovered_at bigint,
+    -- NOT NULL is enforced via CHECK not to default to 'ok'
+    severity enum('ok', 'debug', 'info', 'notice', 'warning', 'err', 'crit', 'alert', 'emerg'),
+
+    CONSTRAINT pk_incident PRIMARY KEY (id),
+    CONSTRAINT ck_incident_severity_notnull CHECK (severity IS NOT NULL),
+    CONSTRAINT fk_incident_object FOREIGN KEY (object_id) REFERENCES object(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE TABLE incident_event (
+    incident_id bigint NOT NULL,
+    event_id bigint NOT NULL,
+
+    CONSTRAINT pk_incident_event PRIMARY KEY (incident_id, event_id),
+    CONSTRAINT fk_incident_event_incident FOREIGN KEY (incident_id) REFERENCES incident(id),
+    CONSTRAINT fk_incident_event_event FOREIGN KEY (event_id) REFERENCES event(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE TABLE incident_contact (
+    incident_id bigint NOT NULL,
+    contact_id bigint,
+    contactgroup_id bigint,
+    schedule_id bigint,
+    role enum('recipient', 'subscriber', 'manager'), -- NOT NULL is enforced via CHECK not to default to 'recipient'
+
+    CONSTRAINT uk_incident_contact_incident_id_contact_id UNIQUE (incident_id, contact_id),
+    CONSTRAINT uk_incident_contact_incident_id_contactgroup_id UNIQUE (incident_id, contactgroup_id),
+    CONSTRAINT uk_incident_contact_incident_id_schedule_id UNIQUE (incident_id, schedule_id),
+
+    CONSTRAINT ck_incident_contact_has_exactly_one_recipient CHECK (if(contact_id IS NULL, 0, 1) + if(contactgroup_id IS NULL, 0, 1) + if(schedule_id IS NULL, 0, 1) = 1),
+    CONSTRAINT ck_incident_contact_role_notnull CHECK (role IS NOT NULL),
+    CONSTRAINT fk_incident_contact_incident FOREIGN KEY (incident_id) REFERENCES incident(id),
+    CONSTRAINT fk_incident_contact_contact FOREIGN KEY (contact_id) REFERENCES contact(id),
+    CONSTRAINT fk_incident_contact_contactgroup FOREIGN KEY (contactgroup_id) REFERENCES contactgroup(id),
+    CONSTRAINT fk_incident_contact_schedule FOREIGN KEY (schedule_id) REFERENCES schedule(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE TABLE incident_rule (
+    incident_id bigint NOT NULL,
+    rule_id bigint NOT NULL,
+
+    CONSTRAINT pk_incident_rule PRIMARY KEY (incident_id, rule_id),
+    CONSTRAINT fk_incident_rule_incident FOREIGN KEY (incident_id) REFERENCES incident(id),
+    CONSTRAINT fk_incident_rule_rule FOREIGN KEY (rule_id) REFERENCES rule(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE TABLE incident_rule_escalation_state (
+    incident_id bigint NOT NULL,
+    rule_escalation_id bigint NOT NULL,
+    triggered_at bigint NOT NULL,
+
+    CONSTRAINT pk_incident_rule_escalation_state PRIMARY KEY (incident_id, rule_escalation_id),
+    CONSTRAINT fk_incident_rule_escalation_state_incident FOREIGN KEY (incident_id) REFERENCES incident(id),
+    CONSTRAINT fk_incident_rule_escalation_state_rule_escalation FOREIGN KEY (rule_escalation_id) REFERENCES rule_escalation(id)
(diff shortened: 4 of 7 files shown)


schema/mysql/schema.sql
```diff
@@ -0,0 +1,285 @@
...
+    rule_escalation_id bigint REFERENCES rule_escalation(id),
+    event_id bigint REFERENCES event(id),
+    contact_id bigint REFERENCES contact(id),
+    contactgroup_id bigint REFERENCES contactgroup(id),
+    schedule_id bigint REFERENCES schedule(id),
+    rule_id bigint REFERENCES rule(id),
+    channel_id bigint REFERENCES channel(id),
+    time bigint NOT NULL,
+    message text,
+    -- Order to be honored for events with identical microsecond timestamps.
```
julianbrost, line 272: Timestamps are in milliseconds.

julianbrost (changes requested):
Did you spin this up for a test without any other compatibility issues?

Al2Klimov:
Good catch! I completely forgot the GHA!

schema/pgsql/upgrades/025.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE incident ALTER COLUMN severity SET DEFAULT 'ok';
+ALTER TABLE incident_contact ALTER COLUMN role SET DEFAULT 'recipient';
+ALTER TABLE incident_history ALTER COLUMN type SET DEFAULT 'opened';
+
+ALTER TABLE available_channel_type ALTER COLUMN type TYPE varchar(255);
+ALTER TABLE channel ALTER COLUMN type TYPE varchar(255);
+ALTER TABLE contact ALTER COLUMN username TYPE varchar(254);
+ALTER TABLE contact_address ALTER COLUMN type TYPE varchar(255);
+ALTER TABLE object_id_tag ALTER COLUMN tag TYPE varchar(255);
+ALTER TABLE object_extra_tag ALTER COLUMN tag TYPE varchar(255);
```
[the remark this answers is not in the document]
Al2Klimov: > How did you choose the allowed length, in particular, why is one column shorter than the rest?

Icinga Web user names are <=254. The others... 🤷‍♂️

> Also, when adding constraints to the allowed values, these have to be added to the corresponding input validation functions as well.

Honestly speaking, I'm new here and have no idea what you're talking about. Or do you mean the Web part which writes to the DB via forms?

schema/mysql/schema.sql
```diff
@@ -0,0 +1,296 @@
...
+
+CREATE INDEX idx_incident_history_time_type ON incident_history(time, type) COMMENT 'Incident History ordered by time/type';
+
+CREATE TABLE browser_session (
+    php_session_id varchar(256) NOT NULL,
+    username varchar(254) NOT NULL,
+    user_agent varchar(255) NOT NULL,
+    authenticated_at bigint NOT NULL,
+
+    CONSTRAINT pk_browser_session PRIMARY KEY (php_session_id, username, user_agent)
```
Al2Klimov, line 293: @ncosta-ic Don't say you REALLY need such a combined PK?

schema/pgsql/schema.sql
```diff
@@ -322,9 +320,9 @@ COMMENT ON INDEX idx_incident_history_time_type IS 'Incident History ordered by
 
 CREATE TABLE browser_session (
     php_session_id varchar(256) NOT NULL,
-    username citext NOT NULL,
-    user_agent text NOT NULL,
-    authenticated_at bigint NOT NULL DEFAULT (EXTRACT(EPOCH FROM CURRENT_TIMESTAMP) * 1000),
+    username varchar(254) NOT NULL,
+    user_agent varchar(255) NOT NULL,
+    authenticated_at bigint NOT NULL,
```
Al2Klimov, line 325: @ncosta-ic This breaks the Web part btw.

schema/mysql/schema.sql
ncosta-ic: Yes, I do. I initially used a small [jooat hash](https://en.wikipedia.org/wiki/Jenkins_hash_function#one_at_a_time) for the calculated device id (see https://github.com/Icinga/icinga-notifications/pull/136#discussion_r1407742077), but other contributors wanted me to change it to include the full useragent string (which is truncated to the [first 4096 chars by the PHP daemon](https://github.com/Icinga/icinga-notifications-web/blob/9f9517701ff8b83dd98387a8a903f4b99aee9bb0/library/Notifications/ProvidedHook/SessionStorage.php#L42)).

schema/pgsql/schema.sql
ncosta-ic: @Al2Klimov I do not quite get why you would even change those fields as they are solely used by the PHP Daemon and not touched by the Go Daemon at all.
I know that you're trying to port it over to MariaDB but you need to find appropriate equivalents. I opted for using `citext` for the username field as it was mentioned by Julian in https://github.com/Icinga/icinga-notifications/pull/136#discussion_r1409326979.

As for the `user_agent` field, you would need to change it to a `varchar(4096)` and not only to a `varchar(255)`.
I am personally still more convinced that a device hash would be more appropriate, but I don't make the rules.

Al2Klimov: I meant Web will have to supply authenticated_at by itself.

ncosta-ic: Right, that's why there's the default value. And it would still break when a browser supplies a user_agent  longer than 255 characters as the php daemon allows up to 4096 chars.

Al2Klimov: > As for the `user_agent` field, you would need to change it to a `varchar(4096)` and not only to a `varchar(255)`.

I'm sorry, Noe. I'm afraid I can't do that:

`ERROR 1071 (42000): Specified key was too long; max key length is 3072 bytes`

ncosta-ic: > > As for the `user_agent` field, you would need to change it to a `varchar(4096)` and not only to a `varchar(255)`.
> 
> I'm sorry, Noe. I'm afraid I can't do that:
> 
> `ERROR 1071 (42000): Specified key was too long; max key length is 3072 bytes`

And that's one of the reasons I initially went for a `jooat` hash, but that approach was turned down by others :)

schema/mysql/schema.sql
nilmerg: How much of the user agent is stored isn't relevant here to me. It's rather the fact that it's part of the constraint. If the session id is part of it, the user agent is redundant. If we really add it, it would mean that session hijacking is supported.

ncosta-ic: @nilmerg Valid point. I think we included the user agent to narrow down the possibility of a duplicate session identifier, as we are not able to detect a session cleanup peformed by PHP's session routine reliably.
Thus one identifier could still exist in our database for up to 24 hours after it got cleaned up by PHP.
However, we later introduced an additional check on the session creation, which [cleans up an old entry in the database](https://github.com/Icinga/icinga-notifications-web/blob/a6063d1959435398b08b46acdb586cb95be75857/library/Notifications/ProvidedHook/SessionStorage.php#L47-L68) if it happens to have the same php session id as the current connection.

I think we might be able to completely get rid of the user agent in the constraint.

nilmerg: Then let's remove it.

schema/pgsql/upgrades/026.sql
```diff
@@ -8,4 +8,5 @@ ALTER TABLE contact ALTER COLUMN username TYPE varchar(254);
 ALTER TABLE contact_address ALTER COLUMN type TYPE varchar(255);
 ALTER TABLE object_id_tag ALTER COLUMN tag TYPE varchar(255);
 ALTER TABLE object_extra_tag ALTER COLUMN tag TYPE varchar(255);
+ALTER TABLE browser_session ALTER COLUMN username TYPE varchar(254);
```
Al2Klimov, line 11: FYI. 4e39e8b28402928d6160df1e7b9bd9bfa848d94d:

> Schema: make browser_session.username case-sensitive to limit its length
>
> This is equivalent to Icinga Web 2 where the users are case-sensitive
> in Postgres, but not in MySQL. Also the length limitation enables creation
> of an equivalent schema in MySQL. MySQL requires its keys to have a length.

(@nilmerg NOBODY will mix different RDBMS types in Icinga context.)

nilmerg:
> * [ ]  [Schema: make columns w/ IW2 user names case-insensitive(?) icinga-notifications-web#196](https://github.com/Icinga/icinga-notifications-web/issues/196)

Then, why is the `username` column of `contact` not case insensitive in neither schema yet?

nilmerg:
Then open a bug, referencing the Web issue. Don't reference it here, if you don't plan to change it here.

Al2Klimov:
Whether I change something here depends on whether this PR or https://github.com/Icinga/icinga-notifications-web/issues/196 is completed first.

nilmerg:
That's fine. But then create a separate issue! It doesn't magically change otherwise. This PR is about the MySQL port, not about fixes in the Postgres schema.

schema/pgsql/upgrades/030.sql
```diff
@@ -0,0 +1,15 @@
...
+ALTER TABLE available_channel_type ALTER COLUMN type TYPE varchar(255);
+ALTER TABLE channel ALTER COLUMN type TYPE varchar(255);
+ALTER TABLE contact ADD CHECK (length(username) <= 254);
+ALTER TABLE contact_address ALTER COLUMN type TYPE varchar(255);
+ALTER TABLE object_id_tag ALTER COLUMN tag TYPE varchar(255);
+ALTER TABLE object_extra_tag ALTER COLUMN tag TYPE varchar(255);
+ALTER TABLE browser_session ADD CHECK (length(username) <= 254);
+
+DROP INDEX idx_browser_session_username_agent;
+CREATE INDEX idx_browser_session_username ON browser_session (username);
```
Al2Klimov, line 15: I'm afraid browser_session.user_agent is still too large, so I had to shorten the index. @nilmerg @ncosta-ic

ncosta-ic: Well you can't just drop our index on PostgreSQL's side to make it portable for MariaDB. You would need to find a proper compromise.
Now, here are the limits of both database implementations:
- PostgreSQL has technically no limit on its index sizes, however, the standard access method using btree's caps out at a maximum length of 2704 bytes[^1].
- MariaDB has other limitations caused by the underlying storage engine. Let's assume we use InnoDB, as that's the default for MariaDB / MySQL databases. The documentation about indexes states the following[^2]:
> A primary key is **_unique_** and can never be null.
>...
> In [InnoDB](https://mariadb.com/kb/en/innodb/) tables, **_all indexes contain the primary key as a suffix_**. Thus, when using this storage engine, keeping the primary key as small as possible is particularly important. If a primary key does not exist and there are no UNIQUE indexes, InnoDB creates a 6-bytes clustered index which is invisible to the user.

And if we check the information about unique indexes, it states this[^3]:
> ...
> MariaDB starting with [10.5](https://mariadb.com/kb/en/what-is-mariadb-105/)
> Unique, if index type is not specified, **_is normally a BTREE index_** that can also be used by the optimizer to find rows. **_If the key is longer than the max key length for the used storage engine and the storage engine supports long unique index, a HASH key will be created._** This enables MariaDB to enforce uniqueness for any type or number of columns.

Considering the fact that the only reason for a combined index of `username` and `user_agent` would be to find exact matches, a hash index would be more than sufficient. However, that one is apparently not supported with the InnoDB storage engine[^4].
Thus, I would suggest we change the logic again and just store a hash value (Jenkins One-At-A-Time[^5]) with a very small collision rate of the supplied user-agent, like we did in the first iteration of our php daemon code.

I would need the approval of @julianbrost or @nilmerg though.

[^1]: https://dba.stackexchange.com/a/69164
[^2]: https://mariadb.com/kb/en/getting-started-with-indexes/#primary-key
[^3]: https://mariadb.com/kb/en/getting-started-with-indexes/#unique-index
[^4]: https://mariadb.com/kb/en/storage-engine-index-types/#content
[^5]: https://en.wikipedia.org/wiki/Jenkins_hash_function#one_at_a_time , http://burtleburtle.net/bob/hash/doobs.html#statistics

Al2Klimov: <sup>I know, but he can.</sup>

Jokes aside, what if browser_session.user_agent used a single-byte collation (ASCII) and was slightly < 3KB large?

julianbrost: Have you considered [using a prefix of the column for the index](https://dev.mysql.com/doc/refman/5.7/en/create-index.html#create-index-column-prefixes)?

In general, how much need is there for such an index? How many sessions may reasonably exist per user? Unless the table is never cleaned up and accumulates rows over the years, iterating over something like 10 sessions of a user shouldn't hurt.

> Considering the fact that the only reason for a combined index of `username` and `user_agent` would be to find exact matches, a hash index would be more than sufficient.

What will this index be used for in the end? Like I'd have expected this to be used for some "my sessions" overview where you'd do a `select ... where username = '...' order by user_agent`. Shouldn't the primary way of selecting a specific session be the session ID anyways?

Al2Klimov: > Have you considered [using a prefix of the column for the index](https://dev.mysql.com/doc/refman/5.7/en/create-index.html#create-index-column-prefixes)?

Good idea! Especially this would enable un-limiting the UA length.

Our current 4k is nice, but there is no limit actually: https://stackoverflow.com/a/77072435

Header limits:

* Apache: 8k https://lists.apache.org/thread/vq8z74zl0tsb887n9t7y8f3fo7kp3klx
* nginx 8k: https://nginx.org/en/docs/http/ngx_http_core_module.html#:~:text=A%20request%20header%20field%20cannot,is%20equal%20to%208K%20bytes.

This means I could send you a 7.9k UA.

> In general, how much need is there for such an index? How many sessions may reasonably exist per user? Unless the table is never cleaned up and accumulates rows over the years, iterating over something like 10 sessions of a user shouldn't hurt.

Exactly! Even if MySQL is not smart enough to use the 1-column index when you filter/sort by 2 columns, one can filter/sort by 1 column and do the rest in userspace.

> > Considering the fact that the only reason for a combined index of `username` and `user_agent` would be to find exact matches, a hash index would be more than sufficient.

Apropos hashing, Eric would just store the indexed UA as hash...

> What will this index be used for in the end? Like I'd have expected this to be used for some "my sessions" overview where you'd do a `select ... where username = '...' order by user_agent`. Shouldn't the primary way of selecting a specific session be the session ID anyways?

... but I'd (at least additionally) include the unlimited(!) UA for overview purposes.

ncosta-ic: > In general, how much need is there for such an index? How many sessions may reasonably exist per user? Unless the table is never cleaned up and accumulates rows over the years, iterating over something like 10 sessions of a user shouldn't hurt.
> 
> What will this index be used for in the end? Like I'd have expected this to be used for some "my sessions" overview where you'd do a `select ... where username = '...' order by user_agent`. Shouldn't the primary way of selecting a specific session be the session ID anyways?

Well, to be fair, the only reason that we included the UA to begin with, is to be able to differentiate between multiple valid user sessions. User `A` could be logged in on multiple devices at the same time and should thus receive the desktop notifications on all of them.

Now one might think that the unique PHP session identifiers would be sufficient for this, but there's one problem that we can't really solve in a neat way: We don't know when PHP cleans up a session, as this is dependent on the PHP configuration and on how frequent requests towards the web server are being made. And, one could just close the tab or browser instead of using the logout function, which completely bypasses Icinga's session handling (that we are in control of).

Because of this, we included a cleanup job into our PHP daemon, which cleans up sessions in the database that have been authenticated for more than 24 hours.
This, however, means, that PHP could technically assign the same identifier to two different sessions within a timespan of 24 hours.

We trust PHP's session uniqueness in behalf of its implementations and clean up a (potentionally) existing session in our table, if there's a new, successful login that uses the same identifier. The reason being, that this could only ever happen, if we would still have a zombie session saved to our database, that PHP already cleaned up.

To keep our zombie session count low, the PHP daemon checks for [a potential zombie session on a successful login](https://github.com/Icinga/icinga-notifications-web/blob/ba627b479f02b838fa83151cb8b787dd5af36f5b/library/Notifications/ProvidedHook/SessionStorage.php#L71-L86). This check provides the user-agent, since we want to preserve other valid sessions of that user that are tied to other devices. If we did not include the user-agent, this check would clean up all other sessions of that user, even the ones which are still in use.

I hope that gives you a better idea about the reasoning behind this index and why it's needed. I am still sure, that a simple hash over the complete user-agent would be more than sufficient for this specific demand, but @Al2Klimov has a valid point. For the time being, the desktop notifications feature is the only part of `icinga-notifications-web` that makes use of it.
As for the future, we do plan to merge the PHP daemon with additional logic to the Icinga2 (Web) core. This could then create additional demands, which would want to differentiate between different aspects of a connection (like the browser, operating system, you name it...). If we keep this in mind, there is a reason to store the full user-agent into the database. However, it would not need to be indexed.

ncosta-ic: > If it does, it would probably be a security issue in PHP. That would likely imply that you can guess session IDs because they are not random enough.

You're right. I glanced over the current PHP implementation and it seems like they properly solved it for newer systems and installations:
[session.c](https://github.com/php/php-src/blob/9d3907fd8564b47e8d7e9ebf68161ac12ceb0455/ext/session/session.c#L341-L346) > [csprng.c](https://github.com/php/php-src/blob/9d3907fd8564b47e8d7e9ebf68161ac12ceb0455/ext/random/csprng.c#L69)
For systems without proper csprng support or on a PHP version under 7.1[^1], they still fallback to the old sha1 method:
[php_random.h](https://github.com/php/php-src/blob/9d3907fd8564b47e8d7e9ebf68161ac12ceb0455/ext/random/php_random.h#L45-L47) > [random.c](https://github.com/php/php-src/blob/9d3907fd8564b47e8d7e9ebf68161ac12ceb0455/ext/random/random.c#L685-L687) > [random.c](https://github.com/php/php-src/blob/9d3907fd8564b47e8d7e9ebf68161ac12ceb0455/ext/random/random.c#L613-L630)

> That looks like something where you try to clean up some sessions even before their expiry after 24 hours, i.e. other sessions of the same user that share the same user agent string. I feel like I've said this before, but I wouldn't make the assumption that the user agent string uniquely identifies a user's device. I currently have three devices in this room that will report the very same Chrom(e|ium) on Linux user agent.
> 
> To be honest, that sounds like an optimization that should not be necessary and might even cause issues in some scenarios. If a user signs in five times a day on the same device, would it cause any trouble if those entries in the table would live for a day?

No, not really. I was just trying to follow the session security guidelines by PHP[^2], which state:
> Developers should keep track of all active sessions for every user.
> ...
> Various ways to implement this exist. One possible implementation is setting up a database that keeps track of the required data and stores any relevant information. Since session data is GCed, developers must take care of the GCed data to maintain the active session database consistency. 
Meaning if we don't clean them up (when they have been garbage collected by PHP itself), a session id could still be used to connect to the PHP daemon. This could leak check data to an unwanted user.

This lead me to the idea of manually cleaning up (confirmed) zombie sessions.

> I mean if you think it's necessary, you can have it. To me, solutions that don't require external workarounds for the database sound preferable. And truncating the column for the index sounds like a fair compromise where you still get the help of the index as when the first N characters of the user agent differ (which they typically should if N is chosen sufficiently large so that it extends beyond the useless `Mozilla/5.0` prefix that all browsers send).

I'm with you on that one. Albeit, I don't really think that calculating a hash with the native PHP functions would be considered a workaround, as the only thing it's supposed to do, is, to give information about the equality of two user agents. And yes, just reading a user-agent obviously won't be a 100% solution to detect different devices. But I really cannot condone *implementing fingerprinting* - even for such a feature. That's a no, no...

[^1]: https://www.php.net/manual/en/features.session.security.management.php#features.session.security.management.non-adaptive-session comment about `session_create_id()` and its note.
[^2]: https://www.php.net/manual/en/features.session.security.management.php#features.session.security.management.active-sessions

schema/mysql/schema.sql
```diff
@@ -0,0 +1,321 @@
...
+    php_session_id varchar(256) NOT NULL,
+    username varchar(254) NOT NULL COLLATE utf8mb4_unicode_ci,
+    user_agent varchar(4096) NOT NULL,
+    authenticated_at bigint NOT NULL,
+
+    CONSTRAINT pk_browser_session PRIMARY KEY (php_session_id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin ROW_FORMAT=DYNAMIC;
+
+CREATE INDEX idx_browser_session_authenticated_at ON browser_session (authenticated_at DESC);
+CREATE INDEX idx_browser_session_username_agent ON browser_session (username, user_agent(512));
```
Al2Klimov, line 321: I've implemented our proper compromise.

schema/pgsql/upgrades/025.sql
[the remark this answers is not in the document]
julianbrost: The code that writes to these tables would now receive an error for what worked previously. Given that these were of type `text` before, there are probably no length checks so far.

- For `available_channel_type.type` this affects this function: https://github.com/Icinga/icinga-notifications/blob/32bf26e3217a64697faab1ce9f8497a00e0a9517/internal/channel/plugin.go#L224-L234
- For `object_id_tag.tag` and `object_extra_tag.tag` this affects this function: https://github.com/Icinga/icinga-notifications/blob/32bf26e3217a64697faab1ce9f8497a00e0a9517/internal/event/event.go#L64-L108
- The remaining three columns are written by Web. I guess that might benefit from some "this value is too long" form validation instead of running into a database error.

schema/pgsql/upgrades/025.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE incident ALTER COLUMN severity SET DEFAULT 'ok';
+ALTER TABLE incident_contact ALTER COLUMN role SET DEFAULT 'recipient';
+ALTER TABLE incident_history ALTER COLUMN type SET DEFAULT 'opened';
```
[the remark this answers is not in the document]
julianbrost: The discussion you linked was about integer value ranges. If you use different databases, there will inherently be differences in the behavior. I really don't see why we would want to remove useful checks that are only supported by one database just to be able to say that now all behave poorly. You also added `CHECK` constraints even though they are ignored by some of the [MySQL](https://dev.mysql.com/doc/refman/8.0/en/create-table-check-constraints.html)/[MariaDB](https://mariadb.com/kb/en/constraint/#check-constraints) versions you claim to support:

> Prior to MySQL 8.0.16, CREATE TABLE permits only the following limited version of table CHECK constraint syntax, which is parsed and ignored:
> ```
> CHECK (expr)
> ```

> Constraints are enforced. Before MariaDB 10.2.1 constraint expressions were accepted in the syntax but ignored.

schema/pgsql/upgrades/025.sql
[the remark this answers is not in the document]
Al2Klimov: Actually two. One of the three you likely mean is already a FK, hence a selector – not a text box. But yes, ...

* https://github.com/Icinga/icinga-notifications-web/pull/221

schema/pgsql/upgrades/025.sql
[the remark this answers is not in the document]
julianbrost: > Really, CHECKing only on one RDBMS is pointless.

No, it's not. The check probably represents an assumption made somewhere in the code. With it being checked somewhere, you have a higher chance of noticing earlier during development if you accidentally violate that assumption.

internal/event/event.go
```diff
@@ -68,6 +68,12 @@ func (e *Event) Validate() error {
 		return fmt.Errorf("invalid event: tags must not be empty")
 	}
 
+	for tag := range e.Tags {
+		if len(tag) > 255 {
+			return fmt.Errorf("invalid event: tag %q is too long, at most 255 chars allowed, %d given", tag, len(tag))
+		}
+	}
```
julianbrost, line 75: The same check is also necessary for `e.ExtraTags`.

.github/workflows/tests_with_database.yml
```diff
@@ -7,6 +7,66 @@ on:
...
+        uses: actions/setup-go@v5
+        with:
+          go-version: stable
+
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Importing Schema
+        run: >
+          sleep 15;
```
julianbrost, line 59: Load-bearing sleep? Is there no proper way to wait for the database to be ready?

julianbrost (changes requested):

.github/workflows/tests_with_database.yml
Al2Klimov: Not in MariaDB 11. Ask Yonas, he'll tell you a horror story about this. TL;DR, IGL now uses sleep, too:

* https://github.com/Icinga/icinga-go-library/pull/21

yhabteab: > TL;DR, IGL now uses sleep, too:

That's not true! You've just removed the builtin container health checks from there.

```bash
        options: >-
          --health-cmd ${{ matrix.database.incompat == 'true' && '"healthcheck.sh --connect --innodb_initialized"' || '"mysqladmin ping"' }}
          --health-interval 10s
          --health-timeout 5s
          --health-retries 10
```

julianbrost: @Al2Klimov Does the :+1: mean that the sleep can and will be removed?

schema/pgsql/upgrades/025.sql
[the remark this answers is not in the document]
julianbrost: I think the context of https://github.com/Icinga/icingadb/pull/221#discussion_r756088881 was quite different than this situation:

- It was about the range of allowed integer values in a column. Here it's about intentionally disabling sound input validation in PostgreSQL just because MySQL/MariaDB can't handle it properly.
- For Icinga DB, we ported from MySQL/MariaDB to PostgreSQL, so there you had the additional argument that existing code could possibly rely on the then already present behavior. We're porting the other way around, so all previous code was developed while the restriction was already in place.

Also, if you want to harmonize the behavior of the two implementations, the goal should be to push MySQL/MariaDB towards the sane behavior, not PostgreSQL towards the strange behavior. That might look something like that, but even that looks strange and I don't think it's necessary, but at least it would improve something overall:

```sql
MariaDB [jbrost]> create table enum_check (x enum('a', 'b'), check (x is not null));
Query OK, 0 rows affected (0.027 sec)

MariaDB [jbrost]> insert into enum_check () values ();
ERROR 4025 (23000): CONSTRAINT `CONSTRAINT_1` failed for `jbrost`.`enum_check`
```

.github/workflows/tests_with_database.yml
yhabteab: Yes, you shouldn't have to use a plain sleep command, I've just fixed that in IGL too, see https://github.com/Icinga/icinga-go-library/pull/20. Btw, we don't support `mariadb:10.0`, so why did you add it to the matrix? Apart from that, why are you using images from the `icinga/` namespace for mariadb?

Al2Klimov: The vanilla images miss those settings: https://github.com/Icinga/icingadb/blob/6102d9cb2e30a628ceaf79cc4ba01db1ab50aa8f/doc/02-Installation.md#setting-up-a-mysql-or-mariadb-database

schema/pgsql/upgrades/025.sql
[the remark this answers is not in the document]
Al2Klimov: Oh, not only the enums have magic defaults. 🙈

```
Server version: 5.5.62 MySQL Community Server (GPL)

Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

MySQL [idb]> insert into rotation() values ();
Query OK, 1 row affected, 6 warnings (0,007 sec)

MySQL [idb]> select * from rotation;
+----+-------------+----------+------+------+---------+---------------+----------------+
| id | schedule_id | priority | name | mode | options | first_handoff | actual_handoff |
+----+-------------+----------+------+------+---------+---------------+----------------+
|  1 |           0 |        0 |      | 24-7 |         | 0000-00-00    |              0 |
+----+-------------+----------+------+------+---------+---------------+----------------+
1 row in set (0,002 sec)
```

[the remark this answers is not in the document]
Al2Klimov: Your approach looks nice, but I'm afraid I'll have to make enums like ('NULL', 'a', 'b') for this to work:

```
Server version: 9.0.0 MySQL Community Server - GPL

Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

MySQL [idb]> insert into schedule(name) values ('');
Query OK, 1 row affected (0,009 sec)

MySQL [idb]> alter table rotation add check(mode is not null);
Query OK, 0 rows affected (0,094 sec)
Records: 0  Duplicates: 0  Warnings: 0

MySQL [idb]> insert into rotation(schedule_id,priority,name,options,first_handoff,actual_handoff) values (1,0,'','','2024-01-01',0);
Query OK, 1 row affected (0,040 sec)

MySQL [idb]> select * from rotation;
+----+-------------+----------+------+------+---------+---------------+----------------+
| id | schedule_id | priority | name | mode | options | first_handoff | actual_handoff |
+----+-------------+----------+------+------+---------+---------------+----------------+
|  1 |           1 |        0 |      | 24-7 |         | 2024-01-01    |              0 |
+----+-------------+----------+------+------+---------+---------------+----------------+
1 row in set (0,002 sec)
```

Shall I?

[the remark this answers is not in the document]
julianbrost: > Your approach looks nice, but I'm afraid I'll have to make enums like ('NULL', 'a', 'b') for this to work:

Why? Your output misses the `create table` statement used. The column must not have a `not null` clause inside the column definition, otherwise MySQL would still apply the default before the dedicated `check` constraint.

> Shall I?

No, adding a fake `'NULL'` value looks even worse. Just set no default, you're creating a problem where none exists.

schema/pgsql/schema.sql
```diff
@@ -243,14 +245,15 @@ CREATE TABLE event (
...
+    type event_type,
     severity severity,
     message text,
     username citext,
     mute boolenum,
     mute_reason text,
 
-    CONSTRAINT pk_event PRIMARY KEY (id)
+    CONSTRAINT pk_event PRIMARY KEY (id),
+    CHECK (type IS NOT NULL)
```
julianbrost, line 256: Does this make any difference in behavior? To me, this just looks like it makes the schema definition more complex.

schema/pgsql/upgrades/025.sql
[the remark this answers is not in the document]
julianbrost: > Sorry, overseen that. Your approach is working. But it allows NULL in pre-8.0.16. 🙈 Is this acceptable?

In my opinion, yes. These old versions will go away over time and that problem will solve itself.

A comment that saying that (a) this is a workaround for MySQL/MariaDB implicitly using default values and (b) the constraint can't be relied on old versions would be helpful though.

schema/pgsql/schema.sql
Al2Klimov: It makes the schemas more equal looking to reduce confusion. I'm worried specifically about someone seeing the column definition difference and trying to "fix" the "missing" NOT NULL in MySQL (while overseeing the CHECK). This would cause a regression.

schema/pgsql/upgrades/025.sql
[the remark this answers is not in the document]
Al2Klimov: Actually (b) applies to **all** CHECKs.. 😅

schema/mysql/schema.sql
```diff
@@ -0,0 +1,413 @@
+CREATE TABLE available_channel_type (
+    type varchar(255) NOT NULL,
+    name mediumtext NOT NULL,
+    version mediumtext NOT NULL,
+    author mediumtext NOT NULL,
+    config_attrs mediumtext NOT NULL,
```
Al2Klimov, line 6: I've raised the text columns 64k -> 16M.

schema/pgsql/schema.sql
```diff
@@ -450,7 +452,8 @@ CREATE TABLE browser_session (
     user_agent varchar(4096) NOT NULL,
```
Al2Klimov, line 452: * I've added #239 to OP under _optional_

.github/workflows/tests_with_database.yml
Al2Klimov: In IGL the action steps before schema import just turned out to take long enough for MySQL to meanwhile start up.

julianbrost:
Why are some tables created with `ROW_FORMAT=DYNAMIC` and others without?

Al2Klimov:
Otherwise you get this error:

```
$ perl -pi -e 's/ ROW_FORMAT=DYNAMIC//' schema/mysql/schema.sql
$ mysql -h127.0.0.1 -P6233 -uroot -p1 idb <schema/mysql/schema.sql
ERROR 1709 (HY000) at line 1: Index column size too large. The maximum column size is 767 bytes.
```

But it's not always necessary, so I don't always specify it. (https://github.com/Icinga/icingaweb2/pull/5134#issuecomment-2043145062)

julianbrost:
With which MySQL/MariaDB version did you test that? Is that version specific? With MariaDB 11.4.2, I could import the schema without issues after removing all `ROW_FORMAT=DYNAMIC`. That also matches my expectation if `ROW_FORMAT=DYNAMIC` is the default (as you said in https://github.com/Icinga/icingaweb2/pull/5134#issuecomment-2044479106 and is written in the MySQL documentation for at least versions [5.7](https://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_default_row_format) and [9.0](https://dev.mysql.com/doc/refman/9.0/en/innodb-parameters.html#sysvar_innodb_default_row_format)). Was that different in even older versions?

Al2Klimov:
Exactly. It doesn't work for me up to incl. v5.6.

julianbrost:
That would have been easier if you said that from the beginning.

From the [MySQL 5.7.9 release notes](https://dev.mysql.com/doc/relnotes/mysql/5.7/en/news-5-7-9.html):

> Important Change; InnoDB: DYNAMIC replaces COMPACT as the implicit default row format for InnoDB tables. A new configuration option, [innodb_default_row_format](https://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_default_row_format), specifies the default InnoDB row format. Permitted values include DYNAMIC (the default), COMPACT, and REDUNDANT.
>
> The COMPACT row format remained the default row format until this release to ensure compatibility with older versions of InnoDB in MySQL 5.1 and earlier. Now that MySQL 5.1 has reached the end of its product lifecycle, the newer DYNAMIC row format becomes the default. For information about advantages of the DYNAMIC row format, see [DYNAMIC Row Format](https://dev.mysql.com/doc/refman/5.7/en/innodb-row-format.html#innodb-row-format-dynamic).
>
> Newly created tables use the row format defined by [innodb_default_row_format](https://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_default_row_format) when a ROW_FORMAT option is not specified explicitly or when ROW_FORMAT=DEFAULT is used.
>
> Existing tables retain their current row format if a ROW_FORMAT option was specified explicitly. If a ROW_FORMAT option was not specified explicitly or if ROW_FORMAT=DEFAULT was used, any operation that rebuilds a table also silently changes the row format of the table to the format defined by [innodb_default_row_format](https://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_default_row_format). For more information, see [Defining the Row Format of a Table](https://dev.mysql.com/doc/refman/5.7/en/innodb-row-format.html#innodb-row-format-defining). (WL #8307)

tl;dr: 5.7.9 changed the default from `ROW_FORMAT=COMPACT` to `ROW_FORMAT=DYNAMIC`.

So why don't we want to specify `ROW_FORMAT=DYNAMIC` everywhere? Is there a specific reason that one wants to avoid this row format with older versions if possible? Otherwise, when upgrading the MySQL version, this would keep the `COMPACT` format until some operation rebuilds the table anyways, which differs from the format that would be used if the table was initially created with the newer MySQL version.

Al2Klimov:
My motivation was:

> It is not needed. And consistency is no reason to add it. New code should reflect how it's done properly and not how it was done ages ago.

– @lippserd https://github.com/Icinga/icingaweb2/pull/5134#issuecomment-2044228663

But now as you say it:

> So why don't we want to specify `ROW_FORMAT=DYNAMIC` everywhere? (...) Otherwise, when upgrading the MySQL version, this would keep the `COMPACT` format until some operation rebuilds the table anyways, which differs from the format that would be used if the table was initially created with the newer MySQL version.

This sounds legit (and was even a reason for #226).

schema/mysql/schema.sql
julianbrost: That change doesn't really make sense for most of the columns. The `name` fields are names that users enter in Web, they won't type 64KB+ there. Author and version of a plugin will realistically stay much shorter.

The only columns where I can see that this change makes sense:

- `available_channel_type.config_attrs`: Ff a channel provides many config options and offers translations in all possible languages.
- `channel.config`: With many options set to long values, think of something like pasting multiple X.509 certificates into the config.
- `rotation.options`: maybe, for details you have to ask Web if there's a limit on how big this can get.
- `object.mute_reason`, `event.mute_reason`, `event.message`, `incident_history.message`: Those sound like ones where you want to have a reasonable limit as well. At some point, it's more likely to contain just garbage rather than useful information anyways. For scale, the GPLv2 is roughly 18KB in size, so I think that 64KB should already be more than enough, but having it as `mediumtext` would also allow limiting this to a different size at a later point without having to change the database schema.

Al2Klimov: Definitely a problem of the Postgres schema, if any. But it isn't. If it was an actual problem, you<sub>(plural)</sub> had used varchar(x) in favour of text. The latter [can hold 1G](https://www.postgresql.org/docs/current/datatype-character.html), so my 16M here are peanuts.

```
idb=# update lolcat set s=s||s;
UPDATE 1
idb=# select length(s) from lolcat;
  length
-----------
 536870912
(1 row)

idb=# update lolcat set s=s||s;
ERROR:  invalid memory alloc request size 1073741828
idb=# \d lolcat
             Table "public.lolcat"
 Column | Type | Collation | Nullable | Default
--------+------+-----------+----------+---------
 s      | text |           |          |

idb=#
```

julianbrost:
If this raises the minimal required version (as we discussed), then it has to update it in the documentation as well: https://github.com/Icinga/icinga-notifications/blob/59861289ce06d4b264f1b42e941d3b520cba332a/doc/02-Installation.md#L24

oxzi (commented):
There seems to be an occasionally deadlock during `object.FromEvent` for MySQL and MariaDB, as one can see in some of the Action runs, <https://github.com/Icinga/icinga-notifications/actions/runs/9954287182?pr=203>.

In particular, `TestLoadOpenIncidents` and `TestRestoreMutedObjects` both fail with the exact same error.

```
 === RUN   TestLoadOpenIncidents
    incidents_test.go:165: 
        	Error Trace:	/home/runner/work/icinga-notifications/icinga-notifications/internal/incident/incidents_test.go:165
        	            				/home/runner/work/icinga-notifications/icinga-notifications/internal/incident/incidents_test.go:54
        	Error:      	Received unexpected error:
        	            	failed to insert object extra tags: Error 1213 (40001): Deadlock found when trying to get lock; try restarting transaction
        	Test:       	TestLoadOpenIncidents
```

```
=== RUN   TestRestoreMutedObjects
    objects_test.go:99: 
        	Error Trace:	/home/runner/work/icinga-notifications/icinga-notifications/internal/object/objects_test.go:99
        	            				/home/runner/work/icinga-notifications/icinga-notifications/internal/object/objects_test.go:45
        	Error:      	Received unexpected error:
        	            	failed to insert object extra tags: Error 1213 (40001): Deadlock found when trying to get lock; try restarting transaction
        	Test:       	TestRestoreMutedObjects
```

The code in question:

https://github.com/Icinga/icinga-notifications/blob/edaa61ce5dc8e0f5ad3c636b4161dae49acd9305/internal/object/object.go#L117-L129

I was able to reproduce this locally using `mariadb:10.7` when executing the tests in a loop. After a few iterations, InnoDB deadlocked.

```
2024-07-16 13:27:30 470 [Note] InnoDB: Transactions deadlock detected, dumping detailed information.
2024-07-16 13:27:30 470 [Note] InnoDB:
*** (1) TRANSACTION:

TRANSACTION 184324, ACTIVE 0 sec inserting
mysql tables in use 1, locked 1
LOCK WAIT 7 lock struct(s), heap size 1128, 3 row lock(s), undo log entries 3
MariaDB thread id 470, OS thread handle 140114584385280, query id 733559 172.17.0.1 root Update
INSERT INTO "object_extra_tag" ("object_id", "tag", "value") VALUES (?, ?, ?),(?, ?, ?)
2024-07-16 13:27:30 470 [Note] InnoDB: *** WAITING FOR THIS LOCK TO BE GRANTED:

RECORD LOCKS space id 591 page no 3 n bits 8 index PRIMARY of table "notifications"."object_extra_tag" trx id 184324 lock_mode X insert intention waiting
Record lock, heap no 1 PHYSICAL RECORD: n_fields 1; compact format; info bits 0
 0: len 8; hex 73757072656d756d; asc supremum;;

2024-07-16 13:27:30 470 [Note] InnoDB: *** CONFLICTING WITH:

RECORD LOCKS space id 591 page no 3 n bits 8 index PRIMARY of table "notifications"."object_extra_tag" trx id 184325 lock_mode X
Record lock, heap no 1 PHYSICAL RECORD: n_fields 1; compact format; info bits 0
 0: len 8; hex 73757072656d756d; asc supremum;;

RECORD LOCKS space id 591 page no 3 n bits 8 index PRIMARY of table "notifications"."object_extra_tag" trx id 184324 lock_mode X
Record lock, heap no 1 PHYSICAL RECORD: n_fields 1; compact format; info bits 0
 0: len 8; hex 73757072656d756d; asc supremum;;

2024-07-16 13:27:30 470 [Note] InnoDB:
*** (2) TRANSACTION:

TRANSACTION 184325, ACTIVE 0 sec inserting
mysql tables in use 1, locked 1
LOCK WAIT 7 lock struct(s), heap size 1128, 3 row lock(s), undo log entries 3
MariaDB thread id 469, OS thread handle 140114516154112, query id 733558 172.17.0.1 root Update
INSERT INTO "object_extra_tag" ("value", "object_id", "tag") VALUES (?, ?, ?),(?, ?, ?)
2024-07-16 13:27:30 470 [Note] InnoDB: *** WAITING FOR THIS LOCK TO BE GRANTED:

RECORD LOCKS space id 591 page no 3 n bits 8 index PRIMARY of table "notifications"."object_extra_tag" trx id 184325 lock_mode X insert intention waiting
Record lock, heap no 1 PHYSICAL RECORD: n_fields 1; compact format; info bits 0
 0: len 8; hex 73757072656d756d; asc supremum;;

2024-07-16 13:27:30 470 [Note] InnoDB: *** CONFLICTING WITH:

RECORD LOCKS space id 591 page no 3 n bits 8 index PRIMARY of table "notifications"."object_extra_tag" trx id 184325 lock_mode X
Record lock, heap no 1 PHYSICAL RECORD: n_fields 1; compact format; info bits 0
 0: len 8; hex 73757072656d756d; asc supremum;;

RECORD LOCKS space id 591 page no 3 n bits 8 index PRIMARY of table "notifications"."object_extra_tag" trx id 184324 lock_mode X
Record lock, heap no 1 PHYSICAL RECORD: n_fields 1; compact format; info bits 0
 0: len 8; hex 73757072656d756d; asc supremum;;

2024-07-16 13:27:30 470 [Note] InnoDB: *** WE ROLL BACK TRANSACTION (1)
```

After inspecting this further, the issue seems to be the parallel execution of Go packages in `go test`. Disabling all parallelization `go test -p 1 ./...` "solved" this.

As the initial errors above either rooted in `TestLoadOpenIncidents` or `TestRestoreMutedObjects` from the `internal/incident` or the `internal/object` package, there seems to be the race, resulting in a deadlock.

To be continued...

oxzi:
_Continuing <https://github.com/Icinga/icinga-notifications/pull/203#pullrequestreview-2180268854>._

Turns out, the `DELETE` query sets the lock. When removing it, my test does not run into a deadlock.

https://github.com/Icinga/icinga-notifications/blob/edaa61ce5dc8e0f5ad3c636b4161dae49acd9305/internal/object/object.go#L118-L121

To be continued... again.

.github/workflows/tests_with_database.yml
```diff
@@ -7,6 +7,67 @@ on:
...
+
+      - name: Importing Schema
+        run: |
+          while ! mysql -h127.0.0.1 -u${{ env.NOTIFICATIONS_TESTS_DB_USER }} -p${{
+          env.NOTIFICATIONS_TESTS_DB_PASSWORD }} ${{ env.NOTIFICATIONS_TESTS_DB }} < /dev/null
+          do
+            sleep 1
+          done
+          mysql -h127.0.0.1 -u${{ env.NOTIFICATIONS_TESTS_DB_USER }} -p${{ env.NOTIFICATIONS_TESTS_DB_PASSWORD
+          }} ${{ env.NOTIFICATIONS_TESTS_DB }} < ${{ github.workspace }}/schema/mysql/schema.sql
```
julianbrost, line 65: Is anything preventing the use of regular shell `\` line continuation here?

Also, these are environment variables, so shouldn't simply using `$VAR` instead of `${{ env.VAR }}` work?

And it's the same `mysql` command two times, this could be put in a variable instead, alternatively, try to use `--health-cmd` like it's already done for PostgreSQL and seems to be possible for MySQL/MariaDB as well, see https://github.com/Icinga/icinga-go-library/pull/20.

.github/workflows/tests_with_database.yml
```diff
@@ -7,6 +7,67 @@ on:
...
+          - mysql:5.7
+          - mysql:8
+          - mysql:latest
+          - mariadb:10.2
+          - mariadb:10.3
+          - mariadb:10.4
+          - mariadb:10.5
+          - mariadb:10.6
+          - mariadb:10.7
+          - mariadb:latest
```
julianbrost, line 26: That version list looks outdated. For MySQL, both 8.0 and 8.4 are LTS versions (see https://en.wikipedia.org/wiki/MySQL#Release_history). Similarly, at least MariaDB 10.11, 11.1, and 11.2 should be added (see https://en.wikipedia.org/wiki/MariaDB#Versioning).

doc/02-Installation.md
```diff
@@ -21,22 +21,12 @@ or install [from source](02-Installation.md.d/From-Source.md).
...
-```
-innodb_file_format=barracuda
-innodb_file_per_table=1
-innodb_large_prefix=1
-```
-
-Set up a MySQL database for Icinga Notifications:
-
 ```
 # mysql -u root -p
```
julianbrost, line 31: I wouldn't remove that line.

julianbrost (changes requested):

oxzi:
_Continuing <https://github.com/Icinga/icinga-notifications/pull/203#pullrequestreview-2180268854>, <https://github.com/Icinga/icinga-notifications/pull/203#issuecomment-2231175110>._

The `object_extra_tag` table has a composite primary key of the `object_id` and `tag` columns, being a `binary(32)` and `varchar(255)` respectively.

A `SELECT` query with a `WHERE` clause against a specific `object_id` results in a [ref](https://dev.mysql.com/doc/refman/8.4/en/explain-output.html#jointype_ref) query. However, when creating a `DELETE` query with the same clause, [range](https://dev.mysql.com/doc/refman/8.4/en/explain-output.html#jointype_range) is chosen. Note that the latter has no `ref`.

```
MariaDB [notifications]> explain select * from object_extra_tag where object_id = 0xc08011a318c4479d2db5b9268b3ff0b7e7c3ea53b2da1aaaabb3378b4d98;
+------+-------------+------------------+------+---------------+---------+---------+-------+------+-------------+
| id   | select_type | table            | type | possible_keys | key     | key_len | ref   | rows | Extra       |
+------+-------------+------------------+------+---------------+---------+---------+-------+------+-------------+
|    1 | SIMPLE      | object_extra_tag | ref  | PRIMARY       | PRIMARY | 32      | const | 1    | Using where |
+------+-------------+------------------+------+---------------+---------+---------+-------+------+-------------+
1 row in set (0.001 sec)

MariaDB [notifications]> explain delete from object_extra_tag where object_id = 0xc08011a318c4479d2db5b9268b3ff0b7e7c3ea53b2da1aaaabb3378b4d98;
+------+-------------+------------------+-------+---------------+---------+---------+------+------+-------------+
| id   | select_type | table            | type  | possible_keys | key     | key_len | ref  | rows | Extra       |
+------+-------------+------------------+-------+---------------+---------+---------+------+------+-------------+
|    1 | SIMPLE      | object_extra_tag | range | PRIMARY       | PRIMARY | 32      | NULL | 1    | Using where |
+------+-------------+------------------+-------+---------------+---------+---------+------+------+-------------+
1 row in set (0.001 sec)
```

While I am not completely certain, I would guess based on `SHOW ENGINE INNODB STATUS` that the `DELETE` query creates a [next-key lock](https://dev.mysql.com/doc/refman/8.4/en/innodb-locking.html#innodb-next-key-locks). The locked interval seems to start from the largest value up to infinity, also known as _supremum_.

```
RECORD LOCKS space id 4595 page no 3 n bits 8 index PRIMARY of table "notifications"."object_extra_tag" trx id 696018 lock_mode X insert intention waiting
Record lock, heap no 1 PHYSICAL RECORD: n_fields 1; compact format; info bits 0
 0: len 8; hex 73757072656d756d; asc supremum;;
```

After some tinkering, I was able to reproduce this with to conflicting sessions. First, I selected the maximum `object_id` to insert two values afterwards, to run into the supremum interval.

```
MariaDB [notifications]> select max(id) from object;
+--------------------------------------------------------------------+
| max(id)                                                            |
+--------------------------------------------------------------------+
| 0xFF5916B73A281B19D0F57696F8973FBF8AD6A7DFB85EF62CD6A64641E4DA2DFF |
+--------------------------------------------------------------------+
1 row in set (0.000 sec)
```

| Transaction 1                                                                                                                     | Transaction 2                                                                                                                                                                                                      |
|-----------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `start transaction;`                                                                                                              |                                                                                                                                                                                                                    |
|                                                                                                                                   | `start transaction;`                                                                                                                                                                                               |
| `insert into object (id, source_id, name) values (0xFF5916B73A281B19D0F57696F8973FBF8AD6A7DFB85EF62CD6A64641E4DAF00F, 1, 'foo');` |                                                                                                                                                                                                                    |
|                                                                                                                                   | `insert into object (id, source_id, name) values (0xFF5916B73A281B19D0F57696F8973FBF8AD6A7DFB85EF62CD6A64641E4DAF0FF, 1, 'bar');`                                                                                  |
| `delete from object_extra_tag where object_id = 0xFF5916B73A281B19D0F57696F8973FBF8AD6A7DFB85EF62CD6A64641E4DAF00F;`              |                                                                                                                                                                                                                    |
|                                                                                                                                   | `delete from object_extra_tag where object_id = 0xFF5916B73A281B19D0F57696F8973FBF8AD6A7DFB85EF62CD6A64641E4DAF0FF;`                                                                                               |
| `insert into object_extra_tag values (0xFF5916B73A281B19D0F57696F8973FBF8AD6A7DFB85EF62CD6A64641E4DAF00F, 'foo', '');`            |                                                                                                                                                                                                                    |
|                                                                                                                                   | `insert into object_extra_tag values (0xFF5916B73A281B19D0F57696F8973FBF8AD6A7DFB85EF62CD6A64641E4DAF0FF, 'foo', '');`<br>`ERROR 1213 (40001): Deadlock found when trying to get lock; try restarting transaction` |

Furthermore, I tried adding a new `INDEX` on `object_extra_tag(object_id)`, but this changed nothing. I tried some other options, but those only made things worse. I even achieved an `ALL` type once.

For comparison, the query plan generated by PostgreSQL looks identical, at least for the selection part.

```
notifications=# explain select * from object_extra_tag where object_id = '\xefb1df788174bb134c82056ce6a850916e030ed0aefe1f2d7be659e5c1bcf131';
                                                  QUERY PLAN
---------------------------------------------------------------------------------------------------------------
 Bitmap Heap Scan on object_extra_tag  (cost=4.45..23.19 rows=5 width=56)
   Recheck Cond: (object_id = '\xefb1df788174bb134c82056ce6a850916e030ed0aefe1f2d7be659e5c1bcf131'::bytea)
   ->  Bitmap Index Scan on pk_object_extra_tag  (cost=0.00..4.45 rows=5 width=0)
         Index Cond: (object_id = '\xefb1df788174bb134c82056ce6a850916e030ed0aefe1f2d7be659e5c1bcf131'::bytea)
(4 rows)

notifications=# explain delete from object_extra_tag where object_id = '\xefb1df788174bb134c82056ce6a850916e030ed0aefe1f2d7be659e5c1bcf131';
                                                     QUERY PLAN
---------------------------------------------------------------------------------------------------------------------
 Delete on object_extra_tag  (cost=4.45..23.19 rows=0 width=0)
   ->  Bitmap Heap Scan on object_extra_tag  (cost=4.45..23.19 rows=5 width=6)
         Recheck Cond: (object_id = '\xefb1df788174bb134c82056ce6a850916e030ed0aefe1f2d7be659e5c1bcf131'::bytea)
         ->  Bitmap Index Scan on pk_object_extra_tag  (cost=0.00..4.45 rows=5 width=0)
               Index Cond: (object_id = '\xefb1df788174bb134c82056ce6a850916e030ed0aefe1f2d7be659e5c1bcf131'::bytea)
(5 rows)
```

schema/mysql/schema.sql
julianbrost: The PostgreSQL schema uses `text` because it's the most obvious type given that we didn't bother spending time choosing the specific `x` for `varchar(x)`. However, I think it's obvious that for most columns, a reasonable choice for `x` will be way less than 65536. Thus, I really don't see a reason why `mediumtext` should be the standard choice for the MySQL/MariaDB schema.

julianbrost:
If MySQL isn't able to handle what we do in concurrent transactions, we have to change something anyways. Even without understanding the exact cause yet, I think there are some obvious things that we can try:
1. The transaction currently isn't retried on such a failure. That can be added. That sounds like something we'd want anyway but of course it would be nicer if we wouldn't have to rely on this often.
2. The current implementation was pretty much the easiest one to update the extra tags: delete all existing ones of the object and then (re)insert all. Alternative implementations could be tried so experimentally check if they show better behavior. Something like reading what's currently stored and just applying the delta. Or maybe add something like `and "tag" not in (?, ?, ...)` to the `delete` statement (however, I have low hopes for that one).

oxzi:
> If MySQL isn't able to handle what we do in concurrent transactions, we have to change something anyways. Even without understanding the exact cause yet, I think there are some obvious things that we can try:
> 
>     1. The transaction currently isn't retried on such a failure. That can be added. That sounds like something we'd want anyway but of course it would be nicer if we wouldn't have to rely on this often.

Even though I am not the biggest fan of retrying everything, it might be a strategy here. Especially since situations like this could be lurking anywhere in the codebase.

>     2. The current implementation was pretty much the easiest one to update the extra tags: delete all existing ones of the object and then (re)insert all. Alternative implementations could be tried so experimentally check if they show better behavior. Something like reading what's currently stored and just applying the delta. Or maybe add something like `and "tag" not in (?, ?, ...)` to the `delete` statement (however, I have low hopes for that one).

After brainstorming together with @yhabteab, he suggested the same. Unfortunately, this does not worked either.

Thus, I just moved the logic from the database into the code, which worked, but isn't something I am proud of.

```diff
diff --git a/internal/object/object.go b/internal/object/object.go
index fc15c26..3a3e9c9 100644
--- a/internal/object/object.go
+++ b/internal/object/object.go
@@ -114,14 +114,26 @@ func FromEvent(ctx context.Context, db *database.DB, ev *event.Event) (*Object,
 		return nil, fmt.Errorf("failed to upsert object id tags: %w", err)
 	}
 
-	extraTag := &ExtraTagRow{ObjectId: newObject.ID}
-	_, err = tx.NamedExecContext(ctx, `DELETE FROM "object_extra_tag" WHERE "object_id" = :object_id`, extraTag)
+	var tags []string
+	stmt = db.Rebind(`SELECT "tag" FROM "object_extra_tag" WHERE "object_id" = ?`)
+	err = tx.SelectContext(ctx, &tags, stmt, newObject.ID)
 	if err != nil {
-		return nil, fmt.Errorf("failed to delete object extra tags: %w", err)
+		return nil, fmt.Errorf("failed to fetch object_extra_tags: %w", err)
+	}
+
+	stmt = db.Rebind(`DELETE FROM "object_extra_tag" WHERE "object_id" = ? AND "tag" = ?`)
+	for _, tag := range tags {
+		if _, ok := ev.ExtraTags[tag]; ok {
+			continue
+		}
+		_, err := tx.ExecContext(ctx, stmt, newObject.ID, tag)
+		if err != nil {
+			return nil, fmt.Errorf("failed to delete object extra tags: %w", err)
+		}
 	}
 
 	if len(ev.ExtraTags) > 0 {
-		stmt, _ := db.BuildInsertStmt(extraTag)
+		stmt, _ := db.BuildUpsertStmt(&ExtraTagRow{ObjectId: newObject.ID})
 		_, err = tx.NamedExecContext(ctx, stmt, mapToTagRows(newObject.ID, ev.ExtraTags))
 		if err != nil {
 			return nil, fmt.Errorf("failed to insert object extra tags: %w", err)
```

julianbrost:
> _Edit:_ This might just work because the `DELETE` query is never being executed. In our test cases, the extra tags do not change.. D'oh!

In general, those should stay unchanged most of the time. So in combination with retrying on deadlocks, this might be good enough.

oxzi:
> > _Edit:_ This might just work because the `DELETE` query is never being executed. In our test cases, the extra tags do not change.. D'oh!
> 
> In general, those should stay unchanged most of the time. So in combination with retrying on deadlocks, this might be good enough.

Out of curiosity, I changed the code to `DELETE` each object extra tag directly addressed by both its primary key fields. Well, this also results in a dead lock, as `EXPLAIN` already tells us.

```
MariaDB [notifications]> explain select * from object_extra_tag where object_id = 0xFFFEE3DA6EE137D96E7FDD6757EE7518E7B1A6687884F7A993397DD74DC4D683 and tag = 'servicegroup/webserver';
+------+-------------+------------------+-------+---------------+---------+---------+-------------+------+-------+
| id   | select_type | table            | type  | possible_keys | key     | key_len | ref         | rows | Extra |
+------+-------------+------------------+-------+---------------+---------+---------+-------------+------+-------+
|    1 | SIMPLE      | object_extra_tag | const | PRIMARY       | PRIMARY | 1054    | const,const | 1    |       |
+------+-------------+------------------+-------+---------------+---------+---------+-------------+------+-------+
1 row in set (0.001 sec)

MariaDB [notifications]> explain delete from object_extra_tag where object_id = 0xFFFEE3DA6EE137D96E7FDD6757EE7518E7B1A6687884F7A993397DD74DC4D683 and tag = 'servicegroup/webserver';
+------+-------------+------------------+-------+---------------+---------+---------+------+------+-------------+
| id   | select_type | table            | type  | possible_keys | key     | key_len | ref  | rows | Extra       |
+------+-------------+------------------+-------+---------------+---------+---------+------+------+-------------+
|    1 | SIMPLE      | object_extra_tag | range | PRIMARY       | PRIMARY | 1054    | NULL | 1    | Using where |
+------+-------------+------------------+-------+---------------+---------+---------+------+------+-------------+
1 row in set (0.001 sec)
```

While starting to question what "database engine" even means[^1], I would second the idea to retry deadlocks.

[^1]:  Maybe just three CSV files in a trench coat.

Al2Klimov:
> While starting to question what "database engine" even means[1](#user-content-fn-1-2d4a1b4467ced36b488bf3efc15520ab), I would second the idea to retry deadlocks.
> 
> ## Footnotes
> 1. Maybe just three CSV files in a trench coat. [↩](#user-content-fnref-1-2d4a1b4467ced36b488bf3efc15520ab)

How ironic. You've built up this whole thing on a DB which allows nothing and now a DB which can nothing crosses your way.

schema/mysql/schema.sql
```diff
@@ -0,0 +1,458 @@
...
+    CONSTRAINT fk_channel_available_channel_type FOREIGN KEY (type) REFERENCES available_channel_type(type)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE INDEX idx_channel_changed_at ON channel(changed_at);
+
+CREATE TABLE contact (
+    id bigint NOT NULL AUTO_INCREMENT,
+    full_name mediumtext NOT NULL COLLATE utf8mb4_unicode_ci,
+    username varchar(254) COLLATE utf8mb4_unicode_ci, -- reference to web user
+    default_channel_id bigint NOT NULL REFERENCES channel(id),
```
yhabteab, line 32: You know what to do here :).

yhabteab:
> I don't know exactly how that's even possible but here you're.
> 
> ```shell
> 2024-07-17T16:02:11.160+0200    INFO    incident        Contact "Icinga Admin" role changed from subscriber to manager  {"object": "docker-master!swap", "incident": "#5"}
> 2024-07-17T16:02:11.185+0200    ERROR   incident        Can't insert incident event to the database     {"object": "docker-master!swap", "incident": "#5", "error": "Error 1062 (23000): Duplicate entry '5-348' for key 'PRIMARY'"}
> 2024-07-17T16:02:11.185+0200    ERROR   icinga2 Cannot process event    {"source-id": 1, "event": "[time=2024-07-17 16:02:11.150006 +0200 CEST m=+0.912499501 type=\"acknowledgement-set\" severity=]", "error": "can't insert incident event to the database"}
> 
> ---
> 
> 2024-07-17T16:08:16.259+0200    ERROR   incident        Can't insert incident event to the database     {"object": "docker-master!swap", "incident": "#5", "error": "Error 1452 (23000): Cannot add or update a child row: a foreign key constraint fails (\"notifications\".\"incident_event\", CONSTRAINT \"fk_incident_event_event\" FOREIGN KEY (\"event_id\") REFERENCES \"event\" (\"id\"))"}
> 2024-07-17T16:08:16.260+0200    ERROR   icinga2 Cannot process event    {"source-id": 1, "event": "[time=2024-07-17 16:08:16.231294 +0200 CEST m=+0.969004418 type=\"acknowledgement-set\" severity=]", "error": "can't insert incident event to the database"}
> ```

Fixed in https://github.com/Icinga/icinga-notifications/pull/243 🙈

yhabteab:
Apart from https://github.com/Icinga/icinga-notifications/pull/203#discussion_r1677993630 looks fine for me, I don't know exactly why you replaced `text` with `mediumtext` throughout the schema, even though it's not necessary in most cases.

yhabteab (approved):
LF<sup>TM</sup>.

julianbrost (approved):

oxzi:
_One last thought regarding <https://github.com/Icinga/icinga-notifications/pull/203#issuecomment-2232917159>:_

I just had the idea to explicitly use [index hints](https://dev.mysql.com/doc/refman/8.0/en/index-hints.html) and gave it a last try. First, I have added another `INDEX` being identical to the primary key.

```sql
CREATE INDEX idx_object_extra_tag_object_id_tag ON object_extra_tag(object_id, tag);
```

Then, I crafted a `DELETE` statement using this index, which requires the quirky multi-table `DELETE` syntax:
```
MariaDB [notifications]> explain delete object_extra_tag.* from object_extra_tag use index(idx_object_extra_tag_object_id_tag) where object_id = 0x00D1C60C34E2B0ADF413EAAFCF31D29CE23A632B57FBC3E7F30312A3F8343571\G
*************************** 1. row ***************************
           id: 1
  select_type: SIMPLE
        table: object_extra_tag
         type: ref
possible_keys: idx_object_extra_tag_object_id_tag
          key: idx_object_extra_tag_object_id_tag
      key_len: 32
          ref: const
         rows: 1
        Extra: Using where
1 row in set (0.000 sec)
``` 

As one can see, this query would now be of the `ref` type. Not `const`, but something[^0]. Thus, I have updated Icinga Notifcations, only to run in the same deadlock again. After manually debugging the query and locks[^1] I found out that even this `ref` query creates a next-key lock, potentially hitting the supremum. So far.

<details><summary>Click here to see the abandoned diff</summary>
<p>

```diff
diff --git a/internal/object/object.go b/internal/object/object.go
index 9e54253..be0591e 100644
--- a/internal/object/object.go
+++ b/internal/object/object.go
@@ -114,8 +114,22 @@ func FromEvent(ctx context.Context, db *database.DB, ev *event.Event) (*Object,
                return nil, fmt.Errorf("failed to upsert object id tags: %w", err)
        }

+       if tx.DriverName() == database.MySQL {
+               // Without referring an index, the query will choose the suboptimal "range" join type query strategy. Doing so
+               // will result in a next-key lock, resulting in a potential deadlock.
+               // https://github.com/Icinga/icinga-notifications/pull/203#issuecomment-2232740619
+               //
+               // Note: MySQL/MariaDB only supports indexes for DELETEs when using multi-table statements.
+               // https://dev.mysql.com/doc/refman/8.0/en/index-hints.html
+               stmt = `DELETE object_extra_tag.* FROM "object_extra_tag" USE INDEX (idx_object_extra_tag_object_id_tag)`
+       } else {
+               // PostgreSQL just does the right thing by default :)
+               stmt = `DELETE FROM "object_extra_tag"`
+       }
+       stmt += ` WHERE "object_id" = :object_id`
+
        extraTag := &ExtraTagRow{ObjectId: newObject.ID}
-       _, err = tx.NamedExecContext(ctx, `DELETE FROM "object_extra_tag" WHERE "object_id" = :object_id`, extraTag)
+       _, err = tx.NamedExecContext(ctx, stmt, extraTag)
        if err != nil {
                return nil, fmt.Errorf("failed to delete object extra tags: %w", err)
        }
diff --git a/schema/mysql/schema.sql b/schema/mysql/schema.sql
index 37040cf..8810f14 100644
--- a/schema/mysql/schema.sql
+++ b/schema/mysql/schema.sql
@@ -269,6 +269,10 @@ CREATE TABLE object_extra_tag (
     CONSTRAINT fk_object_extra_tag_object FOREIGN KEY (object_id) REFERENCES object(id)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;

+-- This index is identical to the primary key, but is allowed to be used as an index hint in both MySQL and MariaDB.
+-- Using it for DELETE statements optimizes the query's "join type" from "range" to "ref".
+CREATE INDEX idx_object_extra_tag_object_id_tag ON object_extra_tag(object_id, tag);
+
 CREATE TABLE event (
     id bigint NOT NULL AUTO_INCREMENT,
     time bigint NOT NULL,
```

</p>
</details> 

[^0]: Creating an index just for `object_id` without `key` does not change a thing.
[^1]: Enabling the InnoDB Monitor and Lock Monitor does help and shows more output in `SHOW ENGINE INNODB STATUS`, <https://mariadb.com/kb/en/innodb-monitors/>.

oxzi:
> > First, I have added another `INDEX` being identical to the primary key.
> 
> I'm not entirely sure what the purpose of that second identical index would have been here. Should it have some or did you just miss the following sentence in the documentation? 😅
> 
> > To refer to a primary key, use the name `PRIMARY`.

Honestly, I was not quite sure if this would work in MariaDB, thus I wanted to be certain. For example, MariaDB's "documentation" does not mention `USE KEY` or keys at all, <https://mariadb.com/kb/en/use-index/>.

I just tested it, and both `USE KEY (PRIMARY)` and `USE INDEX (PRIMARY)` do work on MariaDB. However, it "works" as good as the duplicate index does, not very good.

PR #204, opened 2024-05-28
Change schedule configuration to support new rotations

julianbrost:
This is a fundamental and incompatible change to how schedules are defined. Now, a schedule consists of a list of rotations that's ordered by priority. Each rotation contains multiple members where each is either a contact or a contact group. Each member is linked to some timeperiod entries which defines when this member is active in the rotation.

This PR already includes code for a feature that was planned but is not yet possible using the web interface at the moment: multiple versions of the same rotation where the handoff time defines when a given version becomes active.

With this change, for the time being, the TimePeriod type itself fulfills no real purpose and the timeperiod entries are directly loaded as part of the schedule, bypassing the timeperiod loading code. However, there still is the plan to add standalone timeperiods in the future, thus the timeperiod code is kept.

More context for these changes:
- Icinga/icinga-notifications-web#177
- #193

The commits in this PR:
* 87a92b4d04a4bc5ffa9607646a6b3c86adeb4b59, 0fbbe5200608bcbcd28317601415eac28a713ebd: Some code improvements that prevent code duplication in following commits.
* 501ff1c6385061e175c7ba07a2ba35f0f279eb0b: Schema changes from #193 squashed into a single commit.
* b6f524ef7c929961f46ae0369c1a1a5cd993b0b8: I found the proposed upgrade script to be more complex and more destructive to my personal test data than necessary, thus I simplified it. Also, now it looks more in line with the other preexisting schema upgrades as well.
* 86d8e47d3d2e551be95f26e1c2cff3f044e7ead7: The main change of this PR, see general description above.
* 27c1180ed174ccc9323f704d1e440779e5539e43: Always explicitly call `Init()` on timeperiod entries and require this in other methods (as suggested in https://github.com/Icinga/icinga-notifications/pull/204#discussion_r1622606472).
* 83678ae29fd232e3cea79dc088910969f5ffc744: Add debug logging on how recipients were expanded and a warning if no contact was found like it may happen when a schedule has a gap (as suggested in https://github.com/Icinga/icinga-notifications/pull/204#pullrequestreview-2090414546).
* d47fde41171042dd9d7f096559d02eebabea15d1: Add a `/dump-schedules` debug endpoint that returns how all schedules expand within the next two days (as suggested in https://github.com/Icinga/icinga-notifications/pull/204#pullrequestreview-2090414546)

### Dependencies

To allow testing with Icinga Notifications Web, the following PRs are required there:

- https://github.com/Icinga/icinga-notifications-web/pull/183
- https://github.com/Icinga/ipl-html/pull/137
- https://github.com/Icinga/ipl-web/pull/222
- https://github.com/Icinga/ipl-web/pull/223

> [!CAUTION]
> Applying the schema upgrade is destructive to the current schedule configuration and there is no easy way to rollback. Be prepared for this and better create a backup first.

closes #193 (schema change already contained in this PR)

### DIFF
--- a/internal/config/schedule.go
+++ b/internal/config/schedule.go
@@ -3,6 +3,7 @@ package config
 import (
 	"context"
 	"github.com/icinga/icinga-notifications/internal/recipient"
+	"github.com/icinga/icinga-notifications/internal/timeperiod"
 	"github.com/jmoiron/sqlx"
 	"go.uber.org/zap"
 )
@@ -27,26 +28,91 @@ func (r *RuntimeConfig) fetchSchedules(ctx context.Context, tx *sqlx.Tx) error {
 			zap.String("name", g.Name))
 	}
 
-	var memberPtr *recipient.ScheduleMemberRow
-	stmt = r.db.BuildSelectStmt(memberPtr, memberPtr)
+	var rotationPtr *recipient.Rotation
+	stmt = r.db.BuildSelectStmt(rotationPtr, rotationPtr)
 	r.logger.Debugf("Executing query %q", stmt)
 
-	var members []*recipient.ScheduleMemberRow
+	var rotations []*recipient.Rotation
+	if err := tx.SelectContext(ctx, &rotations, stmt); err != nil {
+		r.logger.Errorln(err)
+		return err
+	}
+
+	rotationsById := make(map[int64]*recipient.Rotation)
+	for _, rotation := range rotations {
+		rotationLogger := r.logger.With(zap.Object("rotation", rotation))
+
+		if schedule := schedulesById[rotation.ScheduleID]; schedule == nil {
+			rotationLogger.Warnw("ignoring schedule rotation for unknown schedule_id")
+		} else {
+			rotationsById[rotation.ID] = rotation
+			schedule.Rotations = append(schedule.Rotations, rotation)
+
+			rotationLogger.Debugw("loaded schedule rotation")
+		}
+	}
+
+	var rotationMemberPtr *recipient.RotationMember
+	stmt = r.db.BuildSelectStmt(rotationMemberPtr, rotationMemberPtr)
+	r.logger.Debugf("Executing query %q", stmt)
+
+	var members []*recipient.RotationMember
 	if err := tx.SelectContext(ctx, &members, stmt); err != nil {
 		r.logger.Errorln(err)
 		return err
 	}
 
+	rotationMembersById := make(map[int64]*recipient.RotationMember)
 	for _, member := range members {
-		memberLogger := makeScheduleMemberLogger(r.logger.SugaredLogger, member)
+		memberLogger := r.logger.With(zap.Object("rotation_member", member))
 
-		if s := schedulesById[member.ScheduleID]; s == nil {
-			memberLogger.Warnw("ignoring schedule member for unknown schedule_id")
+		if rotation := rotationsById[member.RotationID]; rotation == nil {
+			memberLogger.Warnw("ignoring rotation member for unknown rotation_member_id")
 		} else {
-			s.MemberRows = append(s.MemberRows, member)
+			member.TimePeriodEntries = make(map[int64]*timeperiod.Entry)
+			rotation.Members = append(rotation.Members, member)
+			rotationMembersById[member.ID] = member
 
-			memberLogger.Debugw("member")
+			memberLogger.Debugw("loaded schedule rotation member")
 		}
+	}
+
+	var entryPtr *timeperiod.Entry
+	stmt = r.db.BuildSelectStmt(entryPtr, entryPtr) + " WHERE rotation_member_id IS NOT NULL"
+	r.logger.Debugf("Executing query %q", stmt)
+
+	var entries []*timeperiod.Entry
+	if err := tx.SelectContext(ctx, &entries, stmt); err != nil {
+		r.logger.Errorln(err)
+		return err
+	}
+
+	for _, entry := range entries {
+		var member *recipient.RotationMember
+		if entry.RotationMemberID.Valid {
+			member = rotationMembersById[entry.RotationMemberID.Int64]
+		}
+
+		if member == nil {
+			r.logger.Warnw("ignoring entry for unknown rotation_member_id",
+				zap.Int64("timeperiod_entry_id", entry.ID),
+				zap.Int64("timeperiod_id", entry.TimePeriodID))
+			continue
+		}
+
+		err := entry.Init()
+		if err != nil {
+			r.logger.Warnw("ignoring time period entry",
+				zap.Object("entry", entry),
+				zap.Error(err))
+			continue
+		}
+
+		member.TimePeriodEntries[entry.ID] = entry
+	}
+
+	for _, schedule := range schedulesById {
+		schedule.RefreshRotations()
 	}
 
 	if r.Schedules != nil {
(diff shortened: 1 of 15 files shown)


julianbrost:
Force push just updated a comment in `rotations.go` and rebased. The latter looks a bit confusing in the diff view: I just renamed `025.sql` to `026.sql` but it shows the differences between the two different and unrelated versions of `025.sql`.

internal/recipient/schedule.go
```diff
@@ -3,50 +3,69 @@ package recipient
...
+	if r.ContactGroupID.Valid {
+		encoder.AddInt64("contact_group_id", r.ContactGroupID.Int64)
+	}
+	return nil
+}
 
-	return contacts
+// GetContactsAt returns the contacts that are active in the schedule at the given time.
+func (s *Schedule) GetContactsAt(t time.Time) []*Contact {
+	return s.rotationResolver.getContactsAt(t)
```
oxzi, line 68: For debugging reasons, it would be nice to have a logging message when no `Contacts` are being returned. This might also happen somewhere up the call stack.

oxzi (commented):
I have successfully tested the changes together with the mentioned other PRs. Everything seems to work, even when it is sometimes a bit hard to try, especially due to UI limitations.

I would like to see more debugging logs and, as discussed offline the other day, maybe another handler to the `Listener` showing the computed schedule. This might also get in handy to compare the calculated rotations against web.

internal/config/schedule.go
```diff
@@ -27,28 +28,93 @@ func (r *RuntimeConfig) fetchSchedules(ctx context.Context, tx *sqlx.Tx) error {
...
+	var members []*recipient.RotationMember
 	if err := tx.SelectContext(ctx, &members, stmt); err != nil {
 		r.logger.Errorln(err)
 		return err
 	}
 
+	rotationMembersById := make(map[int64]*recipient.RotationMember)
 	for _, member := range members {
-		memberLogger := makeScheduleMemberLogger(r.logger.SugaredLogger, member)
+		memberLogger := r.logger.SugaredLogger.With(zap.Object("rotation_member", member))
```
yhabteab, line 67: It is not necessary to access `SugaredLogger` here and in some other places within this package!

internal/timeperiod/timeperiod.go
```diff
@@ -88,11 +135,11 @@ func (e *Entry) Contains(t time.Time) bool {
 		log.Printf("Can't initialize entry: %s", err)
 	}
 
-	if t.Before(e.Start) {
+	if t.Before(e.StartTime.Time()) {
```
yhabteab, line 138: Seeing as you are already touching some code here, can you please adjust the existing error handling here and probably in some other receivers as well, and always rely on the `Entry` being initialised and panic otherwise. That should then make it possible to drop all `log` package usages.

julianbrost:
Most new changes are within the three new additional commits (please see their commit messages and the updated PR description). Changes to to existing commits should only include changes necessary due to the rebase over #114 (changed import, more `db:"-"` needed) and the removal of the unnecessary explicit access of the embedded `SugaredLogger` (https://github.com/Icinga/icinga-notifications/pull/204#discussion_r1622602012).

internal/listener/listener.go
```diff
@@ -220,3 +221,32 @@ func (l *Listener) DumpIncidents(w http.ResponseWriter, r *http.Request) {
...
+	defer l.runtimeConfig.RUnlock()
+
+	for _, schedule := range l.runtimeConfig.Schedules {
+		fmt.Fprintf(w, "[id=%d] %q:\n", schedule.ID, schedule.Name)
+
+		// Iterate in 30 minute steps as this is the granularity Icinga Notifications Web allows in the configuration.
+		// Truncation to seconds happens only for a more readable output.
+		step := 30 * time.Minute
+		start := time.Now().Truncate(time.Second).Add(step)
+		for t := start; t.Before(start.Add(48 * time.Hour)); t = t.Add(step) {
```
yhabteab, line 246: What do you think about making the `start` and `end` time configurable instead of using hard-coded values? Letting users optionally specify the desired start and end time would make it IMHO more flexible and one would be able to evaluate timestamps from the past as well and not just starting from now.

Apart from that, I would change the name of the handler to something like `EvaluateSchedules` since you aren't just dumping the schedules as they are, but evaluating them.

julianbrost: The main intention for providing this was to allow opening this next to Icinga Notifications Web and check for consistency to simplify the review of this PR and possibly to debug inconsistencies between web and the daemon in the future. So my focus was more on keeping the implementation as simple as possible. Do you consider that functionality useful enough to justify the extra code to support it?

yhabteab: Actually, I was just thinking of *was wäre, wenn :)* use cases that aren't really related to this PR, but rather for the future where you can specify an arbitrary start and end time. Though this functionality can also be implemented when it's actually needed, so you can ignore it.

internal/config/schedule.go
```diff
@@ -72,38 +138,29 @@ func (r *RuntimeConfig) applyPendingSchedules() {
...
+				for _, member := range rotation.Members {
+					memberLogger := r.logger.With(
+						zap.Object("rotation", rotation),
+						zap.Object("rotation_member", member))
+
+					if member.ContactID.Valid {
+						member.Contact = r.Contacts[member.ContactID.Int64]
+						if member.Contact == nil {
+							memberLogger.Warnw("ignoring rotation member for unknown contact_id")
+							continue
```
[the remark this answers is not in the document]
julianbrost: > Is the rotation member really ignored in that case?

Actually, it's not and the log message is misleading.

> Unless I'm overlooking something obvious, the member instance still remains in the `rotation.Members` map and will cause the entire `Schedule` being ignored in `debugVerifySchedule()`.

Ignored? That should trigger the following check and therefor trigger a panic:

https://github.com/Icinga/icinga-notifications/blob/0ff18c9e28afb82c3c810b292f2a384253dc2613/internal/config/verify.go#L212-L214
https://github.com/Icinga/icinga-notifications/blob/0ff18c9e28afb82c3c810b292f2a384253dc2613/internal/config/runtime.go#L76-L79

Something similar would also happen if both a contact and a group are set within the same rotation:

https://github.com/Icinga/icinga-notifications/blob/0ff18c9e28afb82c3c810b292f2a384253dc2613/internal/config/verify.go#L216-L218

This would violate the `CHECK (num_nonnulls(contact_id, contactgroup_id) = 1)` constraint, but similarly there should not be invalid IDs due to foreign key constraints, but after all, these are just additional sanity checks.

I think I'd just remove those checks from `debugVerifySchedule()`, the rest of the code will work perfectly fine with neither or both of `member.Contact` and `member.ContactGroup` set. Otherwise, this would just need more code to prevent a situation that should never happen and that wouldn't even cause trouble if it happened.

internal/listener/listener.go
julianbrost: For now, I'd keep the name so that it stays consistent/grouped with the other `/dump-*` endpoints. I'm considering doing another PR that moves all of these to `/debug/*` so that their purpose is more clear and ideally handling auth centrally for all of them.

yhabteab (commented):
Apart from this small change (see comment below), it looks fine for me and works smoothly.

internal/listener/listener.go
```diff
@@ -220,3 +221,32 @@ func (l *Listener) DumpIncidents(w http.ResponseWriter, r *http.Request) {
...
+	l.runtimeConfig.RLock()
+	defer l.runtimeConfig.RUnlock()
+
+	for _, schedule := range l.runtimeConfig.Schedules {
+		fmt.Fprintf(w, "[id=%d] %q:\n", schedule.ID, schedule.Name)
+
+		// Iterate in 30 minute steps as this is the granularity Icinga Notifications Web allows in the configuration.
+		// Truncation to seconds happens only for a more readable output.
+		step := 30 * time.Minute
+		start := time.Now().Truncate(time.Second).Add(step)
```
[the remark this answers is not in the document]
julianbrost: Oh, that `Add()` is indeed a left-over from an earlier version that truncated to 30 minute so that it always gives exactly xx:00 and xx:30 timestamps but that was confusing with how Icinga Notifications Web sets `actual_handoff` (assume its 17:24, you update an entry, then it would have evaluated it for 17:00 and ignore the entry because that's before what was set as `actual_handoff`). That should no longer be an issue when truncating to just seconds though.

yhabteab (approved):

internal/listener/listener.go
```diff
@@ -220,3 +221,32 @@ func (l *Listener) DumpIncidents(w http.ResponseWriter, r *http.Request) {
...
+	l.runtimeConfig.RLock()
+	defer l.runtimeConfig.RUnlock()
+
+	for _, schedule := range l.runtimeConfig.Schedules {
+		fmt.Fprintf(w, "[id=%d] %q:\n", schedule.ID, schedule.Name)
+
+		// Iterate in 30 minute steps as this is the granularity Icinga Notifications Web allows in the configuration.
+		// Truncation to seconds happens only for a more readable output.
+		step := 30 * time.Minute
+		start := time.Now().Truncate(time.Second)
```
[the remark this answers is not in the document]
julianbrost: I did the `Truncate(Second)` to remove the nanosecond precision and the monotonic time from the output in a way where the time shown still exactly matches the time used for evaluation (so I'm not only specifying an output format with limited precision). Truncating to minutes would make little difference in that regard.

```go
package main

import (
	"fmt"
	"time"
)

func main() {
	t := time.Now()

	fmt.Printf("%v\n", t)
	// 2024-06-06 10:49:12.322749534 +0200 CEST m=+0.000012964

	fmt.Printf("%v\n", t.Truncate(time.Second))
	// 2024-06-06 10:49:12 +0200 CEST

	fmt.Printf("%v\n", t.Truncate(time.Minute))
	// 2024-06-06 10:49:00 +0200 CEST
}
```

Editing rotations in Notifications Web still feels weird for me. As of me trying it, it's 10:55 and I'm editing a rotation that's active at that very moment, it sets `actual_handoff` to 11:00, i.e. if I save the currently rotation get disabled until a few minutes in the future. I'd expect it to set `actual_handoff` to now instead, so that it becomes immediately active.

My code was written with that expectation in mind because if it's set to know, you see your change active after a second (assuming clocks are in sync) instead of once the next full minute is reached (or even next half hour if I'd truncate to that).

julianbrost:
Next step: coordinating merging of all the required PRs, that'll be fun.

julianbrost:
> Next step: coordinating merging of all the required PRs, that'll be fun.

That was surprisingly easy.

PR #206, opened 2024-05-29
Channel default value by changing default logic

### LINKED ISSUE #205: Fix default value behavior for channel plugins
We (@nilmerg, @yhabteab, and I) have just discussed this offline and came to the following design conclusion:
- absent keys are resulting in using the default value and
- a `null` value results in an empty value for optional key value pairs.

Those changes must be done in the daemon or, more precisely, the channel plugins and I am going to make the necessary changes tomorrow.

_Originally posted by @oxzi in https://github.com/Icinga/icinga-notifications-web/issues/154#issuecomment-2135583142_

oxzi:
In a nutshell, the newly introduced plugin.PopulateDefaults function
    populates all fields of a Plugin-implementing struct with those fields
    from Info.ConfigAttributes where ConfigOption.Default is set. Thus, a
    single function call before parsing the user-submitted configuration
    within the Plugin.SetConfig method sets default values.

    As a result of the discussion between the Go and the Web team,
    summarized in #205, Web does not store key-value pairs to be omitted.

    Prior, an already JSON-encoded version of the ConfigOption slice was
    present in plugin.Info. Thus, this data wasn't easily available anymore.
    As the new code now needs to access this field, it was changed and a
    custom sql driver.Valuer was implemented for a slice type.

    While working on this, all ConfigOptions were put in the same order as
    the struct fields.

Requires <https://github.com/Icinga/icinga-notifications-web/pull/230>.

Closes #205.

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -1,6 +1,7 @@
 package plugin
 
 import (
+	"database/sql/driver"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -69,13 +70,23 @@ type ConfigOption struct {
 	Max types.Int `json:"max,omitempty"`
 }
 
+// ConfigOptions describes all ConfigOption entries.
+//
+// This type became necessary to implement the database.sql.driver.Valuer to marshal it into JSON.
+type ConfigOptions []ConfigOption
+
+// Value implements database.sql's driver.Valuer to represent all ConfigOptions as a JSON array.
+func (c ConfigOptions) Value() (driver.Value, error) {
+	return json.Marshal(c)
+}
+
 // Info contains plugin information.
 type Info struct {
-	Type             string          `db:"type" json:"-"`
-	Name             string          `db:"name" json:"name"`
-	Version          string          `db:"version" json:"version"`
-	Author           string          `db:"author" json:"author"`
-	ConfigAttributes json.RawMessage `db:"config_attrs" json:"config_attrs"` // ConfigOption(s) as json-encoded list
+	Type             string        `db:"type" json:"-"`
+	Name             string        `db:"name" json:"name"`
+	Version          string        `db:"version" json:"version"`
+	Author           string        `db:"author" json:"author"`
+	ConfigAttributes ConfigOptions `db:"config_attrs" json:"config_attrs"`
 }
 
 // TableName implements the contracts.TableNamer interface.
@@ -129,6 +140,25 @@ type Plugin interface {
 
 	// SendNotification sends the notification, returns an error on failure
 	SendNotification(req *NotificationRequest) error
+}
+
+// PopulateDefaults sets the struct fields from Info.ConfigAttributes where ConfigOption.Default is set.
+//
+// It should be called from each channel plugin within its Plugin.SetConfig before doing any further configuration.
+func PopulateDefaults(typePtr Plugin) error {
+	defaults := make(map[string]any)
+	for _, confAttr := range typePtr.GetInfo().ConfigAttributes {
+		if confAttr.Default != nil {
+			defaults[confAttr.Name] = confAttr.Default
+		}
+	}
+
+	defaultConf, err := json.Marshal(defaults)
+	if err != nil {
+		return err
+	}
+
+	return json.Unmarshal(defaultConf, typePtr)
 }
 
 // RunPlugin reads the incoming stdin requests, processes and writes the responses to stdout
--- a/cmd/channels/email/main.go
+++ b/cmd/channels/email/main.go
@@ -95,7 +95,12 @@ func (ch *Email) Send(reversePath string, recipients []string, msg []byte) error
 }
 
 func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
-	err := json.Unmarshal(jsonStr, ch)
+	err := plugin.PopulateDefaults(ch)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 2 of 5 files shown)


pkg/plugin/plugin.go
```diff
@@ -78,6 +82,55 @@ type Info struct {
...
+	err := defaults.Set(typePtr)
+	if err != nil {
+		return nil, err
+	}
+
+	typeDefaults := make(map[string]any)
+	typeRefT := reflect.TypeOf(typePtr).Elem()
+	typeRefV := reflect.ValueOf(typePtr)
+	for i := 0; i < typeRefT.NumField(); i++ {
+		if !typeRefT.Field(i).IsExported() {
```
yhabteab, line 98: I would simply cache the result of `typeRefT.Field(i)` here instead of repeatedly invoking it.

pkg/plugin/plugin.go
```diff
@@ -78,6 +82,55 @@ type Info struct {
...
+			continue
+		}
+
+		jsonVal, jsonOk := typeRefT.Field(i).Tag.Lookup("json")
+		_, defaultOk := typeRefT.Field(i).Tag.Lookup("default")
+
+		if !jsonOk {
+			return nil, fmt.Errorf("exported field %q misses json struct tag, set to \"-\" to ignore", typeRefT.Field(i).Name)
+		}
```
yhabteab, line 107: Is it really necessary to fail with an error here? Why not just skip it?

pkg/plugin/plugin.go
```diff
@@ -78,6 +82,55 @@ type Info struct {
...
+		if jsonVal == "-" || !defaultOk {
+			continue
+		}
+
+		typeDefaults[jsonVal] = typeRefV.Elem().Field(i).Interface()
```
yhabteab, line 112: I would just use the value from `typeRefT.Field(i).Tag.Lookup(‘default’)` instead of this additional reflection, and you wouldn't have to call `defaults.Set()` at the beginning of the function as well.

pkg/plugin/plugin_test.go
yhabteab, line 1: You might want to use `assert.*` or/and `require.*` instead of manually asserting the test cases.

cmd/channel/email/main_test.go
yhabteab, line 1: Same here!

pkg/plugin/plugin.go
```diff
@@ -78,6 +82,55 @@ type Info struct {
 	ConfigAttributes json.RawMessage `db:"config_attrs" json:"config_attrs"` // ConfigOption(s) as json-encoded list
 }
 
+// NewInfo constructs a new Info to be used in Plugin.GetInfo.
+//
+// The configOptions will be enriched with defaults from the typePtr, being a pointer to the Plugin implementation.
+func NewInfo(typePtr interface{}, name, version, author string, configOptions []*ConfigOption) (*Info, error) {
```
yhabteab, line 88: Shouldn't the type of the `typePtr` argument always be `Plugin` or have I missed something?

pkg/plugin/plugin.go
oxzi: As this function is only being called on known channel plugins, I would like them to fail hard in case something is missing. Otherwise this might only lead to strange errors.

pkg/plugin/plugin.go
oxzi: Fair enough. I haven't set the type to `Plugin` in the first place as none of the `Plugin` interface methods were used, but it makes the API clearer.

pkg/plugin/plugin.go
oxzi: In theory, we already do so. However, currently there are no default - or any - values which are not strings.

https://github.com/Icinga/icinga-notifications/blob/7ff9196bda971ea6741eb42b000c8f7fc4bf1917/pkg/plugin/plugin.go#L31-L34

yhabteab (dismissed):
LFTM!

oxzi:
Rebased after switch to icinga-go-library.

yhabteab (dismissed):

julianbrost:
> To avoid storing each default value twice, it's location was moved up to a "default" struct tag. By utilizing the already included defaults library and adding a small wrapper for plugin.Info, those defaults are now being set to the plugin.ConfigAttributes at runtime.

Have you considered doing it the other way around, i.e. applying the defaults from `ConfigAttributes` to the struct?

I think that this can be done in a much simpler way, without any custom use of reflection even: Iterate over `ConfigAttributes` and populate a `map[string]any` with the defaults (basically, you're already doing this, but it's more difficult to extract the information from the field tags). Then JSON-encode that map, in other words: generate a config like it would be written to the database when it would contain exactly the defaults. Then simply unmarshal it into the config struct before unmarshaling the config from the database.

oxzi:
I liked @julianbrost's suggestion and worked on it. The new design reduces the change's size and is, imo, more readable. Please take another look.

julianbrost (dismissed):
Code change looks fine (and quite a lot nicer than the previous version IMHO), actually using default values by default sounds sane, did not break anything in my test setup including after saving the config from web again, so looks fine to me. Nonetheless, @yhabteab please also have another look at this as you were involved in the discussions mentioned in #205.

oxzi:
For the record, as this changes the channel logic a bit and introduces another helper function, it should either be included in #236 if this got merged first or be an update in this PR otherwise.

julianbrost (approved):

PR #208, opened 2024-05-31
icinga2: Check errors in HTTP status and JSON

oxzi:
The Icinga 2 Event Stream connection did not validate the HTTP status code returned by the server. Consequently, HTTP errors were not identified and their error messages were incorrectly interpreted as valid messages. This occurred in two places: firstly, with regard to the HTTP status code, and secondly, in relation to the error field, which may be present within a JSON struct.

This modification introduces a brief HTTP status validation method that adheres to the specifications outlined in the Icinga 2 API documentation and advanced error checks in UnmarshalEventStreamResponse.

### DIFF
--- a/internal/icinga2/api_responses.go
+++ b/internal/icinga2/api_responses.go
@@ -282,12 +282,23 @@ func UnmarshalEventStreamResponse(bytes []byte) (any, error) {
 	// decompositions in different variables will result in multiple manual fixes. Thus, a two-way deserialization
 	// was chosen which selects the target type based on the first parsed type field.
 
-	var responseType string
+	var (
+		responseType  string
+		responseError int
+	)
 	err := json.Unmarshal(bytes, &struct {
-		Type *string `json:"type"`
-	}{&responseType})
+		Type  *string `json:"type"`
+		Error *int    `json:"error"`
+	}{&responseType, &responseError})
 	if err != nil {
 		return nil, err
+	}
+
+	// Please note: An Event Stream Response SHOULD NOT contain an error field. However, it might be possible that a
+	// message not produced by the Event Stream API might end up here, e.g., a generic API error message. There are
+	// already checks for HTTP error codes in place, so this is more like a second layer of protection.
+	if responseError > 0 {
+		return nil, fmt.Errorf("error field is present, faulty message is %q", bytes)
 	}
 
 	var resp any
--- a/internal/icinga2/client_api.go
+++ b/internal/icinga2/client_api.go
@@ -19,6 +19,25 @@ import (
 
 // This file contains Icinga 2 API related methods.
 
+// checkHTTPResponseStatusCode compares an HTTP response status code against supported Icinga 2 API codes.
+//
+// An error will be returned for invalid status codes according to the documentation. Undefined status codes are going
+// to result in errors as well.
+//
+// https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#http-statuses
+func checkHTTPResponseStatusCode(statusCode int, status string) error {
+	switch {
+	case 200 <= statusCode && statusCode <= 299:
+		return nil
+	case 400 <= statusCode && statusCode <= 499:
+		return fmt.Errorf("invalid HTTP request: %q", status)
+	case 500 <= statusCode && statusCode <= 599:
+		return fmt.Errorf("server HTTP error: %q", status)
+	default:
+		return fmt.Errorf("unknown HTTP error: %q", status)
+	}
+}
+
 // transport wraps http.Transport and overrides http.RoundTripper to set a custom User-Agent for all requests.
 type transport struct {
 	http.Transport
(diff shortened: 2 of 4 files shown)


internal/icinga2/api_responses.go
```diff
@@ -282,14 +282,22 @@ func UnmarshalEventStreamResponse(bytes []byte) (any, error) {
 	// decompositions in different variables will result in multiple manual fixes. Thus, a two-way deserialization
 	// was chosen which selects the target type based on the first parsed type field.
 
-	var responseType string
+	var (
+		responseType  string
+		responseError interface{}
```
yhabteab, line 287: With `go1.22` we should prefer `any` over `interface{}`, I think :), but shouldn't the json key [`error`](https://github.com/Icinga/icinga2/blob/master/lib/remote/httputility.cpp#L67) always be of type `int`?

internal/icinga2/client_api.go
```diff
@@ -91,10 +110,11 @@ func (client *Client) queryObjectsApi(
 		return nil, err
 	}
 
-	if res.StatusCode != http.StatusOK {
+	err = checkHTTPResponseStatusCode(res.StatusCode)
+	if err != nil {
 		_, _ = io.Copy(io.Discard, res.Body)
 		_ = res.Body.Close()
-		return nil, fmt.Errorf("unexpected HTTP status code %d", res.StatusCode)
+		return nil, err
```
yhabteab, line 117: Doesn't this hide the actual status message from Icinga 2? I would either add the `resp.status` message to the error as well or simply derive an error from it, since it should contain enough information/context as to what went wrong.

internal/icinga2/api_responses.go
yhabteab: I have not tested if it is the case here as well, but Icinga 2 doesn't send back json encoded error if the `accept` header is not set to `application/json`.

```bash
$ curl -kvsSiu root:icingas 'https://localhost:5665/v1/objects/hosts/docker-master?pretty=1' -H 'Accept: application/json'
---
* Closing connection
{"error":401,"status":"Unauthorized. Please check your user credentials."}%

$ curl -kvsSiu root:icingas 'https://localhost:5665/v1/objects/hosts/docker-master?pretty=1'
---
* Closing connection
<h1>Unauthorized. Please check your user credentials.</h1>%
```

oxzi: > With `go1.22` we should prefer `any` over `interface{}`, I think :)

:+1: 

> but shouldn't the json key [`error`](https://github.com/Icinga/icinga2/blob/master/lib/remote/httputility.cpp#L67) always be of type `int`?

I wasn't able to find this in the documentation, but this was also what I would expect. I will change it.

internal/icinga2/client_api.go
oxzi: Nope, the error returned from `checkHTTPResponseStatusCode` contains the status code if it is seen as being invalid.

internal/icinga2/api_responses.go
oxzi: Regarding the content not being JSON: in this case the HTTP status check should fire beforehand or, otherwise, the JSON unmarshalling should fail as HTML should not be able to be decoded as JSON.

internal/icinga2/client_api.go
yhabteab: Are you aware of `resp.Status` **<>** `resp.StatusCode`?

internal/icinga2/client_api.go
```diff
@@ -19,6 +19,25 @@ import (
...
+func checkHTTPResponseStatusCode(statusCode int, status string) error {
+	switch {
+	case 200 <= statusCode && statusCode <= 299:
+		return nil
+	case 400 <= statusCode && statusCode <= 499:
+		return fmt.Errorf("invalid HTTP request: %d, %q", statusCode, status)
+	case 500 <= statusCode && statusCode <= 599:
+		return fmt.Errorf("server HTTP error: %d, %q", statusCode, status)
+	default:
+		return fmt.Errorf("unknown HTTP error: %d, %q", statusCode, status)
```
yhabteab, line 37: It looks like the HTTP error code is already included in the status message, as documented (`http.Response.Status string // e.g. "200 OK"`), so you don't need to manually add it to the error message.

```bash
2024-06-03T09:22:35.039+0200	ERROR	icinga2	Establishing an Event Stream API connection failed with an HTTP error, will be retried	{"source-id": 1, "error": "invalid HTTP request: 401, \"401 Unauthorized\"", "delay": "1s"}
2024-06-03T09:22:36.096+0200	ERROR	icinga2	Establishing an Event Stream API connection failed with an HTTP error, will be retried	{"source-id": 1, "error": "invalid HTTP request: 401, \"401 Unauthorized\"", "delay": "2s"}
---
2024-06-03T09:23:04.190+0200	ERROR	icinga2	Establishing an Event Stream API connection failed with an HTTP error, will be retried	{"source-id": 1, "error": "invalid HTTP request: 404, \"404 Not Found\"", "delay": "4s"}
2024-06-03T09:23:08.252+0200	ERROR	icinga2	Establishing an Event Stream API connection failed with an HTTP error, will be retried	{"source-id": 1, "error": "invalid HTTP request: 404, \"404 Not Found\"", "delay": "8s"}
```

oxzi: I was afraid that it might get redacted anyhow, but it seems like this is impossible.
https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/net/http/response.go;l=168-173

yhabteab (dismissed):
LFTM!

julianbrost (approved):

PR #209, opened 2024-05-31
icinga2: Configurable Icinga 2 API HTTP Timeout

oxzi:
First, raise the default HTTP timeout for Icinga 2 API connections to one minute. This became necessary on one development system for high load during startup.

As, however, one minute is kinda arbitrary, it was made configurable in the daemon's YAML configuration file.

~~The already huge timeout of three seconds was hit on a development system with a multitude of objects. Thus, increase it to an even huger value and shortly document the motivation for doing so.~~


yhabteab (approved):
LFTM! Requests that require considerably more time than that, are IMHO not normal.

julianbrost:
> one minute is still a very huge time which shouldn't be reached

While I'd wish for it to be true, a request to `/v1/objects/services` can return a quite huge response (in the hundreds of megabytes) that's slow to generate. A minute sounds long, but I wouldn't bet on that it's never exceeded, especially on startup of the whole system where there's extra load on everything. How much effort would it be to add a config option for the timeout so that it can at least be tweaked without requiring recompilation? (I'd probably go for the config file as that's more of a fine tuning knob that would be somewhat strange to always present in the web interface.)

oxzi:
I have altered my precious one-line-diff according to @julianbrost's valid critique. Now it also defaults to one minute, but is configurable through the daemon's YAML configuration file.

oxzi:
Rebased after switch to icinga-go-library.

internal/icinga2/client.go
```diff
@@ -44,8 +43,19 @@ type Client struct {
 	ApiBaseURL       string
 	ApiBasicAuthUser string
 	ApiBasicAuthPass string
+
+	// ApiHttpTransport is a shared http.Transport and http.RoundTripper for all API connections, set by the Launcher.
 	ApiHttpTransport http.RoundTripper
 
+	// ApiTimeout specifies the timeout for API connections outside the Event Stream API.
+	//
+	// Defaults to 1m if unset within Client.Process.
```
yhabteab, line 52: The way I interpret this is that *it defaults to 1m somewhere else if it's not set in the `Client.Process` method*, you might want to phrase that a little differently, since that doesn't reflect the actual implementation.

internal/daemon/config.go
```diff
@@ -7,12 +7,14 @@ import (
...
 	"github.com/icinga/icinga-go-library/logging"
 	"os"
+	"time"
 )
 
 type ConfigFile struct {
 	Listen           string          `yaml:"listen" default:"localhost:5680"`
 	DebugPassword    string          `yaml:"debug-password"`
 	ChannelPluginDir string          `yaml:"channel-plugin-dir" default:"/usr/libexec/icinga-notifications/channel"`
+	ApiTimeout       time.Duration   `yaml:"api-timeout" default:"1m"`
```
yhabteab, line 17: Please add this to the sample YAML file as well.

yhabteab (approved):

julianbrost (approved):

PR #210, opened 2024-06-03
doc: Initialize documentation

oxzi:
Both structure and huge parts are copied from Icinga DB and were altered afterwards. The already existing Channel Plugin documentation was extended.


oxzi:
To build the docs, please refer to this PR: https://github.com/Icinga/icinga-docs-tools/pull/9

doc/02-Installation.md.d/From-Source.md
julianbrost, line 1: If anything, I'd document for development only. I would not advice any user to do `go build -o /usr/libexec/icinga-notifications/channels`.

I presume this installation from source section is a relic from when many/most Icinga Web modules did not have proper packages. The binary components like Icinga 2 and Icinga DB don't provide such a section as that's not the recommended installation method.

config.example.yml
```diff
@@ -7,7 +7,7 @@
 #debug-password: "put-something-secret-here"
 
 icingaweb2-url: http://localhost/icingaweb2/
-channel-plugin-dir: /usr/libexec/icinga-notifications/channel
+channel-plugin-dir: /usr/libexec/icinga-notifications/channels
```
julianbrost, line 10: Doing that change just here introduces an inconsistency:

On the PR branch:
```console
$ git grep libexec     
Dockerfile:COPY --from=build /src/icinga-notifications/bin/channel /usr/libexec/icinga-notifications/channel
config.example.yml:channel-plugin-dir: /usr/libexec/icinga-notifications/channels
doc/02-Installation.md.d/From-Source.md:go build -o /usr/libexec/icinga-notifications/channels ./cmd/channel/...
doc/03-Configuration.md:For a package install, this directory should be `/usr/libexec/icinga-notifications/channels`.
internal/daemon/config.go:      ChannelPluginDir string                  `yaml:"channel-plugin-dir" default:"/usr/libexec/icinga-notifications/channel"`
```

doc/02-Installation.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+Set up a MySQL database for Icinga Notifications:
+
+```
+# mysql -u root -p
+
+CREATE DATABASE icinga-notifications;
+CREATE USER 'icinga-notifications'@'localhost' IDENTIFIED BY 'CHANGEME';
+GRANT ALL ON icinga-notifications.* TO 'icinga-notifications'@'localhost';
```
julianbrost, line 41: Using `-` in the database and user names requires additional quoting. I'd suggest using `_` instead. Alternatively, other components also use very generic database names in their documentation as well (like `director`, `reporting`, `x509`, yes, as the full database names), so we might also go with just `notifications`.

doc/02-Installation.md
```diff
@@ -0,0 +1,125 @@
...
+
+```
+psql -U icinga-notifications icinga-notifications < /usr/share/icinga-notifications/schema/pgsql/schema.sql
+```
+
+## Configuring Icinga Notifications
+
+Icinga Notifications installs its configuration file to `/etc/icinga-notifications/config.yml`,
+pre-populating most of the settings for a local setup. Before running Icinga Notifications,
+adjust the database credentials and, if necessary, the Icinga Web 2 URL.
```
julianbrost, line 84: I'd omit "if necessary". I doubt there are scenarios where you wouldn't want to set it?

doc/02-Installation.md
```diff
@@ -0,0 +1,125 @@
...
+[installing Icinga Notifications Web on RHEL](https://icinga.com/docs/icinga-notifications-web/latest/doc/02-Installation/04-RHEL/#installing-icinga-notifications-web-package),
+<!-- {% endif %} -->
+<!-- {% if sles %} -->
+[installing Icinga Notifications Web on SLES](https://icinga.com/docs/icinga-notifications-web/latest/doc/02-Installation/05-SLES/#installing-icinga-notifications-web-package),
+<!-- {% endif %} -->
+<!-- {% if ubuntu %} -->
+[installing Icinga Notifications Web on Ubuntu](https://icinga.com/docs/icinga-notifications-web/latest/doc/02-Installation/06-Ubuntu/#installing-icinga-notifications-web-package),
+<!-- {% endif %} -->
+which will also guide you through the setup of the Icinga Web PHP framework,
+which is required to run the Icinga Notifications web module.
```
julianbrost, line 123: > Icinga Web PHP framework

I haven't heard that name before. Does that mean IPL? In anyway, sounds like a level of detail that wouldn't be necessary here anyway and something like "its dependencies" should suffice.

doc/03-Configuration.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Top Level Configuration
+
+### Listener Configuration
+
+The HTTP listener can be used both for submission and for debugging purposes.
+
+| Option         | Description                                                          |
+|----------------|----------------------------------------------------------------------|
+| listen         | Address to be bound to.                                              |
```
julianbrost, line 17: "Address to bind to" sounds better, IMHO. Also, address is unclear whether this includes a port, so an example would be nice.

doc/03-Configuration.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Icinga Web 2
+
+The `icingaweb2-url` is expected to point to the base directory of your Icinga Web 2 installation,
+i.e., `https://example.com/icingaweb2/`, to be used for URL creation.
+
+### Channel Plugin Directory
+
+All available Icinga Notification Channel Plugins should resident in the `channel-plugin-dir`.
+For a package install, this directory should be `/usr/libexec/icinga-notifications/channels`.
```
julianbrost, line 28: For a package install, it should not be necessary to have this option in the config file at all.

julianbrost (changes requested):
Where will be documented how to connect the daemon to Icinga 2? Do you see this as part of the Notifications Web documentation as that provides the config form? Even if that will be the case, a hint on this somewhere might be nice.

doc/02-Installation.md.d/From-Source.md
oxzi: Yes, it was based on/inspired by some web-thingy I took as a foundation. I reworked this to refer to the newly introduced Makefile.

doc/03-Configuration.md
oxzi: Fair enough. I have rephrased this section to mention that this configuration option shouldn't be needed for package installs. However, I dislike hiding some information which might be useful to some users in some special setups.

doc/01-About.md
```diff
@@ -0,0 +1,35 @@
+# Icinga Notifications
+
+Icinga Notifications is a set of components that processes received events from various sources, manages incidents and
+forwards notifications to predefined contacts, consisting of:
+
+* The Icinga Notifications daemon (this project), which receives events and sends notifications.
+* An [Icinga Web module](https://github.com/Icinga/icinga-notifications-web) that provides graphical configuration and further processing of the data collected by the daemon.
+* And Icinga 2 and other custom sources that propagate state updates and acknowledgement events to the daemon.
```
oxzi, line 8: 8uxtuo

oxzi:
I have made some changes inspired by @ncosta-ic's similar PR over there in web world, Icinga/icinga-notifications-web#218.

doc/03-Configuration.md
```diff
@@ -0,0 +1,117 @@
...
+| debug-password | Password expected via HTTP Basic Authentication for debug endpoints. |
+
+### Icinga Web 2
+
+The `icingaweb2-url` is expected to point to the base directory of your Icinga Web 2 installation,
+i.e., `https://example.com/icingaweb2/`, to be used for URL creation.
+
+### Channel Plugin Directory
+
+All available Icinga Notification Channel Plugins should resident in the `channel-plugin-dir`.
```
julianbrost, line 30: s/resident/reside/

doc/03-Configuration.md
```diff
@@ -0,0 +1,117 @@
...
+For a package install, this directory is `/usr/libexec/icinga-notifications/channels` and should not be altered.
```
julianbrost, line 31: The default may differ based on the system. On Suse, it will be inside `/usr/lib/`. Apart from that, it's still singular "channel" for the time being:

https://github.com/Icinga/icinga-notifications/blob/e6f870d03de33f512f9ae8811c48c78ff3f9de11/internal/daemon/config.go#L27

Feel free to create a PR that changes this consistenly.

julianbrost (changes requested):

doc/02-Installation.md.d/From-Source.md
oxzi: The very first sentence states that this way isn't recommended. As a new user looking into the docs I would expect to find some generic[^1] installation instructions somewhere under "Installation".

I am aware that most (or all?) Icinga project documentations misses a generic installation section, but arguing in this direction just brings us to some chicken-and-egg situation.

Further, I would say that a development section has a slightly different target. However, I am in favor for a brief development documentation, easing the potential onboarding and especially (third-party) channel plugin development.

Thus, I am going to add such a development section and would ask you again regarding your opinion about a generic "From Source" installation section.

[^1]: Unix-based operating systems are generic enough.

doc/01-About.md
```diff
@@ -0,0 +1,41 @@
...
+    This is an early preview version for you to try, but do not use this in production.
+    There may still be severe bugs and incompatible changes may happen without any notice.
+    At the moment, we don't yet provide any support for this.
+
+Icinga Notifications is a set of components that processes received events from various sources, manages incidents and
+forwards notifications to predefined contacts, consisting of:
+
+* The Icinga Notifications daemon (this project), which receives events and sends notifications.
+* An [Icinga Web module](https://github.com/Icinga/icinga-notifications-web) that provides graphical configuration and further processing of the data collected by the daemon.
+* Icinga 2 and other custom sources that propagate state updates and acknowledgement events to the daemon.
```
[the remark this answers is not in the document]
oxzi: This sentence is from the current `README.md`. I rephrased it in both places.

doc/03-Configuration.md
oxzi: > On Suse, it will be inside `/usr/lib/`.

[_[Angry Filesystem Hierarchy Standard Noises]_](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s06.html)

> Feel free to create a PR that changes this consistenly.

Will do.

doc/01-About.md
oxzi, line 1: _(This note is not related to this file in particular, I just wanted to create a thread and don't know another way to do this here.)_

I am a bit unhappy with the numbering of the markdown files by their filename. This layout was copied from the other Icinga documentations and I would expect a certain degree of similarity between the projects.

However, this layout enforces a ordering by the names from the beginning, which is hard to change later, as it will break links, resulting in dead URLs. For example, in Icinga DB I once wanted to add a new section between two already existing ones, but there is no real possibility to add something in between[^1].

It would be possible to leave greater gaps, but this would also look weird. I would like to hear your comment on this, @julianbrost. Thanks :)

[^1]: Except of hacks like naming something `03-01-Foo.md` if it should go between `03-Configuration.md` and `04-Upgrading.md`.

oxzi: > This layout was copied from the other Icinga documentations and I would expect a certain degree of similarity between the projects.

If we would consider breaking our own conventions, we could manually populate MKDocs' `nav` and don't prefix files with numbers. This, however, would also require patching the `build-docs.rb` script.

julianbrost: To be honest, I never thought about this. But indeed, it doesn't really make sense to me that the URL for example is https://icinga.com/docs/icinga-2/latest/doc/14-features/ instead of just https://icinga.com/docs/icinga-2/latest/doc/features/.

> we could manually populate MKDocs' `nav` and don't prefix files with numbers. This, however, would also require patching the `build-docs.rb` script.

I'm not familiar with MKDocs either, so I don't know what its `nav` is, but it doesn't sound wrong to me.

By the way, leaving gaps wouldn't be unheard of, see https://github.com/Icinga/icingaweb2-module-director/tree/master/doc and https://github.com/Icinga/icinga-powershell-framework/tree/master/doc for example.

ncosta-ic: Indeed, proper permalinks regardless of the given file name, would be really handy. There's a whole discussion about it over at https://github.com/squidfunk/mkdocs-material/discussions/3758

The whole `build-docs.rb` wrapper is pretty hacky alltogether and even destroyed my local git repository because of the wonderful `rm -rf` instruction contained in it...
A better solution would be to rewrite our existing `build-docs.rb` functionality into a Python hook[^1].
We could also include our own solution to the url problem by doing so.

[^1]: https://www.mkdocs.org/user-guide/configuration/#hooks

julianbrost: I'm all for improving things there if you've looked into it already and have ideas how to do it better. But sounds something to better track in central place rather than here. If we change something, that would affect most projects anyways. Also, this shouldn't block anything here, the only question is whether you want to add gaps in the numbers for the moment.

By the way, keeping URLs working with a change in that direction should be simple by redirecting `NNN-foobar` to `foobar`.

doc/20-HTTP-API.md
```diff
@@ -0,0 +1,47 @@
...
+[Event](https://github.com/Icinga/icinga-notifications/blob/main/internal/event/event.go).
+
+```
+curl -v \
+  -u 'source-1:correct horse battery staple' \
+  -d '{"extra_tags":{"hostgroup/app-container":null ,"hostgroup/department-dev":null,"hostgroup/env-qa":null,"hostgroup/location-rome":null,"servicegroup/app-mail":null,"servicegroup/department-nms":null,"servicegroup/env-prod":null,"servicegroup/location-berlin":null},"message":"You will be honored for contributing your time and skill to a worthy cause.","name":"dummy-809: random fortune","severity":"ok","source_id":2,"tags":{"host":"dummy-809","service":"random fortune"},"url":"http://localhost/icingaweb2/icingadb/service?name=random%20fortune&host.name=dummy-809","username":""}' \
+  'http://localhost:5680/process-event'
+```
+
+## Debugging Endpoints
```
oxzi, line 20: #223 would change the following URLs. However, this isn't a blocker as a subsequent PR can update this.

doc/01-About.md
oxzi: For the moment, I added some gaps between the numbering of the later sections as one might expect something to pop up there.

After some potential future change, a simple `RewriteRule ^/doc/([a-z0-9-]+)/[0-9]+-(.*) /doc/$1/$2 [R]`[^1] should do, as @julianbrost also suggested.

[^1]: Untested, of course :upside_down_face:

doc/03-Configuration.md
oxzi: → #225

doc/30-Development.md
```diff
@@ -0,0 +1,41 @@
...
+
+```
+git clone https://github.com/Icinga/icinga-notifications.git
+cd icinga-notifications
+```
+
+To specify destinations for the final install,
+consult the variables in the `Makefile` which can be overwritten by environment variables.
+
+In particular, the `DESTDIR` variable should be noted,
```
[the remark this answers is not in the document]
oxzi: Oopsi.

doc/02-Installation.md.d/From-Source.md
oxzi: I ended up with a half-baked "Development" section to get people starting compiling the code for development and linked it from the From Source section. Thus, people will have a reference when looking at the installation section, but all the details are somewhere else.

doc/10-Channels.md
```diff
@@ -1,4 +1,21 @@
-## Channel Plugin
+# Channels
+
+After Icinga Notifications decides to send a notification of any kind, it will be passed to a channel plugin.
+Such a channel plugin submits the notification event to a channel, e.g., email or a chat client.
+
+Icinga Notifications comes packed with channel plugins, but also enables you to develop your own plugins.
+
+To make those plugins available to Icinga Notifications, they must be placed in the [plugin channel directory](03-Configuration.md#channel-plugin-directory),
```
julianbrost, line 8: > plugin channel directory

Well that's a new name, I don't think it's called like this anywhere else.

julianbrost (changes requested):

doc/10-Channels.md
oxzi: Just another permutation of the channel directory plugin ^W^W^W directory plugin channel ^W^W^W  ... :upside_down_face: 

With #225 it got "simplified" to `channels-dir` and I have reflected this change here. Thanks.

doc/03-Configuration.md
```diff
@@ -0,0 +1,120 @@
...
+| debug-password | Password expected via HTTP Basic Authentication for debug endpoints. |
+
+### Icinga Web 2
+
+The `icingaweb2-url` is expected to point to the base directory of your Icinga Web 2 installation,
+i.e., `https://example.com/icingaweb2/`, to be used for URL creation.
+
+### Channels Directory
+
+All available Icinga Notification Channels should reside in the `channels-dir` directory.
```
oxzi, line 30: Requires #225.

doc/03-Configuration.md
```diff
@@ -0,0 +1,120 @@
...
+For a package installation, the default will point to the correct location and must not be changed.
```
julianbrost, line 31: I'd write it as "Icinga Notifications channels", i.e. one less upper-case letter.

doc/20-HTTP-API.md
```diff
@@ -0,0 +1,46 @@
...
+
+One possible source next to the Icinga 2 API is event submission to the Icinga Notification HTTP API listener.
+After creating a source with type _Other_ in Icinga Notifications Web,
+the specified credentials can be used for HTTP Basic Authentication of a JSON-encoded
+[Event](https://github.com/Icinga/icinga-notifications/blob/main/internal/event/event.go).
+
+```
+curl -v -u 'source-1:insecure' -d '@-' 'http://localhost:5680/process-event' <<EOF
+{"extra_tags":{"hostgroup/app-container":null ,"hostgroup/department-dev":null,"hostgroup/env-qa":null,"hostgroup/location-rome":null,"servicegroup/app-mail":null,"servicegroup/department-nms":null,"servicegroup/env-prod":null,"servicegroup/location-berlin":null},"message":"You will be honored for contributing your time and skill to a worthy cause.","name":"dummy-809: random fortune","severity":"ok","source_id":2,"tags":{"host":"dummy-809","service":"random fortune"},"url":"http://localhost/icingaweb2/icingadb/service?name=random%20fortune&host.name=dummy-809","username":""}
+EOF
```
julianbrost, line 16: The idea with heredoc was so that it can contain newlines without issues and this could be some pretty-printed JSON.

julianbrost (changes requested):

doc/20-HTTP-API.md
oxzi: Can do. It's just questionable how valuable this JSON is, as this was just the first matching entry in my shell history :upside_down_face:

doc/20-HTTP-API.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Process Event
+
+One possible source next to the Icinga 2 API is event submission to the Icinga Notification HTTP API listener.
+After creating a source with type _Other_ in Icinga Notifications Web,
+the specified credentials can be used for HTTP Basic Authentication of a JSON-encoded
+[Event](https://github.com/Icinga/icinga-notifications/blob/main/internal/event/event.go).
+
+The authentication is performed via HTTP Basic Authentication, expecting `source-${id}` as the username,
+`${id}` being the source's `id` within the database, and the configured password.
```
oxzi, line 14: #227 would change this.

doc/02-Installation.md
```diff
@@ -0,0 +1,117 @@
...
+The configuration file explains general settings.
+All available settings can be found under [Configuration](03-Configuration.md).
+
+## Configuring Icinga 2
+
+There is no special configuration required for Icinga 2 on its own.
+However, you may want to use an Icinga 2 instance as a notification source.
+This requires configuring an `ApiUser` as described later in the
+[Icinga Notifications Web documentation](https://icinga.com/docs/icinga-notifications-web/latest/doc/03-Configuration/).
+At this point please continue with this manual until the Icinga Notification Web documentation leads there.
```
julianbrost, line 98: Do we need this section at all? This sounds more confusing than helpful to me. It has a reference to Icinga Notifications Web, but doesn't want you to follow it but instead ignore it and continue with the following step here anyways.

doc/20-HTTP-API.md
```diff
@@ -0,0 +1,70 @@
+# HTTP API
+
+Icinga Notifications comes with its own HTTP API, [configurable](03-Configuration.md#http-api-configuration)
+via `listen` and `debug-password`.
+
+## Process Event
+
+One possible source next to the Icinga 2 API is event submission to the Icinga Notification HTTP API listener.
```
julianbrost, line 8:
```suggestion
One possible source next to the Icinga 2 API is event submission to the Icinga Notifications HTTP API listener.
```

doc/02-Installation.md
oxzi: I thought that keeping this section with a reference to the web part was the consensus of the discussion earlier. But yes, it also felt weird for me, that's why I added the last sentence, stating that this doesn't really matter at this point.

However, I think I am going to remove this here.

doc/01-About.md
```diff
@@ -0,0 +1,55 @@
...
+forwards notifications to predefined contacts, consisting of:
+
+* The Icinga Notifications daemon, which receives events and sends notifications.
+* The [Icinga Notifications Web](https://icinga.com/docs/icinga-notifications-web/latest/doc/01-About/) module,
+  which provides graphical configuration.
+* Icinga 2 or other sources that provide monitoring events that result in notifications.
+
+## Big Picture
+
+<!-- TODO: add a real picture -->
```
julianbrost, line 19: This should ideally be replaced by a real picture before merging.

julianbrost (commented):
In general, fine for me now. (To be fair, I didn't cross check with the web documentation for possible duplicates, but I wouldn't expect any of what's in here to be duplicated apart from the about page where it's to be expected.)

Just that one obvious thing, but better have it as an unresolved conversation than accepting and forgetting it:

README.md
```diff
@@ -1,40 +1,20 @@
...
 Icinga Notifications is a set of components that processes received events from various sources, manages incidents and
 forwards notifications to predefined contacts, consisting of:
 
-* The Icinga Notifications daemon (this repository), which receives events and sends notifications
-* An [Icinga Web module](https://github.com/Icinga/icinga-notifications-web) that provides graphical configuration and further processing of the data collected by the daemon
-* And Icinga 2 and other custom sources that propagate state updates and acknowledgement events to the daemon
+* The Icinga Notifications daemon (this repository), which receives events and sends notifications.
+* The [Icinga Notifications Web](https://github.com/Icinga/icinga-notifications-web) module,
+  which provides graphical configuration.
+* Icinga 2 or other sources that provide monitoring events that result in notifications.
```
yhabteab, line 14: Icinga 2 is IMHO a must have, so `s/or/and/`. The same applies to the about page.

doc/01-About.md
```diff
@@ -0,0 +1,55 @@
...
+
+First, the Icinga Notifications configuration resides in a SQL database.
+It can be conveniently tweaked via Icinga Notifications Web directly from a web browser.
+The Icinga Notifications daemon uses this database to read the current configuration.
+
+As in any Icinga setup, all host and service checks are defined in Icinga 2.
+By querying the Icinga 2 API, the Icinga Notifications daemon retrieves state changes, acknowledgements, and other events.
+These events are stored in the database and are available for further inspection in the Icinga Notifications Web.
+
+Depending on its configuration, the daemon will take further action on these events.
```
yhabteab, line 32: further ~action~ actions on ..

doc/03-Configuration.md
```diff
@@ -0,0 +1,120 @@
...
+| debug-password | Password expected via HTTP Basic Authentication for debug endpoints. |
+
+### Icinga Web 2
+
+The `icingaweb2-url` is expected to point to the base directory of your Icinga Web 2 installation,
+i.e., `https://example.com/icingaweb2/`, to be used for URL creation.
+
+### Channels Directory
+
+All available Icinga Notification channels should reside in the `channels-dir` directory.
```
yhabteab, line 30: Icinga Notification**s**

doc/03-Configuration.md
```diff
@@ -0,0 +1,120 @@
...
+For a package installation, the default will point to the correct location and must not be changed.
+
+This directory should be `/usr/libexec/icinga-notifications/channels` on systems that follow the Filesystem Hierarchy Standard.
+It may also be `/usr/lib/icinga-notifications/channels`, depending on the operating system conventions.
+
+## Database Configuration
+
+Connection configuration for the database to which Icinga Notifications synchronizes monitoring data.
```
yhabteab, line 38: Does Icinga Notifications *synchronizes monitoring data*?

README.md
oxzi: Is it technically really a hard requirement? I mean, you might end in a hard place without a real host and service source..

But yes, valid point. It's all about Icinga anyways.

ncosta-ic: That's what I meant in our discussion the other day. The daemon itself is essentially "standalone" while still being branded with the Icinga ecosystem.
No, there's no hard-dependency for Icinga 2 for the daemon itself. For the web part on the other end, it's definitely a must, as Icinga Web won't run without it (or at least the login would fail).

doc/01-About.md
oxzi: I am quite certain that it should be _action_ in this case, as "take action" is a collocation referring to some generic action to be "taken care of" and not a specific action, or a multitude of actions.

https://dictionary.cambridge.org/dictionary/english/take-action

doc/03-Configuration.md
oxzi: Thanks. Now I even `grep`ped for it and this was hopefully the last of the singular ones.

doc/03-Configuration.md
oxzi: Oopsie.

julianbrost (dismissed):

oxzi:
Sorry, while testing the packages together with the docs, I found this faulty username.

doc/01-About.md
```diff
@@ -0,0 +1,55 @@
...
+Because Icinga Notifications consists of several components,
+this section tries to help understand how these components relate.
+
+First, the Icinga Notifications configuration resides in a SQL database.
+It can be conveniently tweaked via Icinga Notifications Web directly from a web browser.
+The Icinga Notifications daemon uses this database to read the current configuration.
+
+As in any Icinga setup, all host and service checks are defined in Icinga 2.
+By querying the Icinga 2 API, the Icinga Notifications daemon retrieves state changes, acknowledgements, and other events.
+These events are stored in the database and are available for further inspection in the Icinga Notifications Web.
```
nilmerg, line 30: ~~the~~ Icinga Notifications Web

nilmerg (changes requested):

README.md
nilmerg, line 1: The license description differs from 01-About.md

nilmerg (changes requested):

oxzi:
Next to @nilmerg's comments, I have added a sentence to the _Big Picture_ section mentioning other possible sources, as otherwise the text mismatched the image. Furthermore, I have changed the default PostgreSQL credentials in the example configuration to those of the documentation.

doc/01-About.md
yhabteab: Given that the demon doesn't just take a single additional action, but quite a few, the **action** somehow sounds wrong to my ears.

yhabteab (approved):
Looks perfect TM.

nilmerg (approved):

julianbrost (approved):

ncosta-ic:
This PR got merged while containing a faulty link.

While a relative link to the `LICENSE` file works fine for the `README.md`, it doesn't work for MKDocs (in this case `01-About.md`) as it references a file outside of its directory and that file does not get included into the HTML output.
This results in a broken link when building it through our Docker image and will thus also fail once added to [icinga.com/docs](https://icinga.com/docs).

[icinga-notifications-docs-bug.webm](https://github.com/Icinga/icinga-notifications/assets/146938323/493142b1-4252-41b9-a483-a0cb00454eb5)

oxzi:
> This PR got merged while containing a faulty link.
> 
> While a relative link to the `LICENSE` file works fine for the `README.md`, it doesn't work for MKDocs (in this case `01-About.md`) as it references a file outside of its directory and that file does not get included into the HTML output. This results in a broken link when building it through our Docker image and will thus also fail once added to [icinga.com/docs](https://icinga.com/docs).

Are you sure? The same happens in other documentations as well, e.g., for Icinga DB:

https://github.com/Icinga/icingadb/blob/6102d9cb2e30a628ceaf79cc4ba01db1ab50aa8f/doc/03-Configuration.md?plain=1#L4

On the rendered web documentation, the [resulting page](https://icinga.com/docs/icinga-db/latest/doc/03-Configuration/) links to `../../config.example.yml`. With its URL being `https://icinga.com/docs/icinga-db/latest/doc/03-Configuration/`, the destination[^1] still lies within the project's subdirectory.

However, I guess your point is still valid for the documentation's main page, as it might be accessed directly under `https://icinga.com/docs/icinga-notifications/` and not `…/icinga-notifications/doc/01-About/`. Good catch though.

[^1]: `https://icinga.com/docs/icinga-db/latest/doc/03-Configuration/../../config.example.yml = https://icinga.com/docs/icinga-db/latest/config.example.yml`

ncosta-ic:
> Are you sure? The same happens in other documentations as well, e.g., for Icinga DB
> ...
> However, I guess your point is still valid for the documentation's main page, as it might be accessed directly under `https://icinga.com/docs/icinga-notifications/` and not `…/icinga-notifications/doc/01-About/`. Good catch though.

That's correct. And yes, I was wrong about the includes.
I blindly checked my Nginx logs instead of just going through the generated html folder.
It actually ships the whole git repository and any relative links should work.

The second problem where a user could access the `01-About.md` directly under `.../latest` and not through `.../doc/01-About` still results in a broken link.
But you noticed that as well. Just wanted to clarify my thought process.

PR #211, opened 2024-06-04
Send systemd `READY=1` status message

yhabteab:
The default Icinga Notifications systemd service is going to have `Type=notify` in its unit file, so systemd expects us to send this message when the process becomes ready.

### DIFF
--- a/cmd/icinga-notifications-daemon/main.go
+++ b/cmd/icinga-notifications-daemon/main.go
@@ -14,6 +14,7 @@ import (
 	"github.com/icinga/icinga-notifications/internal/icinga2"
 	"github.com/icinga/icinga-notifications/internal/incident"
 	"github.com/icinga/icinga-notifications/internal/listener"
+	"github.com/okzk/sdnotify"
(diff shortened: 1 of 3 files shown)


oxzi (dismissed):
:shipit:

cmd/icinga-notifications-daemon/main.go
```diff
@@ -60,6 +61,9 @@ func main() {
 		utils.PrintErrorThenExit(err, 1)
 	}
 
+	// When Icinga Notifications is started by systemd, we've to notify systemd that we're ready.
+	_ = sdnotify.Ready()
+
 	logger := logs.GetLogger()
 	defer func() { _ = logger.Sync() }()
```
julianbrost, line 69: I think this is put quite early and provides no real benefit over not sending the ready notification and just using `Type=simple`. Typically, with systemd, services should be report readiness when they can accept requests. So for our case, this would probably be after the listening socket has been opened.

oxzi (approved):

PR #215, opened 2024-06-10
Refactor `browser_session` table

ncosta-ic:
This PR modifies the `browser_session` table to reflect new changes to the data structure.

- The primary key no longer depends on the `username` or `user_agent`, as the session storage hook takes care of zombie entries before inserting new session data.
- Added the additional descending order to the existing `authenticated_at` index.
- Fixed index naming convention.
- A new combined index has been added, which speeds up lookups that use both `username` and `user_agent` conditions.
- The default expression for `authenticated_at` got removed, as the timestamp now gets supplied by the PHP daemon itself.

Fixes https://github.com/Icinga/icinga-notifications-web/issues/194
Fixes https://github.com/Icinga/icinga-notifications-web/issues/195

### DIFF
--- a/schema/pgsql/schema.sql
+++ b/schema/pgsql/schema.sql
@@ -350,7 +350,7 @@ CREATE TABLE browser_session (
     php_session_id varchar(256) NOT NULL,
     username citext NOT NULL,
     user_agent varchar(4096) NOT NULL,
-    authenticated_at bigint NOT NULL DEFAULT (EXTRACT(EPOCH FROM CURRENT_TIMESTAMP) * 1000),
+    authenticated_at bigint NOT NULL,
 
     CONSTRAINT pk_browser_session PRIMARY KEY (php_session_id)
 );
--- a/schema/pgsql/schema.sql
+++ b/schema/pgsql/schema.sql
@@ -349,10 +349,11 @@ COMMENT ON INDEX idx_incident_history_time_type IS 'Incident History ordered by 
 CREATE TABLE browser_session (
     php_session_id varchar(256) NOT NULL,
     username citext NOT NULL,
-    user_agent text NOT NULL,
+    user_agent varchar(4096) NOT NULL,
     authenticated_at bigint NOT NULL DEFAULT (EXTRACT(EPOCH FROM CURRENT_TIMESTAMP) * 1000),
 
-    CONSTRAINT pk_browser_session PRIMARY KEY (php_session_id, username, user_agent)
+    CONSTRAINT pk_browser_session PRIMARY KEY (php_session_id)
 );
 
-CREATE INDEX browser_session_authenticated_at_idx ON browser_session (authenticated_at);
+CREATE INDEX idx_browser_session_authenticated_at ON browser_session (authenticated_at DESC);
+CREATE INDEX idx_browser_session_username_agent ON browser_session (username, user_agent);
--- a/schema/pgsql/upgrades/029.sql
+++ b/schema/pgsql/upgrades/029.sql
@@ -9,5 +9,8 @@ ALTER TABLE browser_session
 ALTER TABLE browser_session
     ADD CONSTRAINT pk_browser_session PRIMARY KEY (php_session_id);
 
+ALTER TABLE IF EXISTS browser_session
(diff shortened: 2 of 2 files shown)


julianbrost:
> * Reverted the usage of `citext` on the `username` property, as this idea has not been finally discussed (see: [schema: Require & use `citext` extension #71](https://github.com/Icinga/icinga-notifications/pull/71)).

Is that really due to that? Is there any remaining doubt that the username can be treated case-insensitive? To my current knowledge, there seems to be some inconsistency in Icinga Web there between MySQL/MariaDB and PostgreSQL, but neither allows you to have users whose names differ only in case.

ncosta-ic:
> > * Reverted the usage of `citext` on the `username` property, as this idea has not been finally discussed (see: [schema: Require & use `citext` extension #71](https://github.com/Icinga/icinga-notifications/pull/71)).
> 
> Is that really due to that? Is there any remaining doubt that the username can be treated case-insensitive? To my current knowledge, there seems to be some inconsistency in Icinga Web there between MySQL/MariaDB and PostgreSQL, but neither allows you to have users whose names differ only in case.

That's correct. We don't support different usernames solely based on their case sensitivity.
However, I feel the introduction of `citext` should take place in a separate PR once everything's discussed about it.
I'd rather not include in the `browser_session` if that's not our standard yet.

julianbrost:
#71 was merged earlier today so now there's a precedent for using `citext` to store the username.

schema/pgsql/upgrades/028.sql
julianbrost, line 1: I don't think the `IF EXISTS` do anything useful in this file. These would skip the statement if the table didn't exist. However, it should exist and if it didn't, the index creation at the end would fail anyways.

schema/pgsql/schema.sql
```diff
@@ -349,10 +349,11 @@ COMMENT ON INDEX idx_incident_history_time_type IS 'Incident History ordered by
...
-    authenticated_at bigint NOT NULL DEFAULT (EXTRACT(EPOCH FROM CURRENT_TIMESTAMP) * 1000),
+    user_agent varchar(4096) NOT NULL,
+    authenticated_at bigint NOT NULL,
 
-    CONSTRAINT pk_browser_session PRIMARY KEY (php_session_id, username, user_agent)
+    CONSTRAINT pk_browser_session PRIMARY KEY (php_session_id)
 );
 
-CREATE INDEX browser_session_authenticated_at_idx ON browser_session (authenticated_at);
+CREATE INDEX idx_browser_session_authenticated_at ON browser_session (authenticated_at DESC);
```
julianbrost, line 358: For any sane database, specifying `DESC` should have no relevant effect in this case (but it doesn't hurt either, so feel free to keep it if you think it makes the purpose of the index easier to understand). Specifying a sort order in an index is mostly relevant if you have multiple columns where it changes the order in the index besides just reversing it entirely.

julianbrost (changes requested):

schema/pgsql/schema.sql
ncosta-ic: This had more of a verbose meaning to it. We generally only want to read it in descending order as that's the single reason this index exists.

schema/pgsql/upgrades/028.sql
ncosta-ic: Better safe than sorry, but you're right

schema/pgsql/upgrades/028.sql
```diff
@@ -0,0 +1,16 @@
+ALTER TABLE IF EXISTS browser_session
+    DROP CONSTRAINT pk_browser_session;
+
+DROP INDEX browser_session_authenticated_at_idx;
+
+ALTER TABLE IF EXISTS browser_session
+    ALTER COLUMN user_agent TYPE varchar(4096) USING user_agent::varchar(4096);
+
+ALTER TABLE IF EXISTS browser_session
+    ADD CONSTRAINT pk_session PRIMARY KEY (php_session_id);
```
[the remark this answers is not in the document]
ncosta-ic: Not sure how that ended up in there. Going to fix it

julianbrost (approved):

PR #216, opened 2024-06-11
Add column `external_uuid` to `contact/contactgroup` table

sukhwinder33445:


schema/pgsql/upgrades/026.sql
julianbrost, line 1: A newline at the end of the file is missing.

schema/pgsql/upgrades/026.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE contact ADD COLUMN external_uuid uuid NOT NULL;
+ALTER TABLE contactgroup ADD COLUMN external_uuid uuid NOT NULL;
+
+ALTER TABLE contact ALTER COLUMN color DROP NOT NULL;
```
julianbrost, line 4: Removing that column is neither mentioned in the commit message, the PR description, nor the linked issue. If this made it into the PR intentionally, please elaborate.

julianbrost (changes requested):

schema/pgsql/upgrades/026.sql
sukhwinder33445: Ahh, there's already a [issue](https://github.com/Icinga/icinga-notifications-web/issues/200) for this, so I'm deleting this change here.

julianbrost (changes requested):
You're missing the change in the full `schema.sql` file (looks like I've totally missed this before).

Also, if the `NOT NULL` constraints remain,
1. a value should be set for existing rows.
2. this should not be merged without https://github.com/Icinga/icinga-notifications-web/pull/199, otherwise this would break creating contacts.

schema/pgsql/schema.sql
```diff
@@ -56,9 +58,11 @@ CREATE TABLE contact (
     full_name citext NOT NULL,
     username citext, -- reference to web user
     default_channel_id bigint NOT NULL REFERENCES channel(id),
+    external_uuid uuid NOT NULL,
 
-    CONSTRAINT pk_contact PRIMARY KEY (id),
-    UNIQUE (username)
+    UNIQUE (username),
+    UNIQUE (external_uuid),
+    CONSTRAINT pk_contact PRIMARY KEY (id)
```
julianbrost, line 65: I'm wondering why you reordered the constraints here. Also, in the rest of the schema file, so far the primary key is given first, I'd keep it that way (it also makes sense to me, one can say it's the most important constraint).

julianbrost (commented):
General question: do you (= team web) expect/want this to be merged soon?

sukhwinder33445:
This can wait, so far it is a stand-alone change without any dependencies to other branches (except the linked web branch). When the corresponding web branch is ready to be merged, this should be merged together with it.

nilmerg:
I've only rebased and fixed the conflicts.

schema/mysql/schema.sql
```diff
@@ -10,6 +10,7 @@ CREATE TABLE available_channel_type (
 
 CREATE TABLE channel (
     id bigint NOT NULL AUTO_INCREMENT,
+    external_uuid char(36) NOT NULL, -- used for external references, lower case
```
julianbrost: The channel table isn't mentioned in the PR title (or description). If this is intended, please update these as well.

julianbrost:
FYI: With these changes, the PR should also make it into a "ready to merge anytime the corresponding Notifications Web PR is ready", similar to https://github.com/Icinga/icinga-notifications/pull/344#pullrequestreview-3393370837.

PR #218, opened 2024-06-12
Contact/Contactgroup: Drop column `color`

sukhwinder33445:
ref https://github.com/Icinga/icinga-notifications-web/pull/203


julianbrost:
Please provide a reference to the corresponding change in Icinga Notifications Web. Due to the `not null` constraint being there so far, both probably have to be merged at the same time, otherwise it would break the functionality.

sukhwinder33445:
The web PR is ready to merge.

julianbrost (approved):

PR #219, opened 2024-06-14
icinga2: Enhance debug logging

yhabteab:
This commit implements the `zapcore.ObjectMarshaler` interface for all API / event stream response result types that we currently sprintf the entire Object with `#%v` and allows us to lazily encode all necessary fields without bloating the debug logs with useless information.

**Example output:**
```bash
2024-06-14T16:10:47.521+0200	DEBUG	icinga2	Skipping SOFT event	{"source-id": 1, "name": "passive", "host": "docker-master", "state": 0, "state_type": 1, "in_downtime": false, "acknowledged": false, "last_state_change": "2024-06-14T15:05:22.434+0200", "groups": [], "check_result": {"exit_status": 0, "state": 0, "output": "This is passive check too"}}
2024-06-14T16:10:47.523+0200	DEBUG	icinga2	Skipping SOFT event	{"source-id": 1, "name": "ping4", "host": "DHCP-server", "state": 0, "state_type": 1, "in_downtime": false, "acknowledged": false, "last_state_change": "2024-06-14T15:05:43.319+0200", "groups": ["ping"], "check_result": {"exit_status": 0, "state": 0, "output": "PING OK - Packet loss = 0%, RTA = 0.82 ms"}}
2024-06-14T16:10:47.527+0200	DEBUG	icinga2	Stopped processing event with superfluous state change	{"source-id": 1, "event": "[time=2024-06-14 16:10:47.523099 +0200 CEST m=+0.168237293 type=\"state\" severity=ok]", "error": "ignoring superfluous state change: ok state event from source 1"}
2024-06-14T16:10:47.527+0200	DEBUG	icinga2	Skipping SOFT event	{"source-id": 1, "name": "random fortune", "host": "DHCP-server", "state": 0, "state_type": 1, "in_downtime": false, "acknowledged": false, "last_state_change": "2024-06-14T15:05:30.854+0200", "groups": ["app-db", "department-nms", "env-acceptance", "location-berlin"], "check_result": {"exit_status": 0, "state": 0, "output": "Your temporary financial embarrassment will be relieved in a surprising manner."}}
```

### DIFF
--- a/internal/icinga2/api_responses.go
+++ b/internal/icinga2/api_responses.go
@@ -3,6 +3,7 @@ package icinga2
 import (
 	"encoding/json"
 	"fmt"
+	"go.uber.org/zap/zapcore"
 	"strconv"
 	"time"
 )
@@ -80,6 +81,14 @@ type CheckResult struct {
 	ExecutionEnd   UnixFloat `json:"execution_end"`
 }
 
+// MarshalLogObject implements the zapcore.ObjectMarshaler interface.
+func (cr *CheckResult) MarshalLogObject(encoder zapcore.ObjectEncoder) error {
+	encoder.AddInt("exit_status", cr.ExitStatus)
+	encoder.AddInt("state", cr.State)
+	encoder.AddString("output", cr.Output)
+	return nil
+}
+
 // Downtime represents the Icinga 2 API Downtime object.
 //
 // NOTE:
@@ -104,6 +113,21 @@ type Downtime struct {
 	// for the very (same) event. Flexible downtimes, on the other hand, only emits a trigger event, and
 	// don't produce duplicates for the same event.
 	IsFixed bool `json:"fixed"`
+}
+
+// MarshalLogObject implements the zapcore.ObjectMarshaler interface.
+func (d *Downtime) MarshalLogObject(encoder zapcore.ObjectEncoder) error {
+	encoder.AddString("host", d.Host)
+	if d.Service != "" {
+		encoder.AddString("service", d.Service)
+	}
+
+	encoder.AddString("author", d.Author)
+	encoder.AddBool("fixed", d.IsFixed)
+	encoder.AddBool("was_cancelled", d.WasCancelled())
+	encoder.AddTime("remove_time", d.RemoveTime.Time())
+	encoder.AddString("comment", d.Comment)
+	return nil
 }
 
 // WasCancelled returns true when the current downtime was cancelled prematurely by a user.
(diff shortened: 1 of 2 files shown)


internal/icinga2/api_responses.go
```diff
@@ -103,6 +112,19 @@ type Downtime struct {
...
+// MarshalLogObject implements the zapcore.ObjectMarshaler interface.
+func (d *Downtime) MarshalLogObject(encoder zapcore.ObjectEncoder) error {
+	encoder.AddString("host", d.Host)
+	if d.Service != "" {
+		encoder.AddString("service", d.Service)
+	}
+
+	encoder.AddString("author", d.Author)
+	encoder.AddBool("fixed", d.IsFixed)
+	encoder.AddString("comment", d.Comment)
```
oxzi, line 124: Maybe keep this in the same order as the struct fields? Also, why did you exclude `RemoveTime`?

internal/icinga2/api_responses.go
```diff
@@ -135,6 +157,31 @@ type HostServiceRuntimeAttributes struct {
...
+		for _, group := range hsra.Groups {
+			encoder.AppendString(group)
+		}
+		return nil
+	}))
+	if err != nil {
+		return err
+	}
+
+	return encoder.AddObject("check_result", &hsra.LastCheckResult)
```
oxzi, line 182: Again, could you try to keep the same order as for the struct fields? And why was only `AcknowledgementLastChange` excluded?

internal/icinga2/api_responses.go
```diff
@@ -178,6 +225,21 @@ type StateChange struct {
...
+	if sc.Service != "" {
+		encoder.AddString("service", sc.Service)
+	}
+
+	encoder.AddInt("state", sc.State)
+	encoder.AddInt("state_type", sc.StateType)
+	encoder.AddBool("in_downtime", sc.DowntimeDepth != 0)
+	encoder.AddBool("acknowledged", sc.Acknowledgement)
+	return encoder.AddObject("check_result", &sc.CheckResult)
+}
```
oxzi, line 241: Order again.

internal/icinga2/api_responses.go
yhabteab: > Again, could you try to keep the same order as for the struct fields?

Same thing here too! The check result is a separate json object, and I was just trying to reorder things so they can be read easily, just like you wouldn't want to add the `zap.Error()` field before something else, but always as the last one.

> And why was only AcknowledgementLastChange excluded?

Same here as well! I don't see any use case where this could be helpful for troubleshooting.

internal/icinga2/api_responses.go
yhabteab: Same here as well :).

internal/icinga2/api_responses.go
oxzi: > > Maybe keep this in the same order as the struct fields?
> 
> I actually did this on purpose, as you can have huge downtime comments and the more important with far too short and relevant values should come first IMHO.

Makes sense. It just felt a bit.. "untidy" for me :upside_down_face:

> > Also, why did you exclude RemoveTime?
> 
> I just didn't want to add too much information that isn't needed for debugging in anyway, or do you have use cases where this might be relevant? The `was_cancelled` key should be added though.

I also see your point here. No, at the moment I don't have a need for this field, but this might be in the future and then one either misses the information due to its unavailability or has to patch this again.

internal/icinga2/api_responses.go
oxzi: For me its also the same again :D

internal/icinga2/api_responses.go
```diff
@@ -80,6 +81,14 @@ type CheckResult struct {
 	ExecutionEnd   UnixFloat `json:"execution_end"`
 }
 
+// MarshalLogObject implements the zapcore.ObjectMarshaler interface.
+func (cr *CheckResult) MarshalLogObject(encoder zapcore.ObjectEncoder) error {
+	encoder.AddInt("exit_status", cr.ExitStatus)
+	encoder.AddInt("state", cr.State)
+	encoder.AddString("output", cr.Output)
```
oxzi, line 88: The output might be very long and might even contain sensitive data, that should not be logged. Otherwise, without this field the log message might be next to useless.

Thus, as long as it is only used for debug logging, this should be fine. And it is currently being logged anyway.

The same applies for the other types.

oxzi (approved):

PR #222, opened 2024-06-21
Don't retry DB errors for `5m` on daemon startup

yhabteab:
See the inline comments for more context.

### DIFF
--- a/cmd/icinga-notifications/main.go
+++ b/cmd/icinga-notifications/main.go
@@ -70,8 +70,19 @@ func main() {
 	defer cancel()
 
 	logger.Infof("Connecting to database at '%s'", db.GetAddr())
-	if err := db.PingContext(ctx); err != nil {
-		logger.Fatalf("Cannot connect to the database: %+v", err)
+	{
+		// When installing Icinga Notifications via our packages, the systemd service will be started automatically
+		// e.g. on Debian based systems, and since the unit type is set to `notify`, make sure you don't block the
+		// installation if no database is configured yet.
+		dbCtx, cancelFunc := context.WithTimeout(ctx, 5*time.Second)
+		defer cancelFunc()
+		if err := db.PingContext(dbCtx); err != nil {
+			logger.Fatalf("Cannot connect to the database: %+v", err)
+		}
+
+		// The deferred call above is going to be invoked when main exits, but we want to release the resources now.
+		// So call it here and the deferred one becomes a no-op.
+		cancelFunc()
 	}
 
 	channel.UpsertPlugins(ctx, conf.ChannelPluginDir, logs.GetChildLogger("channel"), db)


cmd/icinga-notifications/main.go
```diff
@@ -70,8 +70,19 @@ func main() {
...
 
 	logger.Infof("Connecting to database at '%s'", db.GetAddr())
-	if err := db.PingContext(ctx); err != nil {
-		logger.Fatalf("Cannot connect to the database: %+v", err)
+	{
+		// When installing Icinga Notifications via our packages, the systemd service will be started automatically
+		// e.g. on Debian based systems, and since the unit type is set to `notify`, make sure you don't block the
+		// installation if no database is configured yet.
+		dbCtx, cancelFunc := context.WithTimeout(ctx, 5*time.Second)
+		defer cancelFunc()
```
oxzi, line 78: This `defer` seems useless. In case of a `logger.Fatalf`, the logger exits the program with a call to `os.Exit` and otherwise, `cancelFunc` will be called directly.

julianbrost:
> When installing Icinga Notifications via our packages, the systemd service will be started automatically e.g. on Debian based systems, and since the unit type is set to `notify`, make sure you don't block the installation if no database is configured yet.

This change somewhat defeats the purpose of the retry mechanism. It exists in part to account for delayed database startup.

Even with this change, the service wouldn't start properly, it would just fail earlier. So wouldn't this still result in an error on package installation?

What is the point in automatically starting a service if it can't work without configuring it first? Not sure what Debian's packaging policies say about this but to me not automatically starting a service sounds like the only reasonable thing in that situation.

yhabteab:
> Even with this change, the service wouldn't start properly, it would just fail earlier. So wouldn't this still result in an error on package installation?

I don't think so, yes, of course the service wouldn't actually start properly, but it shouldn't interfere with the package installation in any way. Btw, this is how Debian starts the service in the end `deb-systemd-invoke start 'icinga-notifications.service' >/dev/null || true`.

> What is the point in automatically starting a service if it can't work without configuring it first? Not sure what Debian's packaging policies say about this but to me not automatically starting a service sounds like the only reasonable thing in that situation.

Well, I don't think Debian will automatically start your systemd service out of nowhere, but only if you include this small but magical comment `#DEBHELPER#` in your `postinst` script.

julianbrost:
> > Even with this change, the service wouldn't start properly, it would just fail earlier. So wouldn't this still result in an error on package installation?
> 
> I don't think so, yes, of course the service wouldn't actually start properly, but it shouldn't interfere with the package installation in any way. Btw, this is how Debian starts the service in the end `deb-systemd-invoke start 'icinga-notifications.service' >/dev/null || true`.

Depends on how you define interfere. It would still block it for 5 seconds. Also, there's still stderr, so there might still be an error message. And it will leave behind a failed unit, so `systemctl status` will show the system to be in a degraded state.

> > What is the point in automatically starting a service if it can't work without configuring it first? Not sure what Debian's packaging policies say about this but to me not automatically starting a service sounds like the only reasonable thing in that situation.
> 
> Well, I don't think Debian will automatically start your systemd service out of nowhere, but only if you include this small but magical comment `#DEBHELPER#` in your `postinst` script.

Well yes and no. If you omit `#DEBHELPER#` from a custom `postinst` script, the auto-generated commands won't make it to the final package. However, if there was no custom `postinst`, it should still add one with just the auto-generated commands. You can tweak the behavior of what debhelper generates, see [`dh_installsystemd`](https://manpages.debian.org/bookworm/debhelper/dh_installsystemd.1.en.html).

yhabteab:
Fixed by adding `dh_installsystemd --no-start` in the Debian rules script.

PR #191, opened 2024-04-30
Incremental Configuration Updates

oxzi:
Previously, the entire configuration stored in the database was synchronized every second. With the growth of configurations in live environments on the horizon, this would simply not scale well. This brings us to incremental updates.

By introducing two new columns - "changed_at" as a Unix millisecond timestamp and "deleted" as a boolean - for all tables referenced in the ConfigSet structure, SQL queries can be modified to retrieve only those rows with a more recent timestamp. The "deleted" column became necessary to detect disappearances, since the synchronization now only takes newer items into account. Some additional fields needed to be added to the ConfigSet to track relationships.

Even though the codebase served well at the time, there was some code that did almost the same thing as other code, just in different ways.  So a huge refactoring was done. This resulted in an internal generic function that handles all synchronization with custom callbacks.

The web counterpart is being developed in <https://github.com/Icinga/icinga-notifications-web/pull/187>.

Closes #5.

PR #226, opened 2024-07-02
Schema: name all constraints, not just PKs

### LINKED ISSUE #202: Schema: name all constraints, not just PKs (before v1.0)

Al2Klimov:
fixes #202


Al2Klimov:
Fresh schema dump matches upgraded schema dump.

incident_contact's CONSTRAINTs are already named properly.

schema/pgsql/schema.sql
```diff
@@ -112,51 +120,60 @@ CREATE TABLE rotation (
...
     -- basically ignored in the UNIQUE constraint over contactgroup_id and vice versa. The CHECK constraint below
     -- ensures that each row has only non-NULL values in one of these constraints.
-    UNIQUE (rotation_id, contact_id),
-    UNIQUE (rotation_id, contactgroup_id),
-    CHECK (num_nonnulls(contact_id, contactgroup_id) = 1),
-
-    CONSTRAINT pk_rotation_member PRIMARY KEY (id)
+    CONSTRAINT key_rotation_member_rotation_contact UNIQUE (rotation_id, contact_id),
+    CONSTRAINT key_rotation_member_rotation_contactgroup UNIQUE (rotation_id, contactgroup_id),
+    CONSTRAINT nonnulls_rotation_member_recipients_check CHECK (num_nonnulls(contact_id, contactgroup_id) = 1),
```
julianbrost, line 155: That's not a good name choice for keeping the names consistent with MySQL/MariaDB as there's no such function there. Maybe something like `c(hec)k_one_of_contact_or_contactgroup` or `c(hec)k_one_recipient`.

schema/pgsql/schema.sql
```diff
@@ -187,40 +204,42 @@ CREATE TABLE source (
     -- The hash is a PHP password_hash with PASSWORD_DEFAULT algorithm, defaulting to bcrypt. This check roughly ensures
     -- that listener_password_hash can only be populated with bcrypt hashes.
     -- https://icinga.com/docs/icinga-web/latest/doc/20-Advanced-Topics/#manual-user-creation-for-database-authentication-backend
-    CHECK (listener_password_hash IS NULL OR listener_password_hash LIKE '$2y$%'),
-    CHECK (type != 'icinga2' OR (icinga2_base_url IS NOT NULL AND icinga2_auth_user IS NOT NULL AND icinga2_auth_pass IS NOT NULL)),
+    CONSTRAINT bcrypt_source_listener_password_hash_check CHECK (listener_password_hash IS NULL OR listener_password_hash LIKE '$2y$%'),
+    CONSTRAINT nonnulls_source_icinga2_check CHECK (type != 'icinga2' OR (icinga2_base_url IS NOT NULL AND icinga2_auth_user IS NOT NULL AND icinga2_auth_pass IS NOT NULL)),
```
julianbrost, line 208: That's a different check from the other `num_nonnulls` checks. I'd suggest a name like `c(hec)k_source_icinga2_has_config`.

schema/pgsql/schema.sql
```diff
@@ -187,40 +204,42 @@ CREATE TABLE source (
...
 
     url text,
     -- mute_reason indicates whether an object is currently muted by its source, and its non-zero value is mapped to true.
     mute_reason text,
 
-    CHECK (length(id) = 256/8),
-
-    CONSTRAINT pk_object PRIMARY KEY (id)
+    CONSTRAINT pk_object PRIMARY KEY (id),
+    CONSTRAINT length_id32_check CHECK (length(id) = 256/8),
```
julianbrost, line 223: I'd suggest `c(hec)k_object_id_is_sha256`.

schema/pgsql/schema.sql
```diff
@@ -242,111 +261,129 @@ CREATE TYPE severity AS ENUM ('ok', 'debug', 'info', 'notice', 'warning', 'err',
...
     name citext, -- if not set, recipients are used as a fallback for display purposes
-    fallback_for bigint REFERENCES rule_escalation(id),
+    fallback_for bigint,
 
     CONSTRAINT pk_rule_escalation PRIMARY KEY (id),
-
-    UNIQUE (rule_id, position),
-    CHECK (NOT (condition IS NOT NULL AND fallback_for IS NOT NULL))
+    CONSTRAINT key_rule_escalation_rule_position UNIQUE (rule_id, position),
+    CONSTRAINT onenull_rule_escalation_condition_fallback_for_check CHECK (NOT (condition IS NOT NULL AND fallback_for IS NOT NULL)),
```
julianbrost, line 297: That will probably already be named in #191 due to a second `check` being added. Current name there is `ck_rule_escalation_fallback_has_no_fallback`.

julianbrost (dismissed):
- You use a prefix for everything expect for checks where you use a `_check` suffix. #191 started using `ck_` as a prefix, it's not yet merged, but it should stay consistent, so I'd prefer a prefix, I'd be fine with both `ck_*` and `check_*`.
- `key_` doesn't sound like a good prefix for unique constraints to me. Primary and foreign keys are also keys and in MySQL/MariaDB, key actually means index (from https://dev.mysql.com/doc/refman/8.4/en/create-table.html: "KEY is normally a synonym for INDEX."). If we want to keep everything abbreviated, `uk_*` (unique key) might be an option, otherwise simply `unique_*` would also do.

schema/pgsql/upgrades/032.sql
```diff
@@ -0,0 +1,61 @@
...
+ALTER TABLE incident_contact RENAME CONSTRAINT incident_contact_incident_id_fkey TO fk_incident_contact_incident;
+ALTER TABLE incident_contact RENAME CONSTRAINT incident_contact_contact_id_fkey TO fk_incident_contact_contact;
+ALTER TABLE incident_contact RENAME CONSTRAINT incident_contact_contactgroup_id_fkey TO fk_incident_contact_contactgroup;
+ALTER TABLE incident_contact RENAME CONSTRAINT incident_contact_schedule_id_fkey TO fk_incident_contact_schedule;
+ALTER TABLE incident_rule RENAME CONSTRAINT incident_rule_incident_id_fkey TO fk_incident_rule_incident;
+ALTER TABLE incident_rule RENAME CONSTRAINT incident_rule_rule_id_fkey TO fk_incident_rule_rule;
+ALTER TABLE incident_rule_escalation_state RENAME CONSTRAINT incident_rule_escalation_state_incident_id_fkey TO fk_incident_rule_escalation_state_incident;
+ALTER TABLE incident_rule_escalation_state RENAME CONSTRAINT incident_rule_escalation_state_rule_escalation_id_fkey TO fk_incident_rule_escalation_state_rule_escalation;
+ALTER TABLE incident_history RENAME CONSTRAINT incident_history_incident_id_fkey1 TO fk_incident_history_incident_rule_escalation_state;
```
julianbrost, line 53: Interesting, which PostgreSQL version did you use for testing? For me (with PostgreSQL 16.3), the upgrade fails here due to the automatically created constraint being named `incident_history_incident_id_rule_escalation_id_fkey` instead of `incident_history_incident_id_fkey1`. Have we accidentally found a first example of the auto-generated names changing between versions?

Should be easy to work around by creating the new constraint and removing both old names with an additional `IF EXISTS`.

Al2Klimov: God bless my testing setup with PostgreSQL 9.6 and MySQL 5.5 (as well as Boost 1.66 and OpenSSL 1.0.2k). But, jokes aside: I hope this accident pointed out how serious this issue always had been.

## TODO

* [ ] resolve conflicts (@Al2Klimov, shouldn't take very long)

schema/pgsql/schema.sql
```diff
@@ -146,61 +152,67 @@ CREATE TABLE rotation (
     changed_at bigint NOT NULL,
     deleted boolenum NOT NULL DEFAULT 'n',
 
+    CONSTRAINT pk_rotation PRIMARY KEY (id),
+
     -- Each schedule can only have one rotation with a given priority starting at a given date.
     -- Columns schedule_id, priority, first_handoff must be NULLed for deletion via "deleted = 'y'".
-    UNIQUE (schedule_id, priority, first_handoff),
-    CHECK (deleted = 'y' OR priority IS NOT NULL AND first_handoff IS NOT NULL),
+    CONSTRAINT uk_rotation_schedule_schedule_priority_first_handoff UNIQUE (schedule_id, priority, first_handoff),
```
julianbrost, line 159: I don't see where the second "schedule" in "uk_rotation_**schedule_schedule**_priority_first_handoff" is coming from.

schema/pgsql/schema.sql
```diff
@@ -146,61 +152,67 @@ CREATE TABLE rotation (
...
+    CONSTRAINT ck_rotation_deleted_priority_first_handoff CHECK (deleted = 'y' OR priority IS NOT NULL AND first_handoff IS NOT NULL),
```
julianbrost, line 160: This is a very similar constraint like `ck_rotation_member_non_deleted_needs_position` and thus should be named similarly: `ck_rotation_non_deleted_needs_priority_first_handoff`.

schema/pgsql/schema.sql
```diff
@@ -247,8 +261,8 @@ CREATE TABLE source (
     -- The hash is a PHP password_hash with PASSWORD_DEFAULT algorithm, defaulting to bcrypt. This check roughly ensures
     -- that listener_password_hash can only be populated with bcrypt hashes.
     -- https://icinga.com/docs/icinga-web/latest/doc/20-Advanced-Topics/#manual-user-creation-for-database-authentication-backend
-    CHECK (listener_password_hash IS NULL OR listener_password_hash LIKE '$2y$%'),
-    CHECK (type != 'icinga2' OR (icinga2_base_url IS NOT NULL AND icinga2_auth_user IS NOT NULL AND icinga2_auth_pass IS NOT NULL)),
+    CONSTRAINT ck_bcrypt_source_listener_password_hash CHECK (listener_password_hash IS NULL OR listener_password_hash LIKE '$2y$%'),
```
julianbrost, line 264: This name seems to be an exception as it's not named `ck_${table}_...`  like the others.

schema/pgsql/schema.sql
yhabteab: This would probably also better describe what this check is used for, and of course be more consistent with the existing check names.

schema/pgsql/schema.sql
```diff
@@ -344,89 +362,106 @@ CREATE TABLE rule_escalation (
...
+    schedule_id bigint,
     role incident_contact_role NOT NULL,
 
-    CONSTRAINT key_incident_contact_contact UNIQUE (incident_id, contact_id),
-    CONSTRAINT key_incident_contact_contactgroup UNIQUE (incident_id, contactgroup_id),
-    CONSTRAINT key_incident_contact_schedule UNIQUE (incident_id, schedule_id),
-    CONSTRAINT nonnulls_incident_recipients_check CHECK (num_nonnulls(contact_id, contactgroup_id, schedule_id) = 1)
+    CONSTRAINT uk_incident_contact_contact UNIQUE (incident_id, contact_id),
+    CONSTRAINT uk_incident_contact_contactgroup UNIQUE (incident_id, contactgroup_id),
+    CONSTRAINT uk_incident_contact_schedule UNIQUE (incident_id, schedule_id),
```
julianbrost, line 428: These three constraints are referenced in the Go code and have to be changed there as well: https://github.com/Icinga/icinga-notifications/blob/860ea0ce3b74a0e8e463b3c2eccbe78fb7b58f62/internal/incident/db_types.go#L43-L54

schema/pgsql/schema.sql
oxzi: Descriptive names are already being used elsewhere, for example in:
https://github.com/Icinga/icinga-notifications/blob/3fd457278684894214a97d98a4ac1a406748a3a0/schema/pgsql/schema.sql#L202

schema/pgsql/schema.sql
```diff
@@ -344,89 +362,106 @@ CREATE TABLE rule_escalation (
 
     -- Each position in an escalation can only be used once.
     -- Column position must be NULLed for deletion via "deleted = 'y'"
-    UNIQUE (rule_id, position),
+    CONSTRAINT uk_rule_escalation_rule_position UNIQUE (rule_id, position),
```
oxzi, line 365:
```suggestion
    CONSTRAINT uk_rule_escalation_rule_id_position UNIQUE (rule_id, position),
```

schema/pgsql/schema.sql
```diff
@@ -344,89 +362,106 @@ CREATE TABLE rule_escalation (
...
+    schedule_id bigint,
+    channel_id bigint,
 
     changed_at bigint NOT NULL,
     deleted boolenum NOT NULL DEFAULT 'n',
 
     CONSTRAINT pk_rule_escalation_recipient PRIMARY KEY (id),
-
-    CHECK (num_nonnulls(contact_id, contactgroup_id, schedule_id) = 1)
+    CONSTRAINT ck_rule_escalation_recipient_one_recipient CHECK (num_nonnulls(contact_id, contactgroup_id, schedule_id) = 1),
```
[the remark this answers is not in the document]
oxzi: Maybe exactly instead of only?

schema/pgsql/schema.sql
oxzi: Maybe adding a comment in the schema to remind one about this would be helpful for possible future changes.

Al2Klimov: Also, I assume you guys are aware of MySQL not having an equivalent. Hence, it will update whenever ANY constraint conflicts.

schema/pgsql/schema.sql
```diff
@@ -146,61 +152,67 @@ CREATE TABLE rotation (
...
+    contactgroup_id bigint,
     position integer,
 
     changed_at bigint NOT NULL,
     deleted boolenum NOT NULL DEFAULT 'n',
 
     -- Each position in a rotation can only be used once.
     -- Column position must be NULLed for deletion via "deleted = 'y'".
-    UNIQUE (rotation_id, position),
+    CONSTRAINT uk_rotation_member_rotation_position UNIQUE (rotation_id, position),
```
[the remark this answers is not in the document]
yhabteab: I'm not sure what this is supposed to mean, but unique constraint names should be named after the `table` name and columns that are part of the index, not some random names like this `uk_rotation_member_rotation_position`.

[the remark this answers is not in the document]
Al2Klimov: "Random" or not, my names are descriptive, uniform and not-overengineered.

schema/pgsql/schema.sql
julianbrost: You mean using a specific unique constraint in `ON CONFLICT`? We don't that because we need it, but because we [have to](https://www.postgresql.org/docs/current/sql-insert.html#SQL-ON-CONFLICT):

> For `ON CONFLICT DO UPDATE`, a *`conflict_target`* *must* be provided.

The code would be happy with an "on any conflict", we're just lucky to be able to predict which one would produce the conflict (due to the `num_nonnulls` check).

julianbrost: Oh and regarding adding a comment: I think that totally makes sense and has nothing to do with expected further changes. It wouldn't have been a bad idea adding one right from the beginning, probably nobody was expecting it to change at that time either.

schema/pgsql/schema.sql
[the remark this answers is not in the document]
yhabteab: Again, I don't know what you are implying by *not-overengineered* but **you** wanted to add these names in the first place and I'm suggesting you to name them consistently and not over-engineer them.

schema/pgsql/schema.sql
oxzi: > Which future changes? The only reason for _this_ constant renaming is the (OP) issue that they're not named at all. And even that took me lots of persuading.

Well, I don't know which future change will come. That's the thing about the future, right?

In the current situation, however, I would not have expected some constraint to be referred in the code. Thus, adding a one line comment might safe an hour of debugging in the future.

schema/pgsql/schema.sql
```diff
@@ -344,89 +362,108 @@ CREATE TABLE rule_escalation (
...
+    contact_id bigint,
+    contactgroup_id bigint,
+    schedule_id bigint,
     role incident_contact_role NOT NULL,
 
-    CONSTRAINT key_incident_contact_contact UNIQUE (incident_id, contact_id),
-    CONSTRAINT key_incident_contact_contactgroup UNIQUE (incident_id, contactgroup_id),
-    CONSTRAINT key_incident_contact_schedule UNIQUE (incident_id, schedule_id),
-    CONSTRAINT nonnulls_incident_recipients_check CHECK (num_nonnulls(contact_id, contactgroup_id, schedule_id) = 1)
+    -- Keep in sync with internal/incident/db_types.go!
```
yhabteab, line 426: Added that little hint here but forgot to adjust the references in the GO code :).

julianbrost (commented):
Looks fine for me now from looking at the changes. If the others are happy as well, mark their comments as resolved, I'd apply it to my test setup, spin it up for a quick test and give an actual approval.

schema/pgsql/schema.sql
```diff
@@ -344,89 +362,108 @@ CREATE TABLE rule_escalation (
...
+    schedule_id bigint,
+    channel_id bigint,
 
     changed_at bigint NOT NULL,
     deleted boolenum NOT NULL DEFAULT 'n',
 
     CONSTRAINT pk_rule_escalation_recipient PRIMARY KEY (id),
-
-    CHECK (num_nonnulls(contact_id, contactgroup_id, schedule_id) = 1)
+    CONSTRAINT ck_rule_escalation_recipient_only_one_recipient CHECK (num_nonnulls(contact_id, contactgroup_id, schedule_id) = 1),
```
yhabteab, line 387: Sorry, but there's still a verb missing here and maybe you want to replace `only` with `exactly` like Alvar suggested. `ck_rule_escalation_recipient_has_exactly_one_recipient`

yhabteab (approved):
LFTM!

julianbrost (approved):

PR #230, opened 2024-07-08
Ignore flapping state if flapping detection isn't enabled

yhabteab:
The Icinga 2 Checkable `flapping` runtime attribute is always updated even when flapping detection is disabled and leads to false positives, causing Icinga Notifications to falsely suppress an incident. To prevent this, this PR additionally checks for the `enable_flapping` attribute of a checkable and globally for the `IcingaApplication` type.

And YES, I can safely use `app` as a json key because of this.
```bash
[2024-07-08 11:20:18 +0200] critical/config: Error: Validation failed for object 'my-app' of type 'IcingaApplication'; Attribute 'name': Application object must be named 'app'.
Location: in /Users/yhabteab/Workspace/icinga2/prefix/etc/icinga2/features-enabled/app.conf: 1:0-1:32
/Users/yhabteab/Workspace/icinga2/prefix/etc/icinga2/features-enabled/app.conf(1): object IcingaApplication "my-app" {}
                                                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/Users/yhabteab/Workspace/icinga2/prefix/etc/icinga2/features-enabled/app.conf(2): 
/Users/yhabteab/Workspace/icinga2/prefix/etc/icinga2/features-enabled/app.conf(3):
```

### DIFF
--- a/internal/icinga2/api_responses.go
+++ b/internal/icinga2/api_responses.go
@@ -161,6 +161,7 @@ type HostServiceRuntimeAttributes struct {
 	Acknowledgement           int         `json:"acknowledgement"`
 	IsFlapping                bool        `json:"flapping"`
 	AcknowledgementLastChange UnixFloat   `json:"acknowledgement_last_change"`
+	EnableFlapping            bool        `json:"enable_flapping"`
 }
 
 // MarshalLogObject implements the zapcore.ObjectMarshaler interface.
@@ -350,6 +351,14 @@ type ObjectCreatedDeleted struct {
 	ObjectName string `json:"object_name"`
 	ObjectType string `json:"object_type"`
 	EventType  string `json:"type"`
+}
+
+// IcingaApplication represents the Icinga 2 API status endpoint query result of type IcingaApplication.
+// https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#status-and-statistics
+type IcingaApplication struct {
+	App struct {
+		EnableFlapping bool `json:"enable_flapping"`
+	} `json:"app"`
 }
 
 // UnmarshalEventStreamResponse unmarshal a JSON response line from the Icinga 2 API Event Stream.
--- a/internal/icinga2/client_api.go
+++ b/internal/icinga2/client_api.go
@@ -149,6 +149,44 @@ func (client *Client) queryObjectsApiQuery(ctx context.Context, objType string, 
 		})
 }
 
+// fetchIcingaAppStatus retrieves the global state of the IcingaApplication type via the /v1/status endpoint.
+func (client *Client) fetchIcingaAppStatus(ctx context.Context) (*IcingaApplication, error) {
+	response, err := client.queryObjectsApi(
+		ctx,
+		[]string{"/v1/status/IcingaApplication/"},
+		http.MethodGet,
+		nil,
+		map[string]string{"Accept": "application/json"})
+	if err != nil {
+		return nil, err
+	}
+
+	defer func() {
+		_, _ = io.Copy(io.Discard, response)
+		_ = response.Close()
+	}()
+
+	type status struct {
+		Status struct {
+			IcingaApplication *IcingaApplication `json:"icingaapplication"`
+		} `json:"status"`
+	}
+
+	var results []status
+	err = json.NewDecoder(response).Decode(&struct {
+		Results *[]status `json:"results"`
+	}{&results})
+	if err != nil {
+		return nil, err
+	}
+
+	if len(results) == 0 {
+		return nil, fmt.Errorf("unable to fetch IcingaApplication status")
+	}
+
+	return results[0].Status.IcingaApplication, nil
+}
+
 // fetchCheckable fetches the Checkable config state of the given Host/Service name from the Icinga 2 API.
 func (client *Client) fetchCheckable(ctx context.Context, host, service string) (*ObjectQueriesResult[HostServiceRuntimeAttributes], error) {
 	objType, objName := "host", host
@@ -260,8 +298,13 @@ func (client *Client) checkMissedChanges(ctx context.Context, objType string, ca
 		}
 
 		attrs := objQueriesResult.Attrs
+		checkableIsMuted, err := isMuted(ctx, client, &objQueriesResult)
+		if err != nil {
+			return err
+		}
+
 		var fakeEv *event.Event
-		if attrs.Acknowledgement != AcknowledgementNone {
+		if checkableIsMuted && attrs.Acknowledgement != AcknowledgementNone {
 			ackComment, err := client.fetchAcknowledgementComment(ctx, hostName, serviceName, attrs.AcknowledgementLastChange.Time())
 			if err != nil {
 				return fmt.Errorf("fetching acknowledgement comment for %q failed, %w", objectName, err)
@@ -275,17 +318,17 @@ func (client *Client) checkMissedChanges(ctx context.Context, objType string, ca
 			if err != nil {
 				return fmt.Errorf("failed to construct Event from Acknowledgement response, %w", err)
 			}
-		} else if isMuted(&objQueriesResult) {
+		} else if checkableIsMuted {
 			fakeEv, err = client.buildCommonEvent(ctx, hostName, serviceName)
 			if err != nil {
 				return fmt.Errorf("failed to construct checkable fake mute event: %w", err)
 			}
 
 			fakeEv.Type = event.TypeMute
-			if attrs.IsFlapping {
-				fakeEv.SetMute(true, "Checkable is flapping, but we missed the Icinga 2 FlappingStart event")
-			} else {
+			if attrs.DowntimeDepth != 0 {
 				fakeEv.SetMute(true, "Checkable is in downtime, but we missed the Icinga 2 DowntimeStart event")
+			} else {
+				fakeEv.SetMute(true, "Checkable is flapping, but we missed the Icinga 2 FlappingStart event")
 			}
 		} else {
 			// This could potentially produce numerous superfluous database (event table) entries if we generate such
--- a/internal/icinga2/util.go
+++ b/internal/icinga2/util.go
@@ -1,6 +1,7 @@
 package icinga2
 
 import (
+	"context"
 	"net/url"
 	"strings"
 )
@@ -18,6 +19,24 @@ func rawurlencode(s string) string {
 }
 
 // isMuted returns true if the given checkable is either in Downtime, Flapping or acknowledged, otherwise false.
-func isMuted(checkable *ObjectQueriesResult[HostServiceRuntimeAttributes]) bool {
-	return checkable.Attrs.IsFlapping || checkable.Attrs.Acknowledgement != AcknowledgementNone || checkable.Attrs.DowntimeDepth != 0
+//
+// When the checkable is Flapping, and neither the flapping detection for that Checkable nor for the entire zone is
+// enabled, this will always return false.
+//
+// Returns an error if it fails to query the status of IcingaApplication from the /v1/status endpoint.
+func isMuted(ctx context.Context, client *Client, checkable *ObjectQueriesResult[HostServiceRuntimeAttributes]) (bool, error) {
+	if checkable.Attrs.Acknowledgement != AcknowledgementNone || checkable.Attrs.DowntimeDepth != 0 {
+		return true, nil
+	}
+
+	if checkable.Attrs.IsFlapping && checkable.Attrs.EnableFlapping {
+		status, err := client.fetchIcingaAppStatus(ctx)
+		if err != nil {
+			return false, err
+		}
+
+		return status.App.EnableFlapping, nil
+	}
+
+	return false, nil
 }
--- a/internal/icinga2/client.go
+++ b/internal/icinga2/client.go
@@ -265,7 +265,9 @@ func (client *Client) buildAcknowledgementEvent(ctx context.Context, ack *Acknow
 		if err != nil {
 			return nil, err
 		}
-		if !isMuted(queryResult) {
+		if muted, err := isMuted(ctx, client, queryResult); err != nil {
+			return nil, err
+		} else if !muted {
 			ev.Message = queryResult.Attrs.LastCheckResult.Output
 			ev.SetMute(false, "Acknowledgement cleared")
 		}
@@ -310,7 +312,9 @@ func (client *Client) buildDowntimeEvent(ctx context.Context, d Downtime, startE
 		if err != nil {
 			return nil, err
 		}
-		if !isMuted(queryResult) {
+		if muted, err := isMuted(ctx, client, queryResult); err != nil {
+			return nil, err
+		} else if !muted {
 			// When a downtime is cancelled/expired and there's no other active downtime/ack, we're going to send some
 			// notifications if there's still an active incident. Therefore, we need the most recent CheckResult of
 			// that Checkable to use it for the notifications.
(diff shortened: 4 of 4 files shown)


internal/icinga2/util.go
```diff
@@ -18,6 +19,23 @@ func rawurlencode(s string) string {
...
+func isMuted(ctx context.Context, client *Client, checkable *ObjectQueriesResult[HostServiceRuntimeAttributes]) (bool, error) {
+	if checkable.Attrs.IsFlapping {
+		if checkable.Attrs.EnableFlapping {
+			return true, nil
+		}
+
+		status, err := client.fetchIcingaAppStatus(ctx)
+		if err != nil {
+			return false, err
+		}
```
julianbrost, line 33: In this case, it may still be muted by something else. The other reasons should probably be checked first to avoid this API call if not necessary.

internal/icinga2/util.go
```diff
@@ -18,6 +19,23 @@ func rawurlencode(s string) string {
...
+
+		if status.App.EnableFlapping {
+			return true, nil
+		}
```
julianbrost, line 37: Have you verified that this actually matches the logic in Icinga 2? Like does the checkable attribute really have precedence over the global attribute? My intuition is that it has to be enabled both globally and on the checkable to actually be active, i.e. the global setting is more like a kill switch.

julianbrost (changes requested):

internal/icinga2/client_api.go
```diff
@@ -130,6 +130,31 @@ func (client *Client) queryObjectsApiDirect(ctx context.Context, objType, objNam
 		map[string]string{"Accept": "application/json"})
 }
 
+// fetchIcingaAppStatus retrieves the global state of the IcingaApplication type via the /v1/status endpoint.
+func (client *Client) fetchIcingaAppStatus(ctx context.Context) (*IcingaAppStatus, error) {
+	response, err := client.queryObjectsApi(
+		ctx,
+		[]string{"/v1/status/IcingaApplication/"},
```
oxzi, line 137: This will require additional permissions for the Icinga 2 API user. At the moment, `status/query` is not part of the permissions list, <https://icinga.com/docs/icinga-notifications-web/latest/doc/03-Configuration/#adding-an-icinga-2-source>.

Thus, the documentation over at :spider_web: needs an update.

internal/icinga2/client_api.go
```diff
@@ -130,6 +130,31 @@ func (client *Client) queryObjectsApiDirect(ctx context.Context, objType, objNam
 		map[string]string{"Accept": "application/json"})
 }
 
+// fetchIcingaAppStatus retrieves the global state of the IcingaApplication type via the /v1/status endpoint.
+func (client *Client) fetchIcingaAppStatus(ctx context.Context) (*IcingaAppStatus, error) {
```
oxzi, line 134: The position of this method does not fit right. It's currently located between between the two `queryObjectsApiDirect` and `queryObjectsApiQuery` methods. Please move it after the `queryObjects..` methods to the other `fetch..` methods.

internal/icinga2/client_api.go
```diff
@@ -275,17 +305,17 @@ func (client *Client) checkMissedChanges(ctx context.Context, objType string, ca
...
 			fakeEv, err = client.buildCommonEvent(ctx, hostName, serviceName)
 			if err != nil {
 				return fmt.Errorf("failed to construct checkable fake mute event: %w", err)
 			}
 
 			fakeEv.Type = event.TypeMute
-			if attrs.IsFlapping {
-				fakeEv.SetMute(true, "Checkable is flapping, but we missed the Icinga 2 FlappingStart event")
-			} else {
+			if attrs.DowntimeDepth != 0 {
```
oxzi, line 315: Could you please add a another note to `HostServiceRuntimeAttributes`' API documentation stating the relationship between `DowntimeDepth` and flapping.

internal/icinga2/util.go
```diff
@@ -18,6 +19,21 @@ func rawurlencode(s string) string {
 }
 
 // isMuted returns true if the given checkable is either in Downtime, Flapping or acknowledged, otherwise false.
-func isMuted(checkable *ObjectQueriesResult[HostServiceRuntimeAttributes]) bool {
-	return checkable.Attrs.IsFlapping || checkable.Attrs.Acknowledgement != AcknowledgementNone || checkable.Attrs.DowntimeDepth != 0
+// When the checkable is Flapping, and neither the flapping detection for that Checkable nor for the entire zone is
+// enabled, this will always return false.
```
oxzi, line 23: The API documentation does not mention that also an error is being returned.

internal/icinga2/client_api.go
yhabteab: > Thus, the documentation over at 🕸️ needs an update.

That's not something to be done with this PR! But once this is merged, I will open an issue there.

internal/icinga2/client_api.go
yhabteab: Like what for example? I'm not sure what I'm supposed to document here, but there is no relationship between those fields. I have switched the condition here because the conditions in the `isMuted()` function are also switched, so that when this function returns `true` because `downtimedepth != 0`, it doesn't go into the `flapping` branch.

internal/icinga2/client_api.go
yhabteab: Done https://github.com/Icinga/icinga-notifications-web/issues/228

internal/icinga2/client_api.go
oxzi: I don't know and that's why I asked about it. When reading the current code, I am surprised that `DowntimeDepth` decides if the object is either in a downtime or might be flapping. As there are three possibilities leading to a muted event, a bit more prose would at least help me understanding what's going on.

Maybe my suggested place is wrong and the `isMuted` function could get some additional documentation stating the implications or even get multiple returns like, e.g., `(isMutedAck, isMutedDowntime, isMutedFlapping bool, err error)` to remove domain-knowledge later on, like at this very place.

yhabteab: > As there are three possibilities leading to a muted event, a bit more prose would at least help me understanding what's going on.

Right, there are three possibilities and the first case is evaluated here.
https://github.com/Icinga/icinga-notifications/blob/03b8954092abccb320957e9e3f30f56560116a43/internal/icinga2/client_api.go#L294

The second one:
https://github.com/Icinga/icinga-notifications/blob/03b8954092abccb320957e9e3f30f56560116a43/internal/icinga2/client_api.go#L315

When those first two cases yield `false`, only one case remains, and that is `flapping` and has nothing to do with the above downtime check.
https://github.com/Icinga/icinga-notifications/blob/03b8954092abccb320957e9e3f30f56560116a43/internal/icinga2/client_api.go#L317-L319

oxzi: My point was that it adds additional mental load on the caller. This domain knowledge is necessary both within `isMuted` and outside after being called. Thus, there is no clean barrier and I have to exactly know how muting works, even outside the `isMuted` function.

For me, not having the amount of domain specific knowledge you do, I had to jump around the code and the Icinga 2 documentation to actually understand how it works. If I'm going to look at this next month, I have to recreate this mental model. That's all I wanted to point out. Technically this should work, but when I'm looking at the code the next time, I don't exactly know why.

yhabteab: > Thus, there is no clean barrier and I have to exactly know how muting works, even outside the isMuted function.

Sorry, but more than commenting on when `isMuted()` returns true and when it doesn't, I don't really know what else I should do to make it clearer for you. I haven't even changed the existing code that much that could confuse you even more, I have literally just switched that single condition.

> or even get multiple returns like, e.g., `(isMutedAck, isMutedDowntime, isMutedFlapping bool, err error)` to remove domain-knowledge later on, like at this very place.

I don't see how that is supposed to make things any better, the `isMuted` utility is simply meant to return true or false based on the given conditions, and if you want it to return something like that, I don't know why you would need such utility then.

> Maybe my suggested place is wrong and the isMuted function could get some additional documentation stating the implications

Sorry, but I failed to get what implications you are referring to. As I said above, there are no implications, this utility simply returns true or false based on a set condition and I'm checking them here one after the other to generate more specific events as to why an event/object is being muted. If you're having a hard time understanding the general handling of incident muting, we can have a call and discuss this in person, or I may be completely misunderstanding you and @julianbrost could just jump in and put a stop to it if he understands your point better.

oxzi: Sorry if my prior rambling wasn't on point. In a nutshell, I wanted to address a missing [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns). Please take a look at the following diff where I tried to change the signature of `isMuted`, resulting in removing the logic from `checkMissedChanges`.


```diff
diff --git a/internal/icinga2/client.go b/internal/icinga2/client.go
index 82cc20a..a30152b 100644
--- a/internal/icinga2/client.go
+++ b/internal/icinga2/client.go
@@ -265,7 +265,7 @@ func (client *Client) buildAcknowledgementEvent(ctx context.Context, ack *Acknow
 		if err != nil {
 			return nil, err
 		}
-		if muted, err := isMuted(ctx, client, queryResult); err != nil {
+		if muted, _, _, _, err := isMuted(ctx, client, queryResult); err != nil {
 			return nil, err
 		} else if !muted {
 			ev.Message = queryResult.Attrs.LastCheckResult.Output
@@ -312,7 +312,7 @@ func (client *Client) buildDowntimeEvent(ctx context.Context, d Downtime, startE
 		if err != nil {
 			return nil, err
 		}
-		if muted, err := isMuted(ctx, client, queryResult); err != nil {
+		if muted, _, _, _, err := isMuted(ctx, client, queryResult); err != nil {
 			return nil, err
 		} else if !muted {
 			// When a downtime is cancelled/expired and there's no other active downtime/ack, we're going to send some
@@ -351,7 +351,7 @@ func (client *Client) buildFlappingEvent(ctx context.Context, flapping *Flapping
 		if err != nil {
 			return nil, err
 		}
-		if muted, err := isMuted(ctx, client, queryResult); err != nil {
+		if muted, _, _, _, err := isMuted(ctx, client, queryResult); err != nil {
 			return nil, err
 		} else if !muted {
 			ev.Message = queryResult.Attrs.LastCheckResult.Output
diff --git a/internal/icinga2/client_api.go b/internal/icinga2/client_api.go
index a6cc385..9d85991 100644
--- a/internal/icinga2/client_api.go
+++ b/internal/icinga2/client_api.go
@@ -285,13 +285,13 @@ func (client *Client) checkMissedChanges(ctx context.Context, objType string, ca
 		}
 
 		attrs := objQueriesResult.Attrs
-		checkableIsMuted, err := isMuted(ctx, client, &objQueriesResult)
+		muted, mutedAck, mutedDowntime, mutedFlapping, err := isMuted(ctx, client, &objQueriesResult)
 		if err != nil {
 			return err
 		}
 
 		var fakeEv *event.Event
-		if attrs.Acknowledgement != AcknowledgementNone {
+		if mutedAck {
 			ackComment, err := client.fetchAcknowledgementComment(ctx, hostName, serviceName, attrs.AcknowledgementLastChange.Time())
 			if err != nil {
 				return fmt.Errorf("fetching acknowledgement comment for %q failed, %w", objectName, err)
@@ -305,17 +305,19 @@ func (client *Client) checkMissedChanges(ctx context.Context, objType string, ca
 			if err != nil {
 				return fmt.Errorf("failed to construct Event from Acknowledgement response, %w", err)
 			}
-		} else if checkableIsMuted {
+		} else if muted {
 			fakeEv, err = client.buildCommonEvent(ctx, hostName, serviceName)
 			if err != nil {
 				return fmt.Errorf("failed to construct checkable fake mute event: %w", err)
 			}
 
 			fakeEv.Type = event.TypeMute
-			if attrs.DowntimeDepth != 0 {
+			if mutedDowntime {
 				fakeEv.SetMute(true, "Checkable is in downtime, but we missed the Icinga 2 DowntimeStart event")
-			} else {
+			} else if mutedFlapping {
 				fakeEv.SetMute(true, "Checkable is flapping, but we missed the Icinga 2 FlappingStart event")
+			} else {
+				client.Logger.Error("lolwat")
 			}
 		} else {
 			// This could potentially produce numerous superfluous database (event table) entries if we generate such
diff --git a/internal/icinga2/util.go b/internal/icinga2/util.go
index 440f028..110359b 100644
--- a/internal/icinga2/util.go
+++ b/internal/icinga2/util.go
@@ -24,19 +24,30 @@ func rawurlencode(s string) string {
 // enabled, this will always return false.
 //
 // Returns an error if it fails to query the status of IcingaApplication from the /v1/status endpoint.
-func isMuted(ctx context.Context, client *Client, checkable *ObjectQueriesResult[HostServiceRuntimeAttributes]) (bool, error) {
-	if checkable.Attrs.Acknowledgement != AcknowledgementNone || checkable.Attrs.DowntimeDepth != 0 {
-		return true, nil
+func isMuted(ctx context.Context, client *Client, checkable *ObjectQueriesResult[HostServiceRuntimeAttributes]) (
+	muted, mutedAck, mutedDowntime, mutedFlapping bool, err error,
+) {
+	defer func() { muted = err == nil && (mutedAck || mutedDowntime || mutedFlapping) }()
+
+	if checkable.Attrs.Acknowledgement != AcknowledgementNone {
+		mutedAck = true
+		return
+	}
+	if checkable.Attrs.DowntimeDepth != 0 {
+		mutedDowntime = true
+		return
 	}
 
 	if checkable.Attrs.IsFlapping && checkable.Attrs.EnableFlapping {
-		status, err := client.fetchIcingaAppStatus(ctx)
+		var status *IcingaAppStatus
+		status, err = client.fetchIcingaAppStatus(ctx)
 		if err != nil {
-			return false, err
+			return
 		}
 
-		return status.App.EnableFlapping, nil
+		mutedFlapping = status.App.EnableFlapping
+		return
 	}
 
-	return false, nil
+	return
 }
```

oxzi: As discussed offline, we came to the conclusion of using a similar version as before, but hinting the reader by introducing `checkableIsMuted` to the first if clause.

oxzi (approved):

internal/icinga2/api_responses.go
```diff
@@ -288,6 +289,14 @@ type ObjectCreatedDeleted struct {
 	EventType  string `json:"type"`
 }
 
+// IcingaAppStatus represents the Icinga 2 API status endpoint query result of type IcingaApplication.
+// https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#status-and-statistics
+type IcingaAppStatus struct {
+	App struct {
+		EnableFlapping bool `json:"enable_flapping"`
+	} `json:"app"`
+}
```
julianbrost, line 298: This does not match the structure returned by that API endpoint:

```json
{
  "results": [
    {
      "name": "IcingaApplication",
      "perfdata": [],
      "status": {
        "icingaapplication": {
          "app": {
            "enable_event_handlers": true,
            "enable_flapping": true,
            "enable_host_checks": true,
            "enable_notifications": true,
            "enable_perfdata": true,
            "enable_service_checks": true,
            "environment": "",
            "node_name": "master-1",
            "pid": 19570,
            "program_start": 1721306330.03043,
            "version": "v2.14.0-206-gb468e0cee"
          }
        }
      }
    }
  ]
}
```

Thus, `EnableFlapping` always remains `false`.

julianbrost (changes requested):

internal/icinga2/client_api.go
```diff
@@ -149,6 +149,45 @@ func (client *Client) queryObjectsApiQuery(ctx context.Context, objType string,
...
+	if err != nil {
+		return nil, err
+	}
+
+	app := new(IcingaApplication)
+	if len(results) != 0 {
+		app = results[0].Status.IcingaApplication
+	}
+
+	return app, nil
```
julianbrost, line 188: The function already has an `error` return value, so I suggest returning an error if that information is missing (which it never should) instead of returning a zero-initialized struct.

julianbrost (approved):

oxzi (approved):

PR #231, opened 2024-07-10
Use CLI flags parser from Icinga Go Library

yhabteab:
This PR replaces the manual parsing of the CLI flags with the CLI flags parser from our Icinga Go Library and cleans up the main function a bit. The flags are now parsed during initialisation of the `daemon` package via the magical `init()` GO function and panics/exits when encountering unexpected inputs.

### Requires

- https://github.com/Icinga/icinga-go-library/pull/40


julianbrost:
> The flags are now parsed during initialisation of the `daemon` package via the magical `init()` GO function and panics/exits when encountering unexpected inputs.

What is the benefit of doing this in `init()` instead of just calling from `main()` like it's done so far? The latter is where I'd expect that to happen, using `init()` just adds code that's running implicitly but is harder to find (you couldn't set a breakpoint in `main()` to debug it for example). IMHO it's just very strange if programs exit before even entering `main()`.

yhabteab:
> What is the benefit of doing this in `init()` instead of just calling from `main()` like it's done so far?

The only benefit I had from doing this was that I didn't have to explicitly make an `Init` call from within the main function, but now I'm going to change it though.

internal/daemon/config.go
```diff
@@ -3,14 +3,20 @@ package daemon
 import (
 	"errors"
 	"github.com/creasty/defaults"
-	"github.com/goccy/go-yaml"
+	iglConfig "github.com/icinga/icinga-go-library/config"
```
julianbrost, line 6: Personally, I'd try to avoid import aliases and rather rename the existing `config` variable instead. Could be something like `daemonConfig`, `globalConfig` or maybe just shorten it to `conf`.

julianbrost (changes requested):

internal/daemon/config.go
```diff
@@ -31,62 +37,62 @@ func (c *ConfigFile) SetDefaults() {
...
 var _ defaults.Setter = (*ConfigFile)(nil)
 
-// config holds the configuration state as a singleton. It is used from LoadConfig and Config
-var config *ConfigFile
-
-// LoadConfig loads the daemon config from given path. Call it only once when starting the daemon.
-func LoadConfig(path string) error {
-	if config != nil {
-		return errors.New("config already set")
+func (c *ConfigFile) Validate() error {
```
oxzi, line 40: While changing this, maybe add a one-line API documentation stating that this method validates the whole configuration?

oxzi (commented):
Works as intended!

```
$ ./icinga-notifications --help
Usage:
  icinga-notifications [OPTIONS]

Application Options:
      --version  print version and exit
  -c, --config=  path to config file (default: /etc/icinga-notifications/config.yml)

Help Options:
  -h, --help     Show this help message

$ ./icinga-notifications --version
Icinga Notifications version: 0.0.0-gef68dc4-dirty

Build information:
  Go version: go1.22.5 (linux, amd64)
  Git commit: ef68dc48715792b1bf7d069aca9b4bdde9542c81 (modified)

System information:
  Platform: Red Star OS
  Platform version: 4.0

$ ./icinga-notifications
can't open YAML file /etc/icinga-notifications/config.yml: open /etc/icinga-notifications/config.yml: no such file or directory

$ ./icinga-notifications -c /dev/null
can't parse YAML file /dev/null: EOF

$ ./icinga-notifications --config /dev/null
can't parse YAML file /dev/null: EOF
```

internal/daemon/config.go
yhabteab: I have also added which interface it implements.

internal/daemon/config.go
```diff
@@ -28,65 +34,70 @@ func (c *ConfigFile) SetDefaults() {
...
-		return nil, err
+// daemonConfig holds the configuration state as a singleton.
+// It is initialised by the ParseFlagsAndConfig func and exposed through the Config function.
+var daemonConfig *ConfigFile
+
+// Config returns the config that was loaded while starting the daemon.
+// Panics when ParseFlagsAndConfig was not called earlier.
+func Config() *ConfigFile {
+	if daemonConfig == nil {
+		panic("ERROR: daemon.config not initialized")
```
julianbrost, line 72: This now references an outdated variable name. Also, the variable name is internal to the package, so I'd give the panic message the same treatment as the comment and say that `ParseFlagsAndConfig()` wasn't called.

internal/daemon/config.go
```diff
@@ -28,65 +34,70 @@ func (c *ConfigFile) SetDefaults() {
...
-		return nil, err
+// daemonConfig holds the configuration state as a singleton.
+// It is initialised by the ParseFlagsAndConfig func and exposed through the Config function.
+var daemonConfig *ConfigFile
+
+// Config returns the config that was loaded while starting the daemon.
+// Panics when ParseFlagsAndConfig was not called earlier.
+func Config() *ConfigFile {
+	if daemonConfig == nil {
+		panic("ERROR: daemon.Config called before daemon.ParseFlagsAndConfig()")
```
julianbrost, line 72: Both functions should be given consistently either with or without parenthesis (personally, I prefer it with as it's more obvious that those are functions).

oxzi (approved):

julianbrost (approved):

PR #234, opened 2024-07-12
Add AUTHORS and .mailmap

julianbrost:
Generated using:

```sh
git log --format='format:%aN <%aE>' | grep -vFx 'dependabot[bot] <49699333+dependabot[bot]@users.noreply.github.com>' | sort -u > AUTHORS
```

@Al2Klimov @raviks789 I've made the (non-)abbreviation of your names consistent with the [Icinga DB `AUTHORS` file](https://github.com/Icinga/icingadb/blob/main/AUTHORS), please let me know if you prefer something different.

PR #235, opened 2024-07-12
Version 0.1.0

julianbrost:
Changes in here are inspired by the [Icinga DB release checklist](https://github.com/Icinga/icingadb/blob/main/.github/ISSUE_TEMPLATE/release.md) (`AUTHROS` update was already done in #234).

The schema upgrade files were removed as they were mostly useful during development, with this release, 0.1.0 serves as the new baseline. The initial release is a bit of a special case where just removing them is fine as there's no previous release to
upgrade from. For later releases, they would have to be combined into a 0.2.0.sql instead.


julianbrost:
Apart from this PR, the only other thing that should make it into the `v0.1.0` tag is #249 so that it'll be pushed to Docker Hub automatically.

yhabteab (approved):
LFTM! You shouldn't merge this before #249 though.

PR #236, opened 2024-07-12
Increase details of Channels documentation

oxzi:
With a release approaching, the Channels documentation got more information about how the Channel system actually work and how to build your own channels.

The internal/changes/examples JSON files were inlined to ease the readability and updated, as they contained partially outdated information.

For the pkg/plugin part, some API documentation was added as this might get used by external users.


doc/10-Channels.md
```diff
@@ -1,88 +1,322 @@
...
+For logging or debugging, channels can write to `stderr`, which will be logged by Icinga Notifications.
 
 ### Request
 
-The request must be in JSON format and should contain following keys:
+A channel receives a request as a JSON object with the following fields:
 
-- `method`: The request method to call.
-- `params`: The params for request method.
+- `method`: The request method to call.
```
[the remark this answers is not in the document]
oxzi: :<

doc/10-Channels.md
```diff
@@ -1,88 +1,322 @@
...
+          "en_US": "Response Status Codes"
+        },
+        "help": {
+          "de_DE": "Kommaseparierte Liste erwarteter Status-Code der HTTP-Antwort, z.B.: 200,201,202,208,418",
+          "en_US": "Comma separated list of expected HTTP response status code, e.g., 200,201,202,208,418"
+        },
+        "default": "200",
+        "min": null,
+        "max": null
+      }
```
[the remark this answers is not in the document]
oxzi: Valid point. It was the output from the webhook channel. I have now reduced it to two options, being used later on.

doc/10-Channels.md
```diff
@@ -1,88 +1,298 @@
...
+
+Each request must be answered by the channel with a response JSON object of the following fields:
+
+- `result`: The result as JSON format. Omitted when the method does not return a value, i.e., for setter calls, or an
+  error has occurred.
+- `error`: The error message. Omitted when no error has occurred.
+- `id`: The request id.
+
+In case of a present `error` value, the `result` field should be empty.
+Successfully responses without a `result` do only contain the `id` field.
```
yhabteab, line 70: This (few other places as well) should be rephrased as [`Successful`](https://dictionary.cambridge.org/dictionary/english/successful) instead.

doc/10-Channels.md
```diff
@@ -1,88 +1,298 @@
...
+#### GetInfo
+
+The parameterless `GetInfo` method returns information about the channel.
+
+Its `result` is expected to be a JSON object with the `json` fields defined in the
+[`Info` type](https://pkg.go.dev/github.com/icinga/icinga-notifications/pkg/plugin#Info).
+The `config_attrs` field must be an array of JSON objects according to the
+[`ConfigOption` type](https://pkg.go.dev/github.com/icinga/icinga-notifications/pkg/plugin#ConfigOption).
+Those attributes define configuration options for the channel to be set via the `SetConfig` method.
+Furthermore, they are used to build the channel configuration UI in Icinga Notifications Web.
```
yhabteab, line 108: I think that `UI` is quite out of place. You might wanted to rephrase it as `... build the UI channel configuration in ...`?

pkg/plugin/plugin.go
```diff
@@ -69,57 +69,74 @@ type ConfigOption struct {
...
 	Addresses []*Address `json:"addresses"`
 }
 
+// Address to receive this notification. Each Contact might have multiple addresses.
 type Address struct {
 	Type    string `json:"type"`
 	Address string `json:"address"`
 }
 
+// Object which this NotificationRequest is all about, e.g., a service check.
```
yhabteab, line 105: I don't think a `service check` is an object.

oxzi: The text says that the object might be _about_ a service check, not being a service check.

doc/10-Channels.md
oxzi: I have rephrased the sentence.

pkg/plugin/plugin.go
yhabteab: > The text says that the object might be about a service check

That sentence doesn't make sense to me either. An `Event` could be about a service check, but I don't quite understand what an object being about a service check is supposed to mean.

oxzi: Then please take a look at the example `object`: <https://github.com/Icinga/icinga-notifications/blob/17cb89f00b9618c118bc91e2689dd98c65567980/doc/10-Channels.md#example-sendnotification-request>.

I am also open for alternative suggestions.

oxzi: Or was your point that it does not refer to a service check but more to a service object?

_Btw, I find the name `object` hard to grasp, especially as both in Icinga 2 and OOP lingo everything is a object._

doc/10-Channels.md
```diff
@@ -1,88 +1,298 @@
...
 
 ### Request
 
-The request must be in JSON format and should contain following keys:
+A channel receives a request as a JSON object with the following fields:
 
 - `method`: The request method to call.
-- `params`: The params for request method.
+- `params`: Optional params for the request method.
 - `id`: Unsigned int value. Required to assign the response to its request as responses can be sent out of order.
```
martialblog, line 37: Hi, 

the field description could be normalized:

`fieldname: description of what it does. optional/madatory. data type`

Regards

doc/10-Channels.md
```diff
@@ -1,88 +1,298 @@
...
+    },
+    "id": 1000
+  }
+  ```
+
+### Response
+
+Each request must be answered by the channel with a response JSON object of the following fields:
+
+- `result`: The result as JSON format. Omitted when the method does not return a value, i.e., for setter calls, or an
```
martialblog, line 64: Again, normalization could be useful:

`fieldname: description of what it does. optional/madatory. data type`

doc/10-Channels.md
```diff
@@ -1,88 +1,298 @@
...
+  error has occurred.
+- `error`: The error message. Omitted when no error has occurred.
+- `id`: The request id.
+
+In case of a present `error` value, the `result` field should be empty.
```
martialblog, line 69: Does "empty" mean `null` or the field can be omitted. Or does it mean empty string?

doc/10-Channels.md
```diff
@@ -1,88 +1,298 @@
...
   },
-  "id": 2020
+  "id": 1
 }
 ```
 
+#### SetConfig
+
+The `SetConfig` method configures the channel and
+should be called at least once before sending the first notifications to initialize the channel plugin.
```
martialblog, line 167: "should be called" by whom? The channel plugin?

doc/10-Channels.md
```diff
@@ -1,88 +1,298 @@
...
+To illustrate, the URL template from the above output is configurable with the following JSON object passed in `params`:
+
 ```json
 {
-  "method": "Foo",
+  "url_template": "http://localhost:8000/update/{{.Incident.Id}}"
+}
+```
+
+The response might carry an error if the configuration is invalid.
```
martialblog, line 181: What does "might" mean here? Optional? Recommended?

Might I suggest using the key words from RFC2119?

doc/10-Channels.md
```diff
@@ -1,88 +1,298 @@
...
+#### SendNotification
 
-Examples:
+The `SendNotification` method requests the channel to dispatch notifications.
+
+Within the request's `params`, a JSON object representing a
+[`NotificationRequest`](https://pkg.go.dev/github.com/icinga/icinga-notifications/pkg/plugin#NotificationRequest)
+is expected.
+
+The response might carry an error due to channel-specific reasons.
```
martialblog, line 212: Maybe add an example of what an error response

doc/10-Channels.md
yhabteab: Yes, maybe as a markdown table, like [Icinga 2](https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api/#creating-config-objects) does.

oxzi: Thanks., I liked this idea. I tried to unify all fields to:
> name: Short description. Type.

doc/10-Channels.md
oxzi: Good point. In the eyes of a channel developer, this is straight out confusing.

doc/10-Channels.md
oxzi: `s/might/MAY/g` :^)

I rewrote some sections to remove inaccuracies, use some RFC 2119 language (just w/o the caps), and to address a channel developer.

doc/10-Channels.md
oxzi: I switched to a table, as you suggested @yhabteab. This has the only downside that diffs altering the column width span over the whole table. But, it looks nicer :)

oxzi: @yhabteab: I tried to mimic Icinga 2's layout.

oxzi:
I reworked some parts, especially to mention changes from #206. Furthermore, I described how the channel configuration through the two RPC methods works and how a channel plugin is expected to behave. As the documentation links to the Go API doc, I added more API docs for the fields.

Please take another look.

doc/10-Channels.md
```diff
@@ -1,88 +1,355 @@
...
+
+```json
+{
+  "url_template": "http://localhost:8000/update/{{.Incident.Id}}"
 }
 ```
 
+If the channel plugin successfully configured itself, a response without an `result` must be returned.
+Otherwise, if the channel decides that the provided configuration is faulty, an `error` must be returned.
+This may happen if, i.e., an invalid configuration value was given.
```
yhabteab, line 200: I think you don't want to use [`i.e.`](https://dictionary.cambridge.org/dictionary/english/ie) here, but `for instance/example` instead.

doc/10-Channels.md
```diff
@@ -1,88 +1,317 @@
...
+    "id": 1000
+  }
+  ```
 
-- `method`: The request method to call.
-- `params`: The params for request method.
-- `id`: Unsigned int value. Required to assign the response to its request as responses can be sent out of order.
+### Response
+
+Each request must be answered by the channel with a response JSON object of the following fields:
```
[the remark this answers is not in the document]
oxzi: I removed the "response" but kept JSON object to exactly specify what we are talking about. I don't want to rely on too much implied information.

doc/10-Channels.md
oxzi: There will be a day when I finally learn how to correctly use _e.g._ and _i.e._

yhabteab (approved):
LFTM!

julianbrost (approved):

julianbrost:
@martialblog Do you want to have another look as well, i.e. should I wait with merging this a bit?

martialblog:
@julianbrost Thanks, I was away these last days. If I find something I'll just open a PR. 

Looking forward to v0.1.0

PR #237, opened 2024-07-12
Incremental Updates: honor events before changed_at

oxzi:
As changes might not be committed in the database in a continuous linear fashions, elements out of order might get ignored. This change adds a window to the past for incrementalFetch, honoring older elements. The incrementalApplyPending function now compares the changed_at value for known elements, skipping those without a change.

### DIFF
--- a/internal/config/incremental_sync.go
+++ b/internal/config/incremental_sync.go
@@ -35,6 +35,16 @@ type IncrementalConfigurableInitAndValidatable interface {
 	IncrementalInitAndValidate() error
 }
 
+// incrementalFetchBacklog is an offset for past changes to consider for incrementalFetch.
+//
+// It may happen that updates do not arrive in the database in order of their changed_at value. As this value is set by
+// Icinga Notifications Web on some webserver, there might be a delta between setting the timestamp and the entry
+// getting committed in the database. This delta may not be evenly distributed, especially when the web part processes
+// multiple changes concurrently.
+//
+// Fetched elements can be compared based on their changed_at value and be ignored if it has not changed.
+const incrementalFetchBacklog = 10 * time.Minute
+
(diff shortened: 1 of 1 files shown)


oxzi:
Something to note about this PR change is that it will retry the changes within each table for the last ten minutes. Since it will silently abort if the changed_at value has not changed or the item is deleted again, it does no real work, but it is not really clean, but more a hacky solution.

Another idea I toyed with was to use not only the last `changed_at` value, but also the current timestamp iff a `changed_at` value has already been fetched. For example, use `min(changed_at - 10m, current_timestamp - 30m)`. In my naive thinking, this sliding time window would slowly catch up the fixed `changed_at` timestamp, while still being large enough to catch the weirdest of time travelers[^1].

To be really sure, another layer could be added to query each known element if it has another `changed_at` or `deleted` value and to query for everything with an unknown primary key. This, however, should only be done once in a while.

[^1]: In case of a long lasting split brain situations of a database cluster, other issues might arise as well.

Al2Klimov (commented):
I.e this is a workaround for something web already could and should handle?

As such, it seems a bit over-engineered at first sight. There are other options:

* Write milliseconds where already ~allowed~ _desired_ (https://github.com/Icinga/icinga-notifications-web/pull/306)
* Write increasing values (https://github.com/Icinga/icinga-notifications-web/pull/311)

#

* If we decide to workaround this, here is something less complex: #288

But ideally I'd like web to handle this. I've already given them all instruments for that.

julianbrost:
Reopened because I have no idea what's going on here, especially given there is now [another PR over in the web project that wants to revert the PR that was used to close this PR](https://github.com/Icinga/icinga-notifications-web/pull/329).

Al2Klimov:
* Not necessary (to workaround this) anymore, as https://github.com/Icinga/icinga-notifications-web/pull/329 won't happen.

PR #238, opened 2024-07-12
GHA: complain if PR adds commits from people not yet listed in ./AUTHORS

Al2Klimov:
not to have to update ./AUTHORS or .mailmap after merging.


julianbrost (dismissed):
1. Instead of just copy & pasting this [file from another repository](https://github.com/Icinga/icinga2/blob/master/.github/workflows/authors-file.yml), please evaluate if/how this can be put into a central repo and then included in others.
2. This would fail for dependabot PRs unless we listed it in AUTHORS.

Al2Klimov:
👍 https://github.com/Al2Klimov/gha-AUTHORS/pull/2

yhabteab:
> 👍 [Al2Klimov/gha-AUTHORS#2](https://github.com/Al2Klimov/gha-AUTHORS/pull/2)

No. This doesn't need to be a real GitHub Action but just a workflow template that can be used in here (see https://github.com/Icinga/github-actions/tree/main/.github/workflows for inspiration). Please move the current version of the Icinga 2's workflow into this https://github.com/Icinga/github-actions repository and make use of it here.

yhabteab (approved):

PR #242, opened 2024-07-17
Channels: Rework default ConfigOptions and reorder methods

### LINKED ISSUE #240: Channels: Check Default and Required
Turns out, some defaults are not really constructive.. they might even be seen as destructive.

oxzi:
channels: Mark Required ConfigOptions

    Each plugin.ConfigOption was revisited and potentially being marked as
    Required, which was missing at some. Furthermore, the order of the
    fields was changed to match the struct field order of ConfigOption.

    channels: Reorder methods according to plugin.Plugin

    The plugin.Plugin interface specifies the three methods required for the
    channel plugin RPC API in their logical order. Considering that these
    channels are the blueprint for external channel development, the order
    of implementing channel plugins was changed to correspond to the
    interface.

    For all files, the main function has been moved to the top.

    In addition, the Email.Send method received a documentation string
    indicating that it implements the enmime.Sender interface, as it is not
    directly called anywhere. Further, the Email.GetServer method got
    inlined as it was only called from one other method.

Closes #240.

### DIFF
--- a/cmd/channels/email/main.go
+++ b/cmd/channels/email/main.go
@@ -115,24 +115,24 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
 func (ch *Email) GetInfo() *plugin.Info {
 	configAttrs := plugin.ConfigOptions{
 		{
-			Name:     "host",
-			Type:     "string",
-			Required: true,
+			Name: "host",
+			Type: "string",
 			Label: map[string]string{
 				"en_US": "SMTP Host",
 				"de_DE": "SMTP Host",
 			},
+			Required: true,
 		},
 		{
-			Name:     "port",
-			Type:     "number",
-			Required: true,
+			Name: "port",
+			Type: "number",
 			Label: map[string]string{
 				"en_US": "SMTP Port",
 				"de_DE": "SMTP Port",
 			},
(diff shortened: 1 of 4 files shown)


yhabteab:
> For all files, the main function has been placed at the bottom.

I wouldn't do that, IMHAO the main function should be at the top, before the other stuff.

oxzi:
> > For all files, the main function has been placed at the bottom.
> 
> I wouldn't do that, IMHAO the main function should be at the top, before the other stuff.

Honestly, I don't mind. It just should be at the same position.

If it's at the top, it should be before the struct type definition, as otherwise the `main` function appears between the type definition and its methods.

cmd/channels/email/main.go
```diff
@@ -159,6 +84,10 @@ func (ch *Email) GetInfo() *plugin.Info {
 				"en_US": "SMTP User",
 				"de_DE": "SMTP Benutzer",
 			},
+			Help: map[string]string{
+				"en_US": "When configuring a SMTP user, a SMTP password must also be set.",
```
yhabteab, line 88: I think it should be `... an SMTP ...` in both cases.

yhabteab (approved):

julianbrost (approved):

PR #243, opened 2024-07-17
icinga2: Reset `Event#ID` before resending ack set event

yhabteab:
`Event#Sync()` synchronises an event row with the database only if its ID is set to no other value than `0`. Therefore, we must reset the event ID for the acknowledgement set event generated in the catch-up phase before resending it.

### DIFF
--- a/internal/icinga2/client_api.go
+++ b/internal/icinga2/client_api.go
@@ -301,6 +301,7 @@ func (client *Client) checkMissedChanges(ctx context.Context, objType string, ca
 		}
 
 		fakeEv.Message = attrs.LastCheckResult.Output
+		ackEvent := *fakeEv
 		select {
 		case catchupEventCh <- &catchupEventMsg{eventMsg: &eventMsg{fakeEv, attrs.LastStateChange.Time()}}:
 			if fakeEv.Type == event.TypeUnmute {
@@ -325,7 +326,7 @@ func (client *Client) checkMissedChanges(ctx context.Context, objType string, ca
 				select {
 				// Retry the AckSet event so that the author of the ack is set as the incident
 				// manager if there was no existing incident before the above state change event.
-				case catchupEventCh <- &catchupEventMsg{eventMsg: &eventMsg{fakeEv, attrs.LastStateChange.Time()}}:
+				case catchupEventCh <- &catchupEventMsg{eventMsg: &eventMsg{&ackEvent, attrs.LastStateChange.Time()}}:
 				case <-ctx.Done():
 					return ctx.Err()
 				}


oxzi (commented):
I tried to retrace the steps how this can happened. Thanks for giving me some pointers, @yhabteab, as otherwise I have totally missed it.

- The journey of our new `fakeEv` starts in `checkMissedChanges`, from where it will be send to `catchupEventCh` by reference
  https://github.com/Icinga/icinga-notifications/blob/15392c962f7f84eacf1eaaf9abb0c2c05f6b3ac8/internal/icinga2/client_api.go#L305
- It ends up in the `worker`'s catchup case..
  https://github.com/Icinga/icinga-notifications/blob/683ea699e7cfdf2aa4090883170ef103cfea3a6b/internal/icinga2/client.go#L499
- ..and will be eventually passed to the `CallbackFn`
  https://github.com/Icinga/icinga-notifications/blob/683ea699e7cfdf2aa4090883170ef103cfea3a6b/internal/icinga2/client.go#L538
- In there, our adventurous event will be passed on to `incident.ProcessEvent`
  https://github.com/Icinga/icinga-notifications/blob/7b6ea92672841ad3ba28927d8c00e9ae68443ce0/internal/icinga2/launcher.go#L132
- As it is a new incident, it gets `.Sync`ed
  https://github.com/Icinga/icinga-notifications/blob/7b6ea92672841ad3ba28927d8c00e9ae68443ce0/internal/incident/incidents.go#L248
- And ends up in the database and gets its `.ID` set
  https://github.com/Icinga/icinga-notifications/blob/7b6ea92672841ad3ba28927d8c00e9ae68443ce0/internal/event/event.go#L164-L168
- Back in `checkMissedChanges`, after sending the host service event, the `fakeEv` might be sent again
  https://github.com/Icinga/icinga-notifications/blob/15392c962f7f84eacf1eaaf9abb0c2c05f6b3ac8/internal/icinga2/client_api.go#L322-L329
- Now, I am going to skip some steps as shown above until the event ends up in `Sync`
  https://github.com/Icinga/icinga-notifications/blob/7b6ea92672841ad3ba28927d8c00e9ae68443ce0/internal/event/event.go#L160-L162
- And here it goes silently unnoticed as it got an `ID` in the meantime. Yikes.

I am going to take another look tomorrow.

yhabteab:
> * As it is a new incident, it gets `.Sync`ed
>   https://github.com/Icinga/icinga-notifications/blob/7b6ea92672841ad3ba28927d8c00e9ae68443ce0/internal/incident/incidents.go#L248

If it is synced here, there should be no problems. The real problem, however, is that if there is an existing incident, that fake event will make its way to the fellowing scopes multiple times.
https://github.com/Icinga/icinga-notifications/blob/59861289ce06d4b264f1b42e941d3b520cba332a/internal/incident/incident.go#L147
https://github.com/Icinga/icinga-notifications/blob/59861289ce06d4b264f1b42e941d3b520cba332a/internal/incident/incident.go#L163-L164

When the incident finally notices that this ack set event is superfluous here ...
https://github.com/Icinga/icinga-notifications/blob/59861289ce06d4b264f1b42e941d3b520cba332a/internal/incident/incident.go#L199-L200
It will simply return without committing the transaction, leaving the now dangling `ev.ID` set above.

internal/icinga2/client_api.go
```diff
@@ -322,6 +322,7 @@ func (client *Client) checkMissedChanges(ctx context.Context, objType string, ca
 		case catchupEventCh <- &catchupEventMsg{eventMsg: &eventMsg{ev, attrs.LastStateChange.Time()}}:
 			stateChangeEvents++
 			if fakeEv.Type == event.TypeAcknowledgementSet {
+				fakeEv.ID = 0 // We don't want to reference a previously sent event, so reset its ID!
```
julianbrost, line 325: Instead of doing this, I think it would be much safer to duplicate the whole event before sending it to any channel. That would be safe without having to argue about whether the first event is safe to be used again at this point. It might be safe due to the channel not being buffered, the receiver processing the events serially, **and** another event being sent in between, but change any of this, and resetting `fakeEv.ID` here would become unsafe.

internal/icinga2/client_api.go
```diff
@@ -322,10 +322,12 @@ func (client *Client) checkMissedChanges(ctx context.Context, objType string, ca
 		case catchupEventCh <- &catchupEventMsg{eventMsg: &eventMsg{ev, attrs.LastStateChange.Time()}}:
 			stateChangeEvents++
 			if fakeEv.Type == event.TypeAcknowledgementSet {
+				ackEvent := *fakeEv
+				ackEvent.ID = 0 // We don't want to reference a previously sent event, so reset its ID!
```
julianbrost, line 326: > duplicate the whole event before sending it to any channel

By that, I meant before `fakeEv` is sent to any channel for the first time, i.e. before line 305. That way, the pointer wasn't yet passed anywhere else and duplicating it is safe for sure. So that would mean creating a second variable next to `fakeEv` and duplicating the event right after `client.buildAcknowledgementEvent()`.

oxzi (approved):

julianbrost (approved):

PR #244, opened 2024-07-19
Fix warnings during docker build

julianbrost:
Building the container image started showing the following two warning for me recently which are addressed by this commit:

```
2 warnings found (use --debug to expand):
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 3)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 4)
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # Icinga Notifications | (c) 2023 Icinga GmbH | GPLv2+
 
-FROM docker.io/library/golang as build
-ENV CGO_ENABLED 0
+FROM docker.io/library/golang AS build
+ENV CGO_ENABLED=0
 COPY . /src/icinga-notifications
 WORKDIR /src/icinga-notifications
 


oxzi (approved):
Unfortunately, either my local Docker version is too old to show errors or I just failed at reproducing them. Even calling it by `BUILDX_METADATA_WARNINGS=true docker build --no-cache -t icinga-notifications .` does not do the thing.

However, your change looks good and seems to match the warnings. Thus, if it fixes the warnings on your side, I am fine with it. Both the old and the new Dockerfile works on my machine™.

julianbrost:
Yeah, not sure what exactly brought the change. Currently, I'm on Docker 27.0.3 with buildx 0.15.1.

julianbrost:
Combining my package manager logs and when I think I've seen the warning the first time, it could have been the Docker 26 → 27 upgrade. However, I didn't find something obvious in the changelog.

PR #249, opened 2024-07-24
Docker GHA: Upload Docker images on release

oxzi:
With this change, the Docker GitHub Action now creates multi-platform Docker images on releases and pushes those to Docker Hub. The changes were both inspired by the Docker defaults and <https://github.com/Icinga/icinga-kubernetes/blob/main/.github/workflows/docker-image.yml>.

Note: I am unable to check if the necessary environment variables are already set in this repository. This needs to be done by somebody with the sufficient privileges first. Furthermore, there is no project on Docker Hub so far, maybe it needs to be created first.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,12 +3,41 @@ name: Docker
 on:
   push:
     branches: [ main ]
-  pull_request: {}
+  pull_request:
+    branches: [ main ]
+  release:
+    types: [ published ]
 
 jobs:
   docker:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    - name: Build Image
+
+    - uses: docker/metadata-action@v5
+      id: meta
+      with:
+        images: |
+          icinga/icinga-notifications
+        tags: |
+          type=ref,event=branch,enable={{is_default_branch}}
+          type=semver,pattern={{version}}
+          type=semver,pattern={{major}}.{{minor}}
+
+    - uses: docker/setup-qemu-action@v3
+
+    - uses: docker/setup-buildx-action@v3
+
+    - if: github.event_name != 'pull_request'
+      uses: docker/login-action@v3
+      with:
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+    - name: Build and push
       uses: docker/build-push-action@v6
+      with:
+        labels: ${{ steps.meta.outputs.labels }}
+        platforms: linux/amd64,linux/arm64
+        push: ${{ github.event_name != 'pull_request' }}
+        tags: ${{ steps.meta.outputs.tags }}


.github/workflows/docker.yml
```diff
@@ -3,12 +3,37 @@ name: Docker
...
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    - name: Build Image
+
+    - uses: docker/metadata-action@v5
+      id: meta
+      with:
+        images: |
+          icinga/icinga-notifications
```
julianbrost, line 21: Can you please summarize which tags this will push by default?

@lippserd I wonder why you [went with `type=edge` in Icinga for Kubernetes](https://github.com/Icinga/icinga-kubernetes/blob/854c171a436a6a4d951c618f9f44a2067fd8587d/.github/workflows/docker-image.yml#L20-L21) (I also wonder why it exists in the first place in metadata-action, it seems to do nothing special other than renaming the tag for the default branch to edge?)

.github/workflows/docker.yml
```diff
@@ -3,12 +3,37 @@ name: Docker
 on:
   push:
     branches: [ main ]
-  pull_request: {}
+  pull_request:
+    branches: [ main ]
+  release:
+    types: [ published ]
```
julianbrost, line 9: Have you tested (in another repo for example) that this actually works in combination with `docker/metadata-action`? I know that it was me who suggested doing it on release publish instead of tag push, but after looking at [the documentation](https://github.com/marketplace/actions/docker-metadata-action) which only ever mentions `event=tag` but never `release`, I'm not sure if this will just work (though I'd prefer it if it does).

.github/workflows/docker.yml
julianbrost: I went ahead playing with this in a private repo myself. Results:

### default

```yaml
# tags: (not set)
```

#### push to main

```
icinga/icinga-notifications:main
```

#### release published

```
icinga/icinga-notifications:v0.1.0
icinga/icinga-notifications:latest
```

**Note:** Release tag has `v` prefix in contrast to our other images (Icinga 2, Icinga Web, Icinga DB)

### edge + semver

```yaml
tags: |
  type=edge
  type=semver,pattern={{version}}
  type=semver,pattern={{major}}.{{minor}}
```

#### push to main

```
icinga/icinga-notifications:edge
```

#### release published

```
icinga/icinga-notifications:0.2.0
icinga/icinga-notifications:0.2
icinga/icinga-notifications:latest
```

### ref + semver

```yaml
tags: |
  type=ref,event=branch,enable={{is_default_branch}}
  type=semver,pattern={{version}}
  type=semver,pattern={{major}}.{{minor}}
```

#### push to main

```
icinga/icinga-notifications:main
```

#### release published

```
icinga/icinga-notifications:0.3.0
icinga/icinga-notifications:0.3
icinga/icinga-notifications:latest
```

### Summary

I'd suggest to go with the last variant (ref + semver), unless we decide to go for `:edge` as the snapshot/nightly/bleeding edge tag for all our images.

.github/workflows/docker.yml
julianbrost: The tests I did for https://github.com/Icinga/icinga-notifications/pull/249#discussion_r1689324473 were using the release published event, so yes, it works.

.github/workflows/docker.yml
lippserd: I decided to use `edge` to indicate snapshot builds. At least that's what I've found is more commonly used than `snapshot` for example. At the time of the release of version 0.1.0, I didn't have the right rules configured, so an image with tag 0.1.0/latest was never uploaded and I didn't push one either, as I thought we'd release 0.2.0 faster.

.github/workflows/docker.yml
lippserd: That's why I've also opted for ` tags: [ 'v*' ]`.

.github/workflows/docker.yml
lippserd: > unless we decide to go for `:edge` as the snapshot/nightly/bleeding edge tag for all our images

Sounds like a good idea to me.

.github/workflows/docker.yml
lippserd: > The tests I did for [#249 (comment)](https://github.com/Icinga/icinga-notifications/pull/249#discussion_r1689324473) were using the release published event, so yes, it works.

Missed your comment before posting my comment. I will use that as well in Icinga for Kubernetes then.

.github/workflows/docker.yml
julianbrost: > I didn't have the right rules configured, so an image with tag 0.1.0/latest was never uploaded and I didn't push one either, as I thought we'd release 0.2.0 faster.

FYI: I think the [current version of the Icinga for Kubernetes](https://github.com/Icinga/icinga-kubernetes/blob/main/.github/workflows/docker-image.yml) won't upload for a tag/release as setting `type=edge` should disable all other rules that would by default set version tags.

julianbrost: Settings all version tags properly will be another task. Keep in mind that this becomes extra tricky when you release versions out of order, i.e. release 2.14.0 and afterwards, you want to release a bug fix as 2.13.5. In Icinga 2, we currently have [this beautiful script](https://github.com/Icinga/docker-icinga2/blob/master/mktags.bash) for that.

For this project (and probably Icinga for Kubernetes as well), this won't be an issue until at least version 1.1.0. I think the general plan here is to use this here a bit as a playground for how to build and push the Docker images, next, that should be adapted for the existing images, where that issue should then be solved.

julianbrost:
> Note: I am unable to check if the necessary environment variables are already set in this repository. This needs to be done by somebody with the sufficient privileges first. Furthermore, there is no project on Docker Hub so far, maybe it needs to be created first.

Looks like someone set them an hour ago (maybe @lippserd, it roughly matches when he commented here :sweat_smile:), so that should just work now.

.github/workflows/docker.yml
oxzi: > Settings all version tags properly will be another task. Keep in mind that this becomes extra tricky when you release versions out of order, i.e. release 2.14.0 and afterwards, you want to release a bug fix as 2.13.5. In Icinga 2, we currently have [this beautiful script](https://github.com/Icinga/docker-icinga2/blob/master/mktags.bash) for that.

Valid point, I have only thought about continuous release numbers.

> For this project (and probably Icinga for Kubernetes as well), this won't be an issue until at least version 1.1.0. I think the general plan here is to use this here a bit as a playground for how to build and push the Docker images, next, that should be adapted for the existing images, where that issue should then be solved.

Then I would not include it, or at least not at the moment.

julianbrost (approved):

julianbrost:
`main` branch from this repo uploaded successfully: https://hub.docker.com/r/icinga/icinga-notifications/tags (job: https://github.com/Icinga/icinga-notifications/actions/runs/10074220195/job/27849849365)

I created a [fork to test the release workflow](https://github.com/julianbrost/test-icinga-notifications/actions/runs/10074304888/job/27850117048) with the merged version, of course fails during docker login, but it would set the correct tags:

```
Run docker/metadata-action@v5
Context info
Processing images input
Processing tags input
Processing flavor input
Docker image version
Docker tags
  icinga/icinga-notifications:0.1.0
  icinga/icinga-notifications:0.1
  icinga/icinga-notifications:latest
```

PR #251, opened 2024-07-24
listener: Don't error w/ 500 for superfluous events

oxzi:
When, for example, resubmitting a state change, the listener will error with an internal server error, code 500. This is both a bit heavy and misleading.

The error is now inspected similar to the current behavior for the icinga2.Launcher in its launch method.

### DIFF
--- a/internal/listener/listener.go
+++ b/internal/listener/listener.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/subtle"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"github.com/icinga/icinga-go-library/database"
 	"github.com/icinga/icinga-go-library/logging"
@@ -133,7 +134,10 @@ func (l *Listener) ProcessEvent(w http.ResponseWriter, req *http.Request) {
 
 	l.logger.Infow("Processing event", zap.String("event", ev.String()))
 	err = incident.ProcessEvent(context.Background(), l.db, l.logs, l.runtimeConfig, &ev)
-	if err != nil {
+	if errors.Is(err, event.ErrSuperfluousStateChange) || errors.Is(err, event.ErrSuperfluousMuteUnmuteEvent) {
+		abort(http.StatusNotAcceptable, &ev, "%v", err)
+		return
+	} else if err != nil {
 		l.logger.Errorw("Failed to successfully process event", zap.Stringer("event", &ev), zap.Error(err))
 		abort(http.StatusInternalServerError, &ev, "event could not be processed successfully, see server logs for details")
 		return


internal/listener/listener.go
```diff
@@ -133,7 +134,16 @@ func (l *Listener) ProcessEvent(w http.ResponseWriter, req *http.Request) {
...
 	l.logger.Infow("Processing event", zap.String("event", ev.String()))
 	err = incident.ProcessEvent(context.Background(), l.db, l.logs, l.runtimeConfig, &ev)
-	if err != nil {
+	switch {
+	case errors.Is(err, event.ErrSuperfluousStateChange):
+		abort(http.StatusAlreadyReported, &ev, "Stopped processing event with superfluous state change: %v", err)
+		return
+
+	case errors.Is(err, event.ErrSuperfluousMuteUnmuteEvent):
+		abort(http.StatusAlreadyReported, &ev, "Stopped processing event with superfluous (un)mute object: %v", err)
```
julianbrost, line 143: The messages given here are somewhat redundant with what's already part of `err.Error()`, so this results in responses like this:

```
Stopped processing event with superfluous state change: ignoring superfluous state change: ok state event from source 1
```

julianbrost (changes requested):

internal/listener/listener.go
oxzi: I have unified both cases and picked another <small>arbitrary</small> HTTP status code. Now the returned message looks like:
> ignoring superfluous state change: ok state event from source 2

internal/listener/listener.go
```diff
@@ -133,7 +134,14 @@ func (l *Listener) ProcessEvent(w http.ResponseWriter, req *http.Request) {
...
 	err = incident.ProcessEvent(context.Background(), l.db, l.logs, l.runtimeConfig, &ev)
-	if err != nil {
+	switch {
+	case errors.Is(err, event.ErrSuperfluousStateChange):
+		fallthrough
+	case errors.Is(err, event.ErrSuperfluousMuteUnmuteEvent):
+		abort(http.StatusNotAcceptable, &ev, "%v", err)
+		return
+
+	case err != nil:
```
julianbrost, line 144: I already noticed that you're a way bigger fan of `switch` than I am, but with the `fallthrough` here, a regular `if` looks way nicer to me:
```suggestion
	if errors.Is(err, event.ErrSuperfluousStateChange) || errors.Is(err, event.ErrSuperfluousMuteUnmuteEvent) {
		abort(http.StatusNotAcceptable, &ev, "%v", err)
		return
	} else if err != nil {
```

oxzi: Fair enough.

julianbrost (approved):

PR #269, opened 2024-08-01
webhook: Basic TLS Configuration, Custom Headers

### LINKED ISSUE #256: Custom TLS options for Webhook Channel
Hoi,

the webhook channel could use some TLS options, what if I want to use my shiny custom CA!

Regards

oxzi:
This change allows some basic TLS related configuration changes:
- Fixed Certificate Common Name
- Custom CA
- Insecure Mode w/o any verification

The last setting brought me to an unexpected boolean representation in the JSON configuration, requiring a custom boolean type. It was added to pkg/plugin and its documentation.

In addition, support for arbitrary HTTP request headers was added. This allows custom headers, even carrying values based on the NotificationRequest.

Additionally, an unique User-Agent was assigned.

Closes #256.

### DIFF
--- a/cmd/channels/webhook/main.go
+++ b/cmd/channels/webhook/main.go
@@ -2,12 +2,15 @@ package main
 
 import (
 	"bytes"
+	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"fmt"
 	"github.com/icinga/icinga-notifications/internal"
 	"github.com/icinga/icinga-notifications/pkg/plugin"
 	"io"
 	"net/http"
+	"os"
 	"slices"
 	"strconv"
 	"strings"
@@ -18,16 +21,30 @@ func main() {
 	plugin.RunPlugin(&Webhook{})
 }
 
+// transport is a http.Transport with a custom User-Agent.
+type transport http.Transport
+
+// RoundTrip implements http.RoundTripper with a custom User-Agent header.
+func (trans *transport) RoundTrip(req *http.Request) (*http.Response, error) {
+	req.Header.Set("User-Agent", "icinga-notifications-webhook/"+internal.Version.Version)
+	return (*http.Transport)(trans).RoundTrip(req)
+}
+
 type Webhook struct {
-	Method              string `json:"method"`
-	URLTemplate         string `json:"url_template"`
-	RequestBodyTemplate string `json:"request_body_template"`
-	ResponseStatusCodes string `json:"response_status_codes"`
+	Method                 string      `json:"method"`
+	URLTemplate            string      `json:"url_template"`
+	RequestHeadersTemplate string      `json:"request_headers_template"`
+	RequestBodyTemplate    string      `json:"request_body_template"`
+	ResponseStatusCodes    string      `json:"response_status_codes"`
+	TlsCommonName          string      `json:"tls_common_name"`
+	TlsCaPemFile           string      `json:"tls_ca_pem_file"`
+	TlsInsecure            plugin.Bool `json:"tls_insecure"`
 
-	tmplUrl         *template.Template
-	tmplRequestBody *template.Template
-
-	respStatusCodes []int
+	tmplUrl            *template.Template
+	tmplRequestHeaders map[string]*template.Template
+	tmplRequestBody    *template.Template
+	respStatusCodes    []int
+	httpTransport      http.RoundTripper
 }
 
 func (ch *Webhook) GetInfo() *plugin.Info {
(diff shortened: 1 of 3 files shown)


cmd/channels/webhook/main.go
```diff
@@ -21,16 +26,21 @@ func main() {
...
-	RequestBodyTemplate string `json:"request_body_template"`
-	ResponseStatusCodes string `json:"response_status_codes"`
+	Method                 string `json:"method"`
+	URLTemplate            string `json:"url_template"`
+	RequestHeadersTemplate string `json:"request_headers_template"`
+	RequestBodyTemplate    string `json:"request_body_template"`
+	ResponseStatusCodes    string `json:"response_status_codes"`
+	TlsServerName          string `json:"tls_server_name"`
+	TlsCaPem               string `json:"tls_ca_pem"`
+	TlsInsecure            bool   `json:"tls_insecure"`
```
yhabteab, line 40: If we're adding TLS support then we should also fully support it, i.e., missing client cert here.

cmd/channels/webhook/main.go
```diff
@@ -62,9 +72,22 @@ func (ch *Webhook) GetInfo() *plugin.Info {
...
 		},
+		{
+			Name: "request_headers_template",
+			Type: "text",
+			Label: map[string]string{
+				"en_US": "Request Header Template",
+				"de_DE": "Request Header-Template",
+			},
+			Help: map[string]string{
+				"en_US": "Multiple lines of 'HTTP-HEADER=TEMPLATE' with TEMPLATE being a Go template about the current plugin.NotificationRequest.",
```
yhabteab: How about rephrasing this something like this instead?

```suggestion
				"en_US": "Multiline separated of 'HTTP-HEADER=TEMPLATE' pairs, where TEMPLATE is a Go template over the current plugin.NotificationRequest.",
```

cmd/channels/webhook/main.go
```diff
@@ -62,9 +72,22 @@ func (ch *Webhook) GetInfo() *plugin.Info {
...
+				"de_DE": "Mehrere Zeilen im Format 'HTTP-HEADER=TEMPLATE', wobei TEMPLATE ein Go-Template über das zu verarbeitende plugin.NotificationRequest ist.",
+			},
+			Default: "",
```
yhabteab: Is this default value really needed?

cmd/channels/webhook/main.go
```diff
@@ -21,16 +26,21 @@ func main() {
...
-	URLTemplate         string `json:"url_template"`
-	RequestBodyTemplate string `json:"request_body_template"`
-	ResponseStatusCodes string `json:"response_status_codes"`
+	Method                 string `json:"method"`
+	URLTemplate            string `json:"url_template"`
+	RequestHeadersTemplate string `json:"request_headers_template"`
+	RequestBodyTemplate    string `json:"request_body_template"`
+	ResponseStatusCodes    string `json:"response_status_codes"`
+	TlsServerName          string `json:"tls_server_name"`
+	TlsCaPem               string `json:"tls_ca_pem"`
```
yhabteab: As discussed, this should be renamed to `tls_ca_cert` instead.

cmd/channels/webhook/main.go
```diff
@@ -142,17 +229,47 @@ func (ch *Webhook) SetConfig(jsonStr json.RawMessage) error {
...
+				return fmt.Errorf("cannot add CA PEM to cert pool")
+			}
+		} else if caPem, err := os.ReadFile(tmpWh.TlsCaPem); err == nil {
+			if !tlsConf.RootCAs.AppendCertsFromPEM(caPem) {
+				return fmt.Errorf("cannot add CA PEM from file %q to cert pool", tmpWh.TlsCaPem)
+			}
+		} else {
+			return fmt.Errorf("cannot load CA PEM: %w", err)
+		}
+	}
```
yhabteab: Once you add client cert support too, you can just use our `config.TLS` from IGL instead doing this by yourself here.

cmd/channels/webhook/main.go
```diff
@@ -177,7 +299,17 @@ func (ch *Webhook) SendNotification(req *plugin.NotificationRequest) error {
 	if err != nil {
 		return err
 	}
-	httpResp, err := http.DefaultClient.Do(httpReq) // #nosec G704 -- no SSRF, trusted user input
+
+	httpReq.Header.Set("User-Agent", "icinga-notifications-webhook/"+internal.Version.Version)
```
yhabteab, line 355: Shouldn't this be moved to below the loop, so users can't override it?

yhabteab (changes requested):

cmd/channels/webhook/main.go
oxzi: This was placed intentionally before the user-configured headers, allowing users to overwrite the User-Agent if they seem fit.

cmd/channels/webhook/main.go
oxzi: Renamed to `tls_ca` to mirror `config.TLS` struct field names.

cmd/channels/webhook/main.go
oxzi: Good idea. Fortunately, the exact same logic is already implemented there.

cmd/channels/webhook/main.go
oxzi: As discussed, I have added mTLS support as well.

cmd/channels/webhook/main.go
yhabteab: Ok, I thought you wanted to use a consistent user agent for all installations.

cmd/channels/webhook/main.go
oxzi: Reworded it a bit.

oxzi:
Added a generic _TLS Config_ option and moved all TLS-related configuration entries below it as its child settings. With https://github.com/Icinga/icinga-notifications-web/pull/559, this will allow a foldable config in the Web UI.

Furthermore, I have exchanged the Request Header separator from `=` to `:`, mirroring the actual HTTP standard.

@yhabteab, please take another look :)

yhabteab (approved):

PR #270, opened 2024-08-01
Preparation for sending non-state notifications

yhabteab:
This PR introduces/extracts some of the codes of the existing components in preparation for #188. All these changes were part of that PR, but as it was slowly getting bigger and bigger, I decided to split up those commits which can be reviewed and merged independently.

- 503b8742cc291d9eca7fb5950d0b5afdf15c48e4 Contains changes for the channel plugin, that should allow us to send non-state notifications without an incident in the future and it adds also some missing doc blocks.
- 592ac09f47220e3bfb5f27521f623bb95e638704 Just adds a simple one-line debug log in which superfluous state changes are ignored.
- f9a9becbecc65e5b73de9b12cbfbe99297ba0ae9 Is the somewhat bigger commit and also far more relevant than the others, as it extracts and/or decouples the entire rule and escalation filter evaluation from the incident package into a new common type and adds even more flexibility for #188. It makes it possible, for instance, to create test cases just for the filter evaluation part and also removes a number of obstacles for #188.


yhabteab:
I think, this PR is ready to be reviewed now. It's based on top of #324, so can't be merged before that one but can be reviewed meanwhile. Please take a look when either of you have time. Thanks! :)

PR #288, opened 2025-03-11
incrementalFetch(): fix race condition which would miss rows based on changed_at

Al2Klimov:
* Re: https://github.com/Icinga/icinga-notifications-web/issues/226#issuecomment-2695054673

incrementalFetch() relies on every newly inserted row having a higher changed_at than every already committed one. At the moment, web doesn't guarantee this. It even writes changed_at rounded to whole seconds. Hence, we could skip one of two rows which appeared during the same second.

This change simply ignores the most recent 3s of config updates (for 3s), so that two or more new rows which appear during the same second aren't a problem. They're processed together 3s after they get committed.

closes #237

### DIFF
--- a/internal/config/incremental_sync.go
+++ b/internal/config/incremental_sync.go
@@ -62,9 +62,12 @@ func incrementalFetch[
 		stmtArgs []any
 	)
 	if hasChangedAt {
-		stmtLogger = stmtLogger.With(zap.Time("changed_at", changedAt.Time()))
-		stmt += ` WHERE "changed_at" > ?`
-		stmtArgs = []any{changedAt}
+		const threshold = 3 * time.Second
+		filter := time.Now().Add(-threshold)
+
+		stmtLogger = stmtLogger.With(zap.Time("changed_at", changedAt.Time()), zap.Duration("threshold", threshold))
+		stmt += ` WHERE "changed_at" > ? AND "changed_at" < ?`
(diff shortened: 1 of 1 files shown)


oxzi:
Please rebase to rerun the failing CI job, fixed in the current main branch.

Al2Klimov:
This is just a workaround. I prefer to fix the root cause properly, i.e:

* https://github.com/Icinga/icinga-notifications-web/pull/311

If you agree, please review that PR's intention (description) itself, not necessarily the PHP code.

Al2Klimov:
@nilmerg You said you don't like voluntary thresholds like these 3s. But Icinga DB already gives up after the DB is down for 5m which is also a voluntary threshold. :)

Al2Klimov:
As discussed offline with @nilmerg, I think a changed_at map per row id, and not just per table, could also solve this, but this PR is the way of the least resistance.

julianbrost:
Can you please stop doing random things. Yes, I'm aware that there are limitations with the current implementation, but now doing things without any coordination between @nilmerg and me for such core mechanism makes little sense.

And now this ist getting quite messy here:
20250424_13h22m16s_grim
(Not to mention that you've also closed #237 in the process.)

Al2Klimov:
* Not necessary (to workaround this) anymore, as https://github.com/Icinga/icinga-notifications-web/pull/329 won't happen.

PR #293, opened 2025-03-12
internal/object: Fix gosec G115

### LINKED ISSUE #289: GitHub Actions are broken
https://github.com/Icinga/icinga-notifications/actions/runs/13787470238/job/38558797072

oxzi:
The conversion of an int64 to an uint64 was marked as a potential overflow, since it might lose precision. However, in this case this should be of no issue since only the binary representation matters, not the actual numeric value.

So a "nolint" tag was added next to an explanatory comment.

Fixes #289.


oxzi:
Superseded by #305.

PR #305, opened 2025-04-16
golangci-lint: Address issues and config

### LINKED ISSUE #289: GitHub Actions are broken

oxzi:
Within the last months, some of the linters used by golangci-lint started detecting more issues, resulting in reports for Icinga Notifications. Some errors were addressed, others marked as false positives, and a .golangci.yml configuration was created with a list of additional linters selected by trusting my gut.

---

Fixes #289.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,23 @@
+version: "2"
+
+linters:
+  default: standard
+
+  enable:
+    - bidichk # dangerous Unicode chars
+    - bodyclose # HTTP response body is closed
+    - exptostd # functions from golang.org/x/exp/ that can be replaced by std functions
+    - forcetypeassert # forced type assertions
+    - gosec # security problems
+    - loggercheck # key value pairs for common logger libraries (including zap)
+    - misspell # commonly misspelled English words
+    - nosprintfhostport # misuse of Sprintf to construct a host with port in a URL
+    - rowserrcheck # Rows.Err of rows is checked
+    - sqlclosecheck # sql.Rows, sql.Stmt, sqlx.NamedStmt, pgx.Query are closed
+
+  settings:
+    staticcheck:
+      checks:
+        - all
+        - '-ST1000' # ignore missing package comments
+        - '-ST1003' # ignore capitalization in camel case names
--- a/internal/object/object.go
+++ b/internal/object/object.go
@@ -265,9 +265,15 @@ func (o *Object) EvalExists(key string) bool {
 	return ok
 }
 
+// ID generates a stable identifier based on a source ID and tags.
+//
 // TODO: the return value of this function must be stable like forever
 func ID(source int64, tags map[string]string) types.Binary {
 	h := sha256.New()
+
+	if source < 0 {
+		panic(fmt.Sprintf("source value %d is negative", source))
+	}
 
 	sourceBytes := make([]byte, 8)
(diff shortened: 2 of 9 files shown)


.golangci.yml
```diff
@@ -0,0 +1,24 @@
...
+    - rowserrcheck # Rows.Err of rows is checked
+    - sqlclosecheck # sql.Rows, sql.Stmt, sqlx.NamedStmt, pgx.Query are closed
+
+  settings:
+    staticcheck:
+      checks:
+        - all
+        - '-ST1000' # missing package comment
+        - '-ST1003' # capitalization in camel case names
+        - '-ST1005' # capitalized error strings
```
yhabteab, line 24: I wouldn't disable this check but fix these reports instead:
```bash
internal/config/verify.go:124:11: ST1005: error strings should not be capitalized (staticcheck)
			return fmt.Errorf("Addresses[%d] is nil", i)
			       ^
internal/config/verify.go:128:11: ST1005: error strings should not be capitalized (staticcheck)
			return fmt.Errorf("Addresses[%d] has ContactID = %d instead of %d", i, address.ContactID, id)
			       ^
internal/config/verify.go:133:11: ST1005: error strings should not be capitalized (staticcheck)
			return fmt.Errorf("Addresses[%d]: %w", i, err)
			       ^
```

Apart from that, I think it would be clear if the comments of the above disabled rules were prefixed with `disable ...` , so that it's clear that this is an exclusion list and not includes.

internal/object/object.go
```diff
@@ -265,12 +266,22 @@ func (o *Object) EvalExists(key string) bool {
...
 	h := sha256.New()
 
+	// Ensure source does not overflow when converted to uint64.
+	var uSource uint64
+	if source >= 0 {
+		uSource = uint64(source)
+	} else {
+		uSource = math.MaxInt64 + uint64(-source)
+	}
+
```
yhabteab, line 282: Wouldn't it make sense to just panic when a negative source ID is given? I mean, I don't think there will ever be a negative source ID, since this an _auto increment_ value that is not allowed to be a negative number.

.golangci.yml
oxzi: Stating that these rules are ignored/disabled is a good idea. I am going to alter the errors then.

internal/object/object.go
oxzi: Again, we are dealing with different types in different domain.

| Domain | Type | Min | Max |
|--------|--------|--------|--------|
| Go | `int64` (signed) | $-2^{63}$ | $2^{63}-1$ |
| MySQL schema | `bigint` (signed) | $-2^{63}$ | $2^{63}-1$ |
| PostgreSQL schema | `bigserial` (unsigned) | $1$ (not zero!) | $2^{63}-1$ |

Especially PostgreSQL's `bigserial` left me a bit puzzled, being unsigned but not using all eight bytes and starting at one. Fortunately, no SQL type would overflow Go's `int64`. However, in theory there might be a negative value stored in MySQL, resulting in the overflow.

However, as you have said, the value should be auto incremented and in this case something went horrible wrong or someone messed with the database by hand. I am going to change this to a `panic`.

.golangci.yml
oxzi: @yhabteab: Please take another look.

PR #333, opened 2025-08-28
Incremental Sync: Fix Rule Recipients Deletion

### LINKED ISSUE #330: Notifications are not sent to all recipients of an event rule
I have an event rule which defines no filter and escalation condition but only recipients.  Therefore, notifications should be sent to all recipients, but this is not the case.

## To Reproduce:
Given contacts: Jack, Rick and Tom (Jack and Rick have email as their default channel and have an email address specified, Rick has a webhook: WhatsApp, no address specified).

1. Create a new rule without filter and escalation condition, with all specified contacts as recipients (use default channel).
Image

2. From now on, a notification will be sent to the specified contacts for all new incidents (regardless of whether they have failed or succeeded. Example below).
Image

3. Now change the recipients of the rule so that the e-mail channel is used instead of the default channel.
Image

4. From now on, notifications for new incidents will no longer be sent to all specified contacts. In my case, they are only sent to Tom, for whom no email address has been specified.
Image

The logs show that the notifications are only sent to Tom.

oxzi:
The deletion function for escalation rule recipients cleared all recipients in the escalation object based on the escalation ID, not the recipient ID. Thus, if one escalation recipient was either deleted or altered, all recipients were removed from the escalation.

This change fixes the incorrect deletion filter.

In addition, I have looked for similar errors at other incrementalApplyPending calls. Luckily, I have found none.

Fixes #330.

### DIFF
--- a/internal/config/rule.go
+++ b/internal/config/rule.go
@@ -107,7 +107,7 @@ func (r *RuntimeConfig) applyPendingRules() {
 			}
 
 			escalation.Recipients = slices.DeleteFunc(escalation.Recipients, func(recipient *rule.EscalationRecipient) bool {
-				return recipient.EscalationID == delElement.EscalationID
+				return recipient.ID == delElement.ID
 			})
 			return nil
 		})


yhabteab (approved):
LFTM!

PR #334, opened 2025-08-28
Channel: Recreate Logger After Name Change

### LINKED ISSUE #309: RuntimeConfig: Update logger fields for runtime updates
When updating a channel name within the web UI, the logs are still using the old name. This happens because the logger is created within the `createFn` but not altered/recreated in the `updateFn`.

https://github.com/Icinga/icinga-notifications/blob/860ea0ce3b74a0e8e463b3c2eccbe78fb7b58f62/internal/config/channel.go#L15-L17
https://github.com/Icinga/icinga-notifications/blob/860ea0ce3b74a0e8e463b3c2eccbe78fb7b58f62/internal/config/channel.go#L20-L27

This pattern should be checked for each runtime update, not only for channels.

Reported by @sukhwinder33445, thanks!

oxzi:
When restarting a channel, recreate its logger based on the current fields since they may have changed due to reconfiguration.

As a related change, use the Channel.MarshalLogObject method instead of populating the fields in incrementalApplyPending's createFn handler. Unfortunately, zap caches the result from the zap.Object function call and not re-evaluates it every time the logger is used. That is the reason for recreating the logger in the Channel.Restart method.

Fixes #309.

### DIFF
--- a/internal/channel/channel.go
+++ b/internal/channel/channel.go
@@ -51,7 +51,7 @@ type newConfig struct {
 
 // Start initializes the channel and starts the plugin in the background
 func (c *Channel) Start(ctx context.Context, logger *zap.SugaredLogger) {
-	c.Logger = logger
+	c.Logger = logger.With(zap.Object("channel", c))
 	c.restartCh = make(chan newConfig)
 	c.pluginCh = make(chan *Plugin)
 	c.pluginCtx, c.pluginCtxCancel = context.WithCancel(ctx)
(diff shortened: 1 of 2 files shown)


oxzi:
> > Unfortunately, zap caches the result from the zap.Object function call and not re-evaluates it every time the logger is used.
> 
> I wouldn't call that _unfortunate_! That's a good thing!

_Unfortunate_ for us in this specific case, but yes, generally I would agree :upside_down_face:

PR #344, opened 2025-10-27
Add schedule timezone column

jrauh01:
Add a column to store the timezone the schedule is created for.

Draft until the following PR is merged:

* Blocked by: https://github.com/Icinga/icinga-notifications-web/pull/369 (This schema change is solely needed for Notification Web, so no need to merge this earlier than the corresponding PR there. On the other hand, the schema change breaks Notifications Web without that PR as is couldn't insert rows anymore as it doesn't yet know of the required column.)

### DIFF
--- a/schema/mysql/upgrades/0.2.0-schedule-timezone.sql
+++ b/schema/mysql/upgrades/0.2.0-schedule-timezone.sql
@@ -1,1 +1,12 @@
-ALTER TABLE schedule ADD COLUMN timezone text NOT NULL AFTER name;
+ALTER TABLE schedule ADD COLUMN timezone text AFTER name;
+UPDATE schedule SET timezone = (
+    SELECT entry.timezone
+    FROM timeperiod_entry entry
+    INNER JOIN timeperiod ON timeperiod.id = entry.timeperiod_id
+    INNER JOIN rotation ON rotation.id = timeperiod.owned_by_rotation_id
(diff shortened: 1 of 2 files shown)


schema/mysql/upgrades/002.sql
julianbrost: I'd just name that file something like `0.2.0-schedule-timezone.sql`.

Rationale: The idea for individual files was mostly to make our lives easier during development because there's a clear "a new upgrade file was added" during a git pull (you can just apply the whole file and don't have to pick out the new lines from the updated file). Additionally, it's supposed to avoid conflicts across PRs, but that won't be the case if we want a sequential numbering of upgrade files. So unless there's a dependency on another schema upgrade file, I think our best bet is to just use a more descriptive filename. And for the release, the individual upgrade files will get merged into a big `0.2.0.sql` anyway.

Doing it this way will allow us to bring this PR (and #216) into a state where both are ready to be merged at any time without further action needed (like renaming the upgrade file) when the corresponding web PRs are ready.

schema/mysql/upgrades/002.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE schedule ADD COLUMN timezone text;
```
julianbrost: This could additionally specify `AFTER name` to match the column order of the full schema. (PostgreSQL doesn't provide something like that though.)

PR #346, opened 2025-11-06
incident: reorder `[Un]Mute` & notifications history generation

yhabteab:
So that the incident `muted` history appears logically after the just generated notifications, we must insert the `muted` history last. This way, the history entries will make sense when listed in chronological order in the UI. The `unmute` history entry, on the other hand, must be inserted first, so that the notifications generated below it appear logically after the `unmute` event. This way, when viewing the incident history in the UI, first the `unmute` event will appear and then notifications that were sent after unmuting.

This PR is supposed to allow sending non-state notifications but since this functionality is already implemented, no further changes are necessary here.

refs #106

### DIFF
--- a/internal/incident/incident.go
+++ b/internal/incident/incident.go
@@ -164,11 +164,6 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event) error {
 		return err
 	}
 
-	if err := i.handleMuteUnmute(ctx, tx, ev); err != nil {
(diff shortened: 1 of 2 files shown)


julianbrost:
Just for clarification if I understand this correctly: This PR only changes the order in which things appear the history, but has no effect on the overall notifications sent? So with or without this PR, the same notifications would be sent?

julianbrost:
#106 still describes more than what's done so far and this shouldn't be lost in a closed issue, hence I've removed the "resolves" reference. Though that issue is more of a collection of ideas, it might be a good idea to split it into multiple ones for the individual ideas.

PR #354, opened 2025-11-12
Remove leftover Icinga 2 source references

oxzi:
Replace mentions of Icinga 2 sources with Icinga DB and remove the now unused api-timeout configuration option from the Icinga 2 API client.


yhabteab (dismissed):

oxzi:
I have updated the architecture diagram as well. @yhabteab, would you please take a second look? Thanks.

PR #361, opened 2025-11-18
Listener: add /incidents endpoint

julianbrost:
Add an API endpoint for sources to query all incidents belonging to them. This allows sources to detect incidents for obsolete objects, i.e. objects that disappeared or were deleted while they had an ongoing incident.

Added primarily for Icinga for Kubernetes, but could be useful for other sources as well.


oxzi (approved):
Went over the diff and tested a build in a test setup. Looks fine.

If this works for you, @lippserd, let's merge it.

PR #416, opened 2026-04-15
golangci-lint/sqlclosecheck: Address false positive

oxzi:
With a recent golangci-lint update, the bundled sqlclosecheck version was bumped to v0.6.0, resulting in false positives when calling *DB.QueryContext.

https://github.com/ryanrolds/sqlclosecheck/issues/43

For the moment, a nolint annotation was added to please the CI.


lippserd (commented):
Could you add a brief justification even though there's a link to the issue?

oxzi:
@lippserd: I have added a short comment stating that this is a false positive, failing to find the deferred close operation.

yhabteab (approved):

PR #418, opened 2026-04-30
Evaluate rules on our own

### LINKED ISSUE #406: Evaluate rules on our own again, or in tandem
Transferring SQL Queries is… sub-optimal. Let's get rid of that.

Version 0.1.0 expected sources to include `extra_tags` so that rules match events with corresponding recipients. We want to get back to this behavior, but we also want to avoid their previous shortcomings.

The properties of an event that are relevant in 0.2.0 are:

```
{
  "extra_tags": <object>,
  "rules_version": <hash>,
  "rule_ids": <list-of-ints>
}
```

These are fully obsolete now and only a single new one is required:

```
{
  "relations": <object>
}
```

Example, in case of Icinga DB:

```json
{
  "relations": {
    "host": {
      "name": "web-1-a",
      "vars": {
        "os": "Linux"
      }
    },
    "hostgroups": [
      {
        "name": "web-servers",
        "display_name": "Web Servers"
      }
    ]
  }
}
```

The basic assumption here is that *relations* always have the same structure, so there is no variation requiring conditional checks. Let's take a look at a service why that matters:

```json
{
  "relations": {
    "host": {
      "name": "web-1-a",
      "vars": {
        "os": "Linux"
      }
    },
    "hostgroups": [
      {
        "name": "web-servers",
        "display_name": "Web Servers"
      }
    ],
    "services": [
      {
        "name": "nginx",
        "vars": {
          "http_ports": [80, 443]
        }
      }
    ],
    "servicegroups": []
  }
}
```

Note that this simply contains an additional `services` and `servicegroups` key, `services` with just a single entry. In case of a service event, it is always a list with a single entry. It can simply be extended to a multiple of services in case of a host and the structure didn't change.

Icinga Notifications Web will serialize rule filters in a format easy to parse for Icinga Notifications. Though, if that's not the query string syntax already in use, Web will need to be able to parse it back as well in order to present it back to the user.

Upon retrieval of an event, Icinga Notifications tries to match all rules with it. If any rule cannot match due to missing attributes, they're collected in a list. If this happens, Icinga Notifications needs to contact the source and tell it about those and the source needs to re-transmit either the full event again or updated relations. (That's up to the implementer to decide.) This is called *attribute negotiation*.

> [!IMPORTANT]
> _Generic_ sources are not meant to support *attribute negotiation*, so whether a source is capable of this must be signaled somehow to Icinga Notifications. A request header for example.

The source may signal that some relations are already complete and an update cannot be delivered for them. This is done by using a request header: `X-Icinga-Complete-Relations: hostgroups, services[*].vars`

If the source cannot (will not or is unable to) provide an update, or the the update is incomplete, Icinga Notifications evaluates the rules as far as it can but generates a log entry for rules that couldn't be evaluated.

Let's say Icinga DB missed to include the hostgroups and service custom variables. In this case, Icinga Notifications will ask it for:
- `services[*].vars.http_ports`
- `hostgroups`

And if a group did not contain the display name:
- `hostgroups[*].display_name`

The format used here is JSONpath. Its mature nature and widespread use should pose no problem for any source to understand.

The structure of said request and the expected response _may_ look like this:

Request:
```json
{
  "type": "attribute negotiation",
  "attributes": [
     "hostgroups",
     "services[*].vars.http_ports"
  ]
}
```

Response:
```json
{
  "relations": {
    "hostgroups": [
      {
        "name": "web-servers",
        "display_name": "Web Servers"
      }
    ],
    "services": [
      {
        "vars": {
          "http_ports": [80, 443]
        }
      }
    ]
  }
}
```

yhabteab:
This PR kind of reverts all the changes mad in v0.2.0 with a slight twist. We will evaluate the rules on our own again (just like pre v0.2.0) but the filter string is expected to have a different format. Icinga Notifications Web will now store the filter string as a JSON stringified object instead of a simple filter string format used before. This is done allegedly to make it easier to parse the filter string but the previous filter string parser is still there as it's used by the escalation filters. The new filter string format is as follows:

```json
{
  "op": "&/|/!",
  "rules": [
    {
      "column": "service.state",
      "op": "==/!=/>/</~/!~",
      "value": "OK"
    },
    {
      "op": "&/|/!",
      "rules": [
        ...
      ]
    }
  ]
}
```

The JSON object for filter conditions may include other meta information used by Icinga Notifications Web/Sources but the only relevant information for us is the above-mentioned format. The `op` field is the operator used to combine the filter chain rules and can be either `&` (AND), `|` (OR) or `!` (NOT). The `rules` field is an array of rules which can be either a simple condition or a nested filter chain. A simple condition consists of a `column`, an `op` and a `value`. The `column` is the column name used to extract values from requests as JSONPath expressions, the `op` is the operator used to compare the JSONPath expression result with the provided value and can be either `==`, `!=`, `>`, `<`, `~` (regex match) or `!~` (regex not match). If the `column` is not a valid JSONPath expression, Icinga Notifications will just ignore that rule and won't be loading it from the database. This means that if you have an existing filter string in the old format, they won't be loaded, and you will have to ask the web colleagues if they provide a migration step for the existing filter strings.

Furthermore, prior to v0.2.0, we were evaluating the rules against the `Object` representation of the request, now they are evaluated against the `Event` type itself. Specifically, the JSONPath expressions are evaluated against the new `relations` field of the `Event` type which is a map of generic key-value pairs extracted from the request. When a source doesn't include that field with all the necessary information in its requests, some or even all the configured rules might not match if they reference the missing information in their `object_filter`. If the source wishes to be notified about the missing information, it can include the new `X-Icinga-Enable-Attributes-Negotiation` HTTP header set to `true` in its requests and Icinga Notifications will respond with a `422 Unprocessable Entity` status code (not final, is still up to a discussion) and a JSON body containing the missing attributes. This is a new feature but similar to the previous behavior introduced in v0.2.0 where we were responding with a `412 Precondition Failed` status code when the used rules_version was outdated. The source can then extend its requests with the missing information and retry the request.

In the referenced issue, it was mentioned that sources might only re-send the missing information as a follow-up request instead of retrying the entire request. However, it's not worth implementing this behavior just for the sake of this single use case as it would require a lot of additional work and complexity on both the source and Icinga Notifications sides.

Lastly, in a discussion with @nilmerg yesterday, we agreed that Icinga Notifications should abandon all the effects of the ongoing request if it can't successfully notify the source that the request was successfully processed. Currently, the incident starts a database transaction internally and commits it before trying to send the resulting notifications. However, the client will only receive `200 OK` if the notifications were sent successfully, and that in turn might not reach the source since the connection might be gone in the meantime (errors are ignored blantantly).

https://github.com/Icinga/icinga-notifications/blob/aad8cb253253742bc2a935a59f1b99dee82eab27/internal/listener/listener.go#L181-L185

https://github.com/Icinga/icinga-notifications/blob/aad8cb253253742bc2a935a59f1b99dee82eab27/internal/incident/incident.go#L221-L229

In that case the source will be left in a state where the request was processed but the source wasn't notified about it, which is not ideal and will definitely lead to some issues when we've added HA support. Icinga Notifications will treat every event as a unique one and won't have any mechanism to detect duplicate events, so if the source retries the request, it will be processed again and might lead to some unwanted side effects. To avoid that, we will need to roll back all the changes made by the request if we fail to notify the source about the successful processing of the request. ~I didn't implement this behavior in this PR as I see it as a separate task that can be done in a follow-up PR, but if you think it should be included in this PR, please let me know and I will add it.~ Forget that, that's really something that should be done after or while implementing the HA support.

resolves #406


yhabteab:
# TODO

- [ ] Drop the last `go.mod` bump commit once https://github.com/Icinga/icinga-go-library/pull/199 is merged and released.

nilmerg:
> `~` (regex match) or `!~` (regex not match).

Web doesn't handle those as regex capable operators and what users are using as expressions are simple glob like wildcards only. So if this interprets this as regex it will certainly fail and we have to make sure Web also serializes this as regex.

internal/config/rule.go
```diff
@@ -109,36 +66,25 @@ func (r *RuntimeConfig) applyPendingRules() {
 				curElement.TimePeriod = nil
 			}
 
-			if curElement.SourceID != update.SourceID {
-				delFromRulesBySource(curElement)
-				curElement.SourceID = update.SourceID
-				addToRulesBySource(curElement)
-			}
```
oxzi, line 116: Unsure, but is it still possible to change the source ID of a rule now anymore?

internal/config/verify.go
```diff
@@ -248,6 +248,10 @@ func (r *RuntimeConfig) debugVerifyRule(id int64, rule *rule.Rule) error {
 		}
 	}
 
+	if rule.ObjectFilterExpr.Valid && rule.ObjectFilter == nil {
+		return fmt.Errorf("rule has a ObjectFilterExpr but ObjectFilter is nil")
```
oxzi:
```suggestion
		return fmt.Errorf("rule has an ObjectFilterExpr but ObjectFilter is nil")
```

internal/event/event.go
```diff
@@ -120,6 +123,90 @@ func (e *Event) Sync(ctx context.Context, tx *sqlx.Tx, db *database.DB, objectId
...
+
+func (e *Event) EvalLike(key string, value string) (bool, error) {
+	var builder strings.Builder
+	builder.WriteRune('^')
+	for segment := range strings.SplitSeq(value, "*") {
+		if segment == "" {
+			builder.WriteString(".*")
+		}
+		builder.WriteString(regexp.QuoteMeta(segment))
+	}
```
oxzi: The `segment == ""` does not work for a wildcard within a string.

```diff
diff --git a/internal/event/event_test.go b/internal/event/event_test.go
index 2aa5379..2e31e2a 100644
--- a/internal/event/event_test.go
+++ b/internal/event/event_test.go
@@ -52,6 +52,7 @@ func TestFilter(t *testing.T) {
                // ... expected positive matches
                {Expr: makeJsonFilterExpr(t, "host.name", "=", "test-host"), Expected: true},
                {Expr: makeJsonFilterExpr(t, "host.name", "~", "test*"), Expected: true},
+               {Expr: makeJsonFilterExpr(t, "host.name", "~", "test*host"), Expected: true},
                {Expr: makeJsonFilterExpr(t, "host.vars.dict.key", "=", "value"), Expected: true},
                {Expr: makeJsonFilterExpr(t, "host.vars.dict.key", "!~", "something*"), Expected: true},
                {Expr: makeJsonFilterExpr(t, "host.vars.dict.key_int", ">=", 42), Expected: true},
```

```
--- FAIL: TestFilter (0.00s)
    event_test.go:88:
                Error Trace:    /home/apenning/git/github.com/Icinga/icinga-notifications/internal/event/event_test.go:88
                Error:          Not equal:
                                expected: true
                                actual  : false
                Test:           TestFilter
                Messages:       unexpected result for "{\"column\":\"host.name\",\"op\":\"~\",\"value\":\"test*host\"}"
FAIL
```


```suggestion
	isFirst := true
	for segment := range strings.SplitSeq(value, "*") {
		if !isFirst {
			builder.WriteString(".*")
		}
		isFirst = false
		builder.WriteString(regexp.QuoteMeta(segment))
	}
```

internal/event/event.go
```diff
@@ -120,6 +123,90 @@ func (e *Event) Sync(ctx context.Context, tx *sqlx.Tx, db *database.DB, objectId
...
+	return missing
+}
+
+func (e *Event) EvalEqual(key string, value string) (bool, error) {
+	return slices.ContainsFunc(e.retrieveValuesFor(key), func(v any) bool { return fmt.Sprint(v) == value }), nil
+}
+
+func (e *Event) EvalLess(key string, value string) (bool, error) {
+	return slices.ContainsFunc(e.retrieveValuesFor(key), func(v any) bool { return fmt.Sprint(v) < value }), nil
```
oxzi: A string comparison might result in unexpected results. For example, the string `42` is considered less than `5` as `4<5`. Same applies to `EvalLessOrEqual`.

```diff
diff --git a/internal/event/event_test.go b/internal/event/event_test.go
index 2aa5379..ce22def 100644
--- a/internal/event/event_test.go
+++ b/internal/event/event_test.go
@@ -55,6 +55,7 @@ func TestFilter(t *testing.T) {
                {Expr: makeJsonFilterExpr(t, "host.vars.dict.key", "=", "value"), Expected: true},
                {Expr: makeJsonFilterExpr(t, "host.vars.dict.key", "!~", "something*"), Expected: true},
                {Expr: makeJsonFilterExpr(t, "host.vars.dict.key_int", ">=", 42), Expected: true},
+               {Expr: makeJsonFilterExpr(t, "host.vars.dict.key_int", ">=", 5), Expected: true},
                {Expr: makeJsonFilterExpr(t, "host.vars.dict.key_int", "=", 42.0), Expected: true},
                {Expr: makeJsonFilterExpr(t, "host.vars.dict.key_array[0]", "!=", "value2"), Expected: true},
                {Expr: makeJsonFilterExpr(t, "host.vars.dict.key_array[1]", "!=", "value1"), Expected: true},
```

```
--- FAIL: TestFilter (0.00s)
    event_test.go:88:
                Error Trace:    /home/apenning/git/github.com/Icinga/icinga-notifications/internal/event/event_test.go:88
                Error:          Not equal:
                                expected: true
                                actual  : false
                Test:           TestFilter
                Messages:       unexpected result for "{\"column\":\"host.vars.dict.key_int\",\"op\":\"\\u003e=\",\"value\":5}"
```

yhabteab:
> > `~` (regex match) or `!~` (regex not match).
> 
> Web doesn't handle those as regex capable operators and what users are using as expressions are simple glob like wildcards only. So if this interprets this as regex it will certainly fail and we have to make sure Web also serializes this as regex.

Notifications transforms it into a regexp, so you don't need to change anything regarding this on the web side.

internal/config/rule.go
yhabteab: I'm not sure what that means, was it ever allowed to change after creation? Of course, if you manually edit the database entry then yes, but I don't see how one is allowed to move rules from one source to another from within Notifications Web. Though, adding a simple check shouldn't harm either, so I'll add it.

oxzi: Sorry, I am now a bit confused myself. However, shouldn't it be possible to change an event rule's source as shown in the screenshot below?

Screenshot 2026-05-11 at 10-15-55 Event Rule Icinga DB Catch-All Icinga Web

internal/event/event.go
yhabteab: That's not something new since the previous implementation had also this very same problem and I didn't find any better way than JSON encoding the values and using `bytes.Compare` isntead. If that's something you can live with then I can push these changes.

internal/event/event.go
yhabteab: Fixed.

internal/config/verify.go
yhabteab: Fixed.

internal/config/rule.go
yhabteab: Ok, re-added the check again.

nilmerg:
> Notifications transforms it into a regexp, so you don't need to change anything regarding this on the web side.

Ah, but don't you think it's better then to leave this up to how the serialization is actually performed? I don't quite like the idea to require the daemon to transform the value of a condition prior interpretation. We may even serialize like/unlike to equal/unequal with `pattern` or `regex` instead of `value` in order to further disambiguate the format.

yhabteab:
> > Notifications transforms it into a regexp, so you don't need to change anything regarding this on the web side.
>
> Ah, but don't you think it's better then to leave this up to how the serialization is actually performed?  don't quite like the idea to require the daemon to transform the value of a condition prior interpretation.

The transformation exists because of the previous filter string parsing implementation. If you think it's better to leave it up to the serialization, then it's fine for me too and I don't mind changing it.

> We may even serialize like/unlike to equal/unequal with `pattern` or `regex` instead of `value` in order to further disambiguate the format.

How would merging the like/unlike to equal/unequal disambiguate something? In fact that would make it even more complex for us to distinguish when to and not to use regexp. Compiling regexps is expensive, so equal/unequal are used for simple and plain comparisons unconditionally, while like/unlike are used for regexps. If you merge them, then I would have to add some kind of hacky workaround to determine whether I should use regexp based on the value of the condition, which is not ideal.

nilmerg:
> If you merge them, then I would have to add some kind of hacky workaround to determine whether I should use regexp based on the value of the condition, which is not ideal.

It's not a must. I just thought that if we allow regular expressions, why not make it explicit and remove the operator distinction. That distinction was only necessary for Web due the query-string serialization to get rid of the error-prone inspection of whether the value contained an asterisk or not. That's why I also mentioned a different key other than `value`. Your parser just has to take `value` as the plain expression and e.g. `regex` as a ready-to-use pattern.

yhabteab:
> That's why I also mentioned a different key other than `value`. Your parser just has to take `value` as the plain expression and e.g. `regex` as a ready-to-use pattern.

Ah, ok, that should work just fine then!

internal/event/event.go
oxzi: You have a point, this is not something we can fix easily. I like your current approach with the `CompareJsonEncode` helper function, but comparing byte sizes will have its downside, e.g., for floats where 3.1415 will be greater than 9001.

However, if it is fine for you, we can create this as a new issue and don't consider this a blocker for this issue.

internal/config/rule.go
```diff
@@ -110,35 +67,32 @@ func (r *RuntimeConfig) applyPendingRules() {
 			}
 
 			if curElement.SourceID != update.SourceID {
-				delFromRulesBySource(curElement)
+				if src, ok := r.Sources[curElement.SourceID]; ok {
+					src.deleteRuleID(curElement.ID)
+				}
+				if src, ok := r.Sources[update.SourceID]; ok {
+					src.appendRuleID(curElement.ID)
```
oxzi, line 73:
```suggestion
					src.appendRuleID(update.ID)
```

internal/event/event.go
```diff
@@ -120,6 +123,111 @@ func (e *Event) Sync(ctx context.Context, tx *sqlx.Tx, db *database.DB, objectId
...
+	var missing []string
+	for _, filterColumn := range filterColumns {
+		if _, cached := e.evaluatedRelations[filterColumn]; cached {
+			continue
+		}
+		// This should never panic, as the filter columns have already been validated when loading the rules.
+		path := jpp.MustParse(utils.PrefixWithJSONPathRootSelector(filterColumn))
+		if nodes := path.Select(e.Relations); len(nodes) == 0 {
+			isComplete := slices.ContainsFunc(e.CompleteRelations, func(relation string) bool {
+				return strings.HasPrefix(filterColumn, relation)
```
oxzi: While scrolling over this code a second time, I am not quite sure if this is not too loose. Aren't there possibilities with overlapping prefixes, such as `host` and `hostgroup`?

internal/config/rule.go
yhabteab: Does it make any difference? The rule ID remains the same even if its source ID changes.

internal/event/event.go
yhabteab: > but comparing byte sizes will have its downside, e.g., for floats where 3.1415 will be greater than 9001.

Overseen the 🐘! Ok, I've really tried my best to avoid using `reflect` package for this, but there's no way around it, so I've changed the implementation to a simple reflect based one which seems to work just fine.

internal/event/event.go
yhabteab: Fixed.

internal/config/rule.go
oxzi: Fair enough. It just looked asymmetrical to me. Sorry.

oxzi (dismissed):
Thanks, LGTM.

I went over the diff multiple times and it seems to be fine. Of course, there might be some things I am missing due to the PR's size, but I would say let's merge it!

yhabteab:
Due to https://github.com/Icinga/icinga-notifications/issues/406#issuecomment-4429854963, I had to refactor the `Filterable` interface a bit, so that a `filter.Condition` can provide multiple columns/JSONPaths to them and not just a single column of type string (so the PR got a bit bigger again due to this). Also, the request handling on missing relations has changed in so far, that Notifications will only reject it if the client explicitly requested to do so via the `XIcingaRejectIfRelationsIncomplete` HTTP header, or the source doesn't have any event rules without an object filter and the ongoing event doesn't cause a new incident to be created and there's no existing one yet. Please have a look at it again!

That's not final though, since I'm going to drop the wildard -> regex transformation of filter columns for the `Like` and `Unlike` operators, once it's implemented in Icinga Notifications Web.

yhabteab:
> That's not final though, since I'm going to drop the wildard -> regex transformation of filter columns for the `Like` and `Unlike` operators, once it's implemented in Icinga Notifications Web.

Done, now! See https://github.com/Icinga/icinga-notifications/compare/8aa3308299da355a58c36d7fceff2fd20b0e3fe8..4091faecea0cbd8a1ca03e62babb9455a4027c7c

yhabteab:
I've transformed if/elseif statements to switch case to make the CI happy 😢!

yhabteab:
My last commit makes sure that we don't include the now huge `object_filter` string of event rules like the following in the logs:
```json
{"qs":"service.usergroup.name=icingaadmins&service.name!=http&servicegroup.name~%2Ad%2A&host.name~%2Aicinga%2A&hostgroup.name=linux-servers&host.user.name!=icingaadmin&host.vars.http_vhosts.http.http_uri=%2F&service.vars.disk_partitions!=%2F&host.vars.disks.disk%20%2F.disk_partitions%3E=%2F&host.vars.disks.disk!~%2A&host.vars.notification.mail.groups%5B%2A%5D=icingaadmins","ast":{"op":"&","rules":[{"op":"=","attributes":["services[*].usergroup.name"],"value":"icingaadmins"},{"op":"!=","attributes":["services[*].name"],"value":"http"},{"op":"=","attributes":["servicegroups[*].name"],"regex":"^.*d.*$"},{"op":"=","attributes":["host.name"],"regex":"^.*icinga.*$"},{"op":"=","attributes":["hostgroups[*].name"],"value":"linux-servers"},{"op":"!=","attributes":["host.user.name"],"value":"icingaadmin"},{"op":"=","attributes":["host.vars['http_vhosts']['http']['http_uri']"],"value":"\/"},{"op":"!=","attributes":["services[*].vars.disk_partitions"],"value":"\/"},{"op":">=","attributes":["host.vars['disks']['disk \/']['disk_partitions']","host.vars['disks']['disk']"],"value":"\/"},{"op":"!=","attributes":["host.vars[\"disks\"]['disk \/']['disk_partitions']","host.vars['disks']['disk']"],"regex":"^.*$"},{"op":"=","attributes":["host.vars['notification']['mail']['groups'][*]"],"value":"icingaadmins"}]}}
```

internal/listener/listener.go
```diff
@@ -347,35 +341,40 @@ func (l *Listener) DumpRules(w http.ResponseWriter, r *http.Request) {
...
-			rulesInfo.Rules = make(map[string]string)
-
-			for _, rID := range sourceInfo.RuleIDs {
-				id := strconv.FormatInt(rID, 10)
-				filterExpr := ""
-				if l.runtimeConfig.Rules[rID].ObjectFilterExpr.Valid {
-					filterExpr = l.runtimeConfig.Rules[rID].ObjectFilterExpr.String
-				}
+	resp := map[string]any{
+		"type":       "attrs_negotiation",
```
oxzi: Both in Icinga/icingadb#1098 and in #406, the type is called `attribute negotiation`. Shouldn't it be the same here as well?

yhabteab: Fixed.

yhabteab:
Fixed the requested change and updated the go.mod file to now require the `develop` branch of IGL.

doc/20-HTTP-API.md
```diff
@@ -19,14 +19,14 @@ The authentication is performed via HTTP Basic Authentication using the source's
     When upgrading a setup from an earlier version, these usernames are still valid, but can be changed in Icinga Notifications Web.
 
 Events sent to Icinga Notifications are expected to match rules that describe further event escalations.
-These rules can be created in the web interface.
-Next to an array of `rule_ids`, a `rules_version` must be provided to ensure that the source has no outdated state.
+These rules can be configured in Icinga Notifications Web and should be designed to match the `relations` of the
+submitted events. When submitting an event without the expected relations to evaluate the rules, Icinga Notifications
+will reject the request with a `422 Unprocessable Entity` status code and a message describing the missing relations
+when the `XX-Icinga-Reject-If-Relations-Incomplete` header is set to `true`. Otherwise, the request will be accepted
```
oxzi:
```suggestion
when the `X-Icinga-Reject-If-Relations-Incomplete` header is set to `true`. Otherwise, the request will be accepted
```

https://github.com/Icinga/icinga-go-library/blob/a12c30adeab77229c597d9720a8f920faf5ce036/notifications/http_headers.go#L10

internal/event/event_test.go
```diff
@@ -0,0 +1,155 @@
...
+			{Expr: makeJsonFilterExpr(t, []string{"bar.foo", "host.name"}, "=", "^.*host$", true), Expected: true},
+			{Expr: makeJsonFilterExpr(t, []string{"something.wrong", "host.name"}, "=", "^test.*host$", true), Expected: true},
+			{Expr: makeJsonFilterExpr(t, []string{"host.vars.dict.key", "invalid["}, "=", "^value$", true), Expected: true},
+			{Expr: makeJsonFilterExpr(t, []string{"host.vars.dict.key"}, "!=", "^something.*$", true), Expected: true},
+			{Expr: makeJsonFilterExpr(t, []string{"host.vars.dict.key_int"}, ">=", 42, false), Expected: true},
+			{Expr: makeJsonFilterExpr(t, []string{"host.vars.dict.key_int"}, "<=", 42, false), Expected: true},
+			{Expr: makeJsonFilterExpr(t, []string{"host.vars.dict.key_int"}, ">=", 5, false), Expected: true},
+			{Expr: makeJsonFilterExpr(t, []string{"host.vars.dict.key_int"}, "=", 42.0, false), Expected: true},
+			{Expr: makeJsonFilterExpr(t, []string{"host.vars.dict.key_float"}, "<", 3.17, false), Expected: true},
+			{Expr: makeJsonFilterExpr(t, []string{"host.vars.dict.domain"}, ">", "foo.com", false), Expected: true},
```
oxzi: This testcase seems to only works by accident. The current logic first compares the length, than the content. As `example.com` is longer than `foo.com`, this works.

internal/event/event.go
```diff
@@ -120,6 +123,140 @@ func (e *Event) Sync(ctx context.Context, tx *sqlx.Tx, db *database.DB, objectId
...
+
+	// completePaths caches the parsed JSONPath expressions of the complete relations.
+	completePaths := map[string]*jsonpath.Path{}
+	var result []string
+filterColumnsLoop:
+	for _, columns := range filterColumns {
+		var missing []string
+		for _, filterColumn := range columns {
+			if _, cached := e.evaluatedRelations[filterColumn]; cached {
+				continue
```
oxzi: I am a bit unsure about this one, sorry. Do you might want to `continue filterColumnsLoop` here for the same reasons as commented above the `continue filterColumnsLoop` below in line 172?

internal/event/event.go
```diff
@@ -120,6 +123,140 @@ func (e *Event) Sync(ctx context.Context, tx *sqlx.Tx, db *database.DB, objectId
 	return err
 }
 
+// ExtractMissingRelations determines which of the given filter columns are missing in the Relations field of this event.
+//
+// It evaluates the filter columns as JSONPath expressions against the Relations field and returns
+// a list of filter columns that do not have any matching nodes in the Relations field and are not
+// part of the CompleteRelations field. For filter columns that do have matching nodes, it caches
+// the evaluated nodes for potential later use during rules evaluation.
+func (e *Event) ExtractMissingRelations(filterColumns ...[]string) []string {
```
oxzi, line 132: As we have just discussed offline, it might be easier for sources if the missing relations are absolute, starting with `$.`.

internal/utils/utils.go
```diff
@@ -53,3 +56,49 @@ func ForEachRow[Row, Id any](ctx context.Context, db *database.DB, idColumn stri
...
+			}
+			// Both strings have the same length, compare them lexicographically.
+			return strings.Compare(av, bv), nil
+		}
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		if btype.ConvertibleTo(atype) {
+			return cmp.Compare(reflect.ValueOf(a).Int(), reflect.ValueOf(b).Convert(atype).Int()), nil
+		}
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+		if atype.ConvertibleTo(btype) {
```
oxzi:
```suggestion
		if btype.ConvertibleTo(atype) {
```

Unsure how much the implication `a` is convertible to `b` says about `b.Convert(atype)`.

Same below for the floating point case.

internal/event/event.go
yhabteab: Yep, thanks!

internal/event/event_test.go
yhabteab: It's not by accident but wanted to test just that.

internal/utils/utils.go
yhabteab: Initially, the switch case was based on `b.type` but then forgot to adjust these two checks as well.

doc/20-HTTP-API.md
yhabteab: Fixed.

internal/event/event_test.go
yhabteab: Ok, I've dropped the length check.

internal/event/event.go
yhabteab: Fixed.

internal/event/event.go
yhabteab: Fixed.

yhabteab:
Rebased and fixed the `PrefixWithJSONPathRootSelector` function to account for array like JSONPath expressions.

doc/20-HTTP-API.md
```diff
@@ -37,22 +37,40 @@ curl -v -u 'source-2:insecureinsecure' -d '@-' 'http://localhost:5680/process-ev
...
+        "display_name": "Linux Servers"
+      }
+    ],
+    "servicegroups": [
+      {
+        "name": "production-services",
+        "display_name": "Production Services"
+      }
+    ]
+  }
```
oxzi, line 86: I would expect a `complete_relations` entry here as well, maybe together with the `X-Icinga-Reject-If-Relations-Incomplete` HTTP request header in the `curl` command above.

doc/10-Channels.md
```diff
@@ -256,16 +256,6 @@ or if the channel is missing required configuration values.
...
-      },
-      "extra_tags": {
-        "hostgroup/app-mobile": "",
-        "hostgroup/department-dev": "",
-        "hostgroup/env-prod": "",
-        "hostgroup/location-rome": "",
-        "servicegroup/app-storage": "",
-        "servicegroup/department-ps": "",
-        "servicegroup/env-prod": "",
-        "servicegroup/location-rome": ""
```
oxzi, line 268: While I thought that this `object` could also require an update as for the other docs page, it turns out that the `plugins.Object` still has its `ExtraTags`, also in the `develop` branch.

https://github.com/Icinga/icinga-go-library/blob/337ee9ae3dba287459546c35ab17583f47e66605/notifications/plugin/plugin.go#L146

oxzi (commented):
Otherwise, went over the code again and would proceed with this.

doc/20-HTTP-API.md
yhabteab: Added.

doc/10-Channels.md
yhabteab: > While I thought that this `object` could also require an update as for the other docs page

Do you've any specific suggestions in mind because I don't see what else should be updated here?

> it turns out that the `plugins.Object` still has its `ExtraTags`, also in the `develop` branch.

Yep, fixed with https://github.com/Icinga/icinga-go-library/pull/203.

yhabteab:
I've resolved the conflicts caused by https://github.com/Icinga/icinga-notifications/pull/425.

yhabteab:
Re-bumped the Go mod file and changed one info log to debug.

oxzi (approved):

PR #423, opened 2026-05-13
Don't persist `events,incident_events` to DB

### LINKED ISSUE #392: Do not persist events any longer to database
The event list has already lost its significance in version 0.2.0, as it now only displays events that sources have already assigned to specific event rules. The history of an incident also contains this information. There, it is also possible to see whether and which event rule processed the event.

While event rule matching will return with #390, the remaining reasons still apply and there's no real need to persist events any longer. The event list will therefore be removed in the UI without replacement. This also applies to the tables `event` and `incident_event` in the database.

Events will only be processed once and are then thrown away. There should be a debug log of received events and how they were handled, though. (If there isn't one yet, didn't check)

yhabteab:
This PR ensures that no `event` and `incident_event` records are persisted in the database as requested in the referenced issue. Also, @julianbrost suggested in the referenced issue that the `event` database table might be used in the future for job queuing required for HA, but I don't think we will be storing a serialized form of the event in the database for that purpose, so I have removed the `event` table as well. Additionally, I have added a debug log with all the relevant fields of the request, so that one has something to look at when debugging why or why not an event rule matched or not.

resolves #392

### DIFF
--- a/internal/event/event.go
+++ b/internal/event/event.go
@@ -1,21 +1,20 @@
 package event
 
 import (
-	"context"
 	"errors"
 	"fmt"
-	"github.com/icinga/icinga-go-library/database"
-	baseEv "github.com/icinga/icinga-go-library/notifications/event"
-	"github.com/icinga/icinga-go-library/types"
-	"github.com/icinga/icinga-notifications/internal/pool"
-	"github.com/icinga/icinga-notifications/internal/utils"
-	"github.com/jmoiron/sqlx"
-	"github.com/theory/jsonpath"
 	"net/url"
 	"regexp"
 	"slices"
 	"strings"
 	"time"
+
+	baseEv "github.com/icinga/icinga-go-library/notifications/event"
+	"github.com/icinga/icinga-go-library/types"
+	"github.com/icinga/icinga-notifications/internal/pool"
+	"github.com/icinga/icinga-notifications/internal/utils"
+	"github.com/theory/jsonpath"
+	"go.uber.org/zap/zapcore"
 )
 
 // ErrSuperfluousStateChange indicates a superfluous state change being ignored and stopping further processing.
@@ -34,7 +33,6 @@ var ErrSuperfluousMuteUnmuteEvent = errors.New("ignoring superfluous (un)mute ev
 type Event struct {
 	Time     time.Time `json:"-"`
 	SourceId int64     `json:"-"`
-	ID       int64     `json:"-"`
 
 	baseEv.Event `json:",inline"`
 
@@ -104,23 +102,19 @@ func (e *Event) SetMute(muted bool, reason string) {
 	e.MuteReason = reason
 }
 
-func (e *Event) String() string {
-	return fmt.Sprintf("[time=%s type=%q severity=%s]", e.Time, e.Type, e.Severity.String())
-}
-
-// Sync transforms this event to *event.EventRow and synchronises with the database.
-func (e *Event) Sync(ctx context.Context, tx *sqlx.Tx, db *database.DB, objectId types.Binary) error {
-	if e.ID != 0 {
+// MarshalLogObject implements the [zapcore.ObjectMarshaler] interface to allow logging the event as a structured object.
+func (e *Event) MarshalLogObject(encoder zapcore.ObjectEncoder) error {
+	encoder.AddString("name", e.Name)
+	encoder.AddTime("time", e.Time)
+	encoder.AddInt64("source_id", e.SourceId)
+	encoder.AddString("type", e.Type.String())
+	encoder.AddString("severity", e.Severity.String())
+	return encoder.AddObject("tags", zapcore.ObjectMarshalerFunc(func(objectEncoder zapcore.ObjectEncoder) error {
+		for key, value := range e.Tags {
+			objectEncoder.AddString(key, value)
+		}
 		return nil
-	}
-
-	eventRow := NewEventRow(e, objectId)
-	eventID, err := database.InsertObtainID(ctx, tx, database.BuildInsertStmtWithout(db, eventRow, "id"), eventRow)
-	if err == nil {
-		e.ID = eventID
-	}
-
-	return err
+	}))
 }
 
 // ExtractMissingRelations determines which of the given filter columns are missing in the Relations field of this event.
@@ -255,35 +249,4 @@ func (e *Event) retrieveValuesFor(attrs any) jsonpath.NodeList {
 		}
 	}
 	return nil
-}
-
-// EventRow represents a single event database row and isn't an in-memory representation of an event.
-type EventRow struct {
-	ID         int64           `db:"id"`
-	Time       types.UnixMilli `db:"time"`
-	ObjectID   types.Binary    `db:"object_id"`
-	Type       types.String    `db:"type"`
-	Severity   baseEv.Severity `db:"severity"`
-	Username   types.String    `db:"username"`
-	Message    types.String    `db:"message"`
-	Mute       types.Bool      `db:"mute"`
-	MuteReason types.String    `db:"mute_reason"`
-}
-
-// TableName implements the contracts.TableNamer interface.
-func (er *EventRow) TableName() string {
-	return "event"
-}
-
-func NewEventRow(e *Event, objectId types.Binary) *EventRow {
-	return &EventRow{
-		Time:       types.UnixMilli(e.Time),
-		ObjectID:   objectId,
-		Type:       types.MakeString(e.Type.String(), types.TransformEmptyStringToNull),
-		Severity:   e.Severity,
-		Username:   types.MakeString(e.Username, types.TransformEmptyStringToNull),
-		Message:    types.MakeString(e.Message, types.TransformEmptyStringToNull),
-		Mute:       e.Mute,
-		MuteReason: types.MakeString(e.MuteReason, types.TransformEmptyStringToNull),
-	}
 }
--- a/internal/event/event.go
+++ b/internal/event/event.go
@@ -1,13 +1,11 @@
 package event
 
 import (
-	"context"
 	"errors"
 	"fmt"
-	"github.com/icinga/icinga-go-library/database"
 	baseEv "github.com/icinga/icinga-go-library/notifications/event"
 	"github.com/icinga/icinga-go-library/types"
-	"github.com/jmoiron/sqlx"
+	"go.uber.org/zap/zapcore"
 	"net/url"
 	"strings"
 	"time"
@@ -29,7 +27,6 @@ var ErrSuperfluousMuteUnmuteEvent = errors.New("ignoring superfluous (un)mute ev
 type Event struct {
 	Time     time.Time `json:"-"`
 	SourceId int64     `json:"-"`
-	ID       int64     `json:"-"`
 
 	baseEv.Event `json:",inline"`
 }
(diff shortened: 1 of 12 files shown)


yhabteab:
Resolved the conflicts caused by #425.

yhabteab:
The Icinga Notifications Web counter part to this is https://github.com/Icinga/icinga-notifications-web/pull/467 and was merged long time ago.

yhabteab:
Merged two identical logs into one to avoid the noise a bit.

yhabteab:
Rebased to resolve the conflicts.

internal/incident/incidents.go
```diff
@@ -250,29 +244,13 @@ func ProcessEvent(
...
-
-			return nil
-		case ev.Severity != baseEv.SeverityOK:
+		if ev.Severity != baseEv.SeverityOK && ev.Severity != baseEv.SeverityNone {
 			panic(fmt.Sprintf("cannot process event %v with a non-OK state %v without a known incident", ev, ev.Severity))
-		default:
+		}
+		if ev.Severity == baseEv.SeverityOK {
 			return fmt.Errorf("%w: ok state event from source %d", event.ErrSuperfluousStateChange, ev.SourceId)
 		}
```
oxzi: Consider switching the order of these two `if` statements. By checking `ev.Severity == baseEv.SeverityOk` first, you can remove this from the condition of the other `if`. Btw, the current first `if` is identical to the `CanOpenNewIncident` function.

internal/event/event.go
```diff
@@ -104,23 +102,23 @@ func (e *Event) SetMute(muted bool, reason string) {
...
+
+// MarshalLogObject implements the [zapcore.ObjectMarshaler] interface to allow logging the event as a structured object.
+func (e *Event) MarshalLogObject(encoder zapcore.ObjectEncoder) error {
+	encoder.AddString("name", e.Name)
+	encoder.AddTime("time", e.Time)
+	encoder.AddInt64("source_id", e.SourceId)
+	encoder.AddString("type", e.Type.String())
+	encoder.AddString("severity", e.Severity.String())
+	encoder.AddString("username", e.Username)
+	_ = encoder.AddObject("tags", zapcore.ObjectMarshalerFunc(func(objectEncoder zapcore.ObjectEncoder) error {
```
oxzi: Why discard this error and not just return it?

schema/pgsql/upgrades/004.sql
oxzi, line 1:
```diff
--- Kaez5eiD_postgresql_schema_import.sql       2026-06-02 09:29:26.114718207 +0200
+++ Kaez5eiD_postgresql_schema_upgrade.sql      2026-06-02 09:29:24.485720348 +0200
@@ -2,7 +2,7 @@
 -- PostgreSQL database dump
 --

-\restrict 8UXAcMqapX7ggjZghk9nJK3Ww2OKwCycN5UJVE6mSapf2XRsoRN7kUvhCKk81Rg
+\restrict olL2fDNHTH61Pyoug3U3cqu4eBp7OVM7KmOcpCYjffxdxfHXaDIIjyDhdHxocGV

 -- Dumped from database version 18.4 (Debian 18.4-1.pgdg13+1)
 -- Dumped by pg_dump version 18.4 (Debian 18.4-1.pgdg13+1)
@@ -416,6 +416,7 @@
     id bigint NOT NULL,
     incident_id bigint NOT NULL,
     rule_escalation_id bigint,
+    event_id bigint,
     contact_id bigint,
     contactgroup_id bigint,
     schedule_id bigint,
@@ -1108,7 +1109,7 @@
 -- Data for Name: incident_history; Type: TABLE DATA; Schema: public; Owner: postgres
 --

-COPY public.incident_history (id, incident_id, rule_escalation_id, contact_id, contactgroup_id, schedule_id, rule_id, channel_id, "time", message, type, new_severity, old_severity, new_recipient_role, old_recipient_role, notification_state, sent_at) FROM stdin;
+COPY public.incident_history (id, incident_id, rule_escalation_id, event_id, contact_id, contactgroup_id, schedule_id, rule_id, channel_id, "time", message, type, new_severity, old_severity, new_recipient_role, old_recipient_role, notification_state, sent_at) FROM stdin;
 \.


@@ -1133,7 +1134,7 @@
 --

 COPY public.notifications_schema (id, version, "timestamp") FROM stdin;
-1      v1.0    1780385365825
+1      v1.0    1780385363867
 \.


@@ -2083,5 +2084,5 @@
 -- PostgreSQL database dump complete
 --

-\unrestrict 8UXAcMqapX7ggjZghk9nJK3Ww2OKwCycN5UJVE6mSapf2XRsoRN7kUvhCKk81Rg
+\unrestrict olL2fDNHTH61Pyoug3U3cqu4eBp7OVM7KmOcpCYjffxdxfHXaDIIjyDhdHxocGV
```

---

The `event_id` column needs to be removed from `incident_history`.

schema/mysql/upgrades/004.sql
oxzi, line 1:
```diff
--- Kaez5eiD_mariadb_schema_import.sql  2026-06-02 09:29:25.965718403 +0200
+++ Kaez5eiD_mariadb_schema_upgrade.sql 2026-06-02 09:29:24.328164899 +0200
@@ -325,6 +325,7 @@
   `id` bigint(20) NOT NULL AUTO_INCREMENT,
   `incident_id` bigint(20) NOT NULL,
   `rule_escalation_id` bigint(20) DEFAULT NULL,
+  `event_id` bigint(20) DEFAULT NULL,
   `contact_id` bigint(20) DEFAULT NULL,
   `contactgroup_id` bigint(20) DEFAULT NULL,
   `schedule_id` bigint(20) DEFAULT NULL,
@@ -342,6 +343,7 @@
   PRIMARY KEY (`id`),
   KEY `fk_incident_history_incident_rule_escalation_state` (`incident_id`,`rule_escalation_id`),
   KEY `fk_incident_history_rule_escalation` (`rule_escalation_id`),
+  KEY `fk_incident_history_event` (`event_id`),
   KEY `fk_incident_history_contact` (`contact_id`),
   KEY `fk_incident_history_contactgroup` (`contactgroup_id`),
   KEY `fk_incident_history_schedule` (`schedule_id`),
@@ -455,7 +457,7 @@
 LOCK TABLES `notifications_schema` WRITE;
 /*!40000 ALTER TABLE `notifications_schema` DISABLE KEYS */;
 INSERT INTO `notifications_schema` VALUES
-(1,'v1.0',1780385365000);
+(1,'v1.0',1780385363000);
 /*!40000 ALTER TABLE `notifications_schema` ENABLE KEYS */;
 UNLOCK TABLES;
 COMMIT;
@@ -865,4 +867,4 @@
 /*!40101 SET COLLATION_CONNECTION=@OLD_COLLATION_CONNECTION */;
 /*M!100616 SET NOTE_VERBOSITY=@OLD_NOTE_VERBOSITY */;

--- Dump completed on 2026-06-02  7:29:25
+-- Dump completed on 2026-06-02  7:29:24
```

---

The `event_id` column needs to be removed from `incident_history` as well.

schema/mysql/upgrades/004.sql
```diff
@@ -0,0 +1,3 @@
+DROP TABLE incident_event;
+ALTER TABLE incident_history DROP CONSTRAINT fk_incident_history_event;
```
oxzi: This is not enough, as afterwards a key still exists, as you can see in my schema diff for this file.

For me, the combination of the two following statements worked:

```sql
alter table incident_history drop foreign key fk_incident_history_event;
alter table incident_history drop index fk_incident_history_event;
```

Furthermore, it seems like `DROP CONSTRAINT` does not exists in MySQL 5.7, as it is lacking in the docs https://dev.mysql.com/doc/refman/5.7/en/alter-table.html. As we aim to support 5.7.22, maybe consider using the `DROP FOREIGN KEY` thingy.

I found the following StackOverflow thread quite helpful: https://stackoverflow.com/questions/14122031/how-to-remove-constraints-from-my-mysql-table

internal/incident/incidents.go
yhabteab: Changed.

internal/event/event.go
yhabteab: I mean, there shouldn't be any errors there since my callback doesn't return one but it's probably better to just return it anyway. So, changed it.

schema/pgsql/upgrades/004.sql
yhabteab: Fixed.

schema/mysql/upgrades/004.sql
yhabteab: Fixed.

schema/mysql/upgrades/004.sql
yhabteab: Thanks! I've applied the suggested changes!

internal/event/event.go
```diff
@@ -104,23 +102,21 @@ func (e *Event) SetMute(muted bool, reason string) {
...
 }
 
-func (e *Event) String() string {
-	return fmt.Sprintf("[time=%s type=%q severity=%s]", e.Time, e.Type, e.Severity.String())
-}
-
-// Sync transforms this event to *event.EventRow and synchronises with the database.
-func (e *Event) Sync(ctx context.Context, tx *sqlx.Tx, db *database.DB, objectId types.Binary) error {
-	if e.ID != 0 {
+func (e *Event) String() string { return e.Name }
```
oxzi: Do all events have a name? I just searched for references of `event.Event` in the code and found at least to struct initializations without a populated `Name` field.

https://github.com/Icinga/icinga-notifications/blob/b3acd044916d715b7148a04cc04f40df9ff0d1e3/internal/incident/incident.go#L524-L530

https://github.com/Icinga/icinga-notifications/blob/b3acd044916d715b7148a04cc04f40df9ff0d1e3/internal/incident/incidents.go#L133-L139

internal/event/event.go
```diff
@@ -104,23 +102,21 @@ func (e *Event) SetMute(muted bool, reason string) {
...
+
+// MarshalLogObject implements the [zapcore.ObjectMarshaler] interface to allow logging the event as a structured object.
+func (e *Event) MarshalLogObject(encoder zapcore.ObjectEncoder) error {
```
oxzi, line 106: With the implementation of `zapcore.ObjectMarshaler`, shouldn't we make sure that this is always being called? Quick grepping revealed the following:

https://github.com/Icinga/icinga-notifications/blob/b3acd044916d715b7148a04cc04f40df9ff0d1e3/internal/listener/listener.go#L234

https://github.com/Icinga/icinga-notifications/blob/b3acd044916d715b7148a04cc04f40df9ff0d1e3/internal/listener/listener.go#L411

https://github.com/Icinga/icinga-notifications/blob/b3acd044916d715b7148a04cc04f40df9ff0d1e3/internal/incident/incident.go#L131

(and round about ten other findings in the same `incident.go` file.)

oxzi (commented):
Sorry, I have missed this the last round.

internal/event/event.go
yhabteab: I didn't change those usages on purpose because that would bloat the logs even more for little benefits. But now, since you already pointed out in https://github.com/Icinga/icinga-notifications/pull/423#discussion_r3340284976 that not all events can have names, I've removed the `event` context from all non debug logs and replaced it either with the source name or none at all since the event is always logged by the listener with its full details, I don't think logging these huge json multiple times would make anything better.

oxzi (approved):

PR #424, opened 2026-05-18
incidents: don't load incident recipients twice

yhabteab:
The `LoadOpenIncidents` function already does that for all existing active incidents, so we don't need to re-load them with each ongoing event. This should have been removed with https://github.com/Icinga/icinga-notifications/pull/129 but was overlooked, so we can safely remove it now. Here's the existing code that does the loading of incident recipients for all active incidents:
https://github.com/Icinga/icinga-notifications/blob/710c8184d8ed036fef3bede2838d4c4be5015a53/internal/incident/incidents.go#L115-L118

### DIFF
--- a/internal/incident/incident.go
+++ b/internal/incident/incident.go
@@ -759,27 +759,6 @@ func (i *Incident) getRecipientsChannel(t time.Time) rule.ContactChannels {
 	return contactChs
 }
 
-// restoreRecipients reloads the current incident recipients from the database.
-// Returns error on database failure.
-func (i *Incident) restoreRecipients(ctx context.Context) error {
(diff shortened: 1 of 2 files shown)


yhabteab:
Oh, the branch name says otherwise :)! Please ignore that since I won't recreate this PR just due to the used wrong branch name.

yhabteab:
Just noticied that this just works as intended 😭! Users are allowed to [un]subscribe from/to and manage incidents from within Icinga Notifications Web, so the recipients role can change from time to time and Icinga Notifications needs to refresh each of the recipients role before processing requests. So closing 🙈!

Issue #407, opened 2026-03-19
Let sources affect incidents in multiple ways with a single event


nilmerg:
We now want to allow events to have multiple effects. We need to prevent inconsistencies caused by an unreliable processing order. Therefore, effects should be atomic. For this reason, the `type` key is no longer used. Instead, alternative keys that can coexist in combination will be used. The following requirements must be met in this regard:

- Incidents should be able to be closed for multiple reasons. There are problems that cannot be resolved. An example of this are Director Deployments, which may cause an incident, but which is closed immediately after escalation.
- It should be possible to update incidents without triggering notifications. For example, a plugin’s output must be able to reflect the results of the last check, including in Icinga Notifications.
- In addition to the previous point, however, it should also be possible to explicitly trigger notifications even without changing the severity level. This is derived from so-called “volatile” hosts and services in Icinga.

The current distinction of events is therefore obsolete. The key `type` is replaced by the following ones:

- `incident: true` To open an incident or to trigger escalations
- `close: true` To close an incident irrespective of the severity
- `muted: bool` To mute/unmute an incident (Always in combination with `muted_reason: string`)
- `force: true` To bypass checks that may prevent notifications

## Valid Combinations and their expected outcome

Any key marked with 🚫 cannot occur when any of the other keys are set or absent as described. The endpoint should respond with an error if they occur.

| `incident` | `close` | `muted` | `force` | Outcome |
|---|---|---|---|---|
| true | | | 🚫 | An incident is opened or escalated |
| true | true | 🚫 | 🚫 | Escalates and closes an incident, no matter if just opened or already open |
| true | 🚫 | true | 🚫 | Escalates and mutes an incident, no matter if just opened or already open |
|true| 🚫 | false | 🚫 | Escalates and unmutes an incident, no matter if just opened or already open |
| | true | 🚫 | 🚫 | Closes an open incident prematurely, irrespective of the severity, no effect otherwise |
| | 🚫 | true | 🚫 | Mutes an open incident, notifications are sent, no effect otherwise |
| | 🚫 | false | 🚫 | Unmutes an open incident, notifications are sent, no effect otherwise |
| | | | | An incident is updated with a new message, severity doesn't change, no notification is sent |
| | | | true | An incident is updated with a new message, severity doesn't change, notifications are sent |

Legend:
| What | Meaning |
|--|--|
| Empty cell | no value |
| 🚫 | not applicable |

julianbrost:
[...]

PR #425, opened 2026-05-19
incident: drop special handling of ACK events

### LINKED ISSUE #410: Drop support for sources to designate a manager
Managers are only to be designated using the UI. As a result, events are no longer supposed to be able to designate one using the `manager` property.

yhabteab:
The ACK event will still be processed just like before, but the special handling of the ACK event promoting the event author to incident manager is now completely removed as requested in the referenced issue. The event type `TypeAcknowledgementSet` and `TypeAcknowledgementCleared` are still not removed from Icinga Go Library, but since we're going to drop the `type` field from the event request with https://github.com/Icinga/icinga-notifications/issues/407 anyway, it's not a problem to keep them around for now and not break existing implementations of Icinga DB with this change. This also applies to the `username` field. It's no longer required but as the event table is going to be dropped anyway with #423, I didn't include any changes regarding this apart from the docs update.

resolves #410

### DIFF
--- a/internal/incident/incident.go
+++ b/internal/incident/incident.go
@@ -2,7 +2,6 @@ package incident
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"github.com/icinga/icinga-go-library/database"
 	baseEv "github.com/icinga/icinga-go-library/notifications/event"
@@ -164,8 +163,7 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event) error {
 		return err
 	}
 
-	switch ev.Type {
-	case baseEv.TypeState:
+	if ev.Type == baseEv.TypeState {
 		if !isNew {
 			if err := i.processSeverityChangedEvent(ctx, tx, ev); err != nil {
 				return err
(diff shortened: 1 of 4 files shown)


internal/incident/incident.go
```diff
@@ -652,67 +640,6 @@ func (i *Incident) notifyContact(contact *recipient.Contact, ev *event.Event, ch
...
 
-// errSuperfluousAckEvent is returned when the same ack author submits two successive ack set events on an incident.
-// This is error is going to be used only within this incident package.
-var errSuperfluousAckEvent = errors.New("superfluous acknowledgement set event, author is already a manager")
-
-// processAcknowledgementEvent processes the given ack event.
-// Promotes the ack author to incident.RoleManager if it's not already the case and generates a history entry.
-// Returns error on database failure.
-func (i *Incident) processAcknowledgementEvent(ctx context.Context, tx *sqlx.Tx, ev *event.Event) error {
-	contact := i.runtimeConfig.GetContact(ev.Username)
```
oxzi, line 663: The `RuntimeConfig.GetContact` method would be unused afterwards, but I would say let's keep it for better days.

internal/incident/incident.go
oxzi: The "ACK" part is outdated now.

oxzi (commented):
Just a nit-pick. Otherwise, LGTM!

internal/incident/incident.go
yhabteab: Oh, never thought of it! I'll just drop it now, no need to keep unused code around.

internal/incident/incident.go
yhabteab: Fixed.

oxzi (approved):
LGTM, thanks!

PR #426, opened 2026-05-20
Prevent concurrent object access

### LINKED ISSUE #250: Data Race between object.FromEvent() and  (*Incident).RetriggerEscalations
```
WARNING: DATA RACE
Write at 0x00c0007f3d00 by goroutine 72:
  github.com/icinga/icinga-notifications/internal/object.FromEvent()
      /go/src/github.com/Icinga/icinga-notifications/internal/object/object.go:140 +0x22ce
  github.com/icinga/icinga-notifications/internal/incident.ProcessEvent()
      /go/src/github.com/Icinga/icinga-notifications/internal/incident/incidents.go:218 +0x1a8
  github.com/icinga/icinga-notifications/internal/icinga2.(*Launcher).launch.func1()
      /go/src/github.com/Icinga/icinga-notifications/internal/icinga2/launcher.go:132 +0x335
  github.com/icinga/icinga-notifications/internal/icinga2.(*Client).worker()
      /go/src/github.com/Icinga/icinga-notifications/internal/icinga2/client.go:560 +0xaf5
  github.com/icinga/icinga-notifications/internal/icinga2.(*Client).Process.gowrap1()
      /go/src/github.com/Icinga/icinga-notifications/internal/icinga2/client.go:586 +0x38

Previous read at 0x00c0007f3d00 by goroutine 195174:
  github.com/icinga/icinga-notifications/internal/incident.(*Incident).RetriggerEscalations.func1()
      /go/src/github.com/Icinga/icinga-notifications/internal/incident/incident.go:256 +0x148
  github.com/icinga/icinga-notifications/internal/utils.RunInTx()
      /go/src/github.com/Icinga/icinga-notifications/internal/utils/utils.go:46 +0x26b
  github.com/icinga/icinga-notifications/internal/incident.(*Incident).RetriggerEscalations()
      /go/src/github.com/Icinga/icinga-notifications/internal/incident/incident.go:255 +0xbe6
  github.com/icinga/icinga-notifications/internal/incident.(*Incident).evaluateEscalations.func1()
      /go/src/github.com/Icinga/icinga-notifications/internal/incident/incident.go:500 +0x564

Goroutine 72 (running) created at:
  github.com/icinga/icinga-notifications/internal/icinga2.(*Client).Process()
      /go/src/github.com/Icinga/icinga-notifications/internal/icinga2/client.go:586 +0x5cc
  github.com/icinga/icinga-notifications/internal/icinga2.(*Launcher).launch.gowrap1()
      /go/src/github.com/Icinga/icinga-notifications/internal/icinga2/launcher.go:149 +0x38

Goroutine 195174 (running) created at:
  time.goFunc()
      /usr/local/go/src/time/sleep.go:177 +0x54

```

### LINKED ISSUE #266: object.FromEvent(): avoid database queries while holding mutex
Currently, `object.FromEvent()` performs some database interaction while holding a mutex:
https://github.com/Icinga/icinga-notifications/blob/8102c7797364a1f184526a19083220fa7d7a46c9/internal/object/object.go

This obviously prevents part of the event processing from happening in parallel. It would of course be nicer if that wasn't the case.

There are two parts to this:
1. The mutex currently implicitly prevents two goroutines from interfering with each other when fetching/updating the same object. To address this, some kind of more sophisticated locking mechanism would be necessary, i.e. one that takes the object ID into account and only prevents concurrent calls for the same object, maybe something similar to `ObjectNameLock` from https://github.com/Icinga/icinga2/pull/10013.
2. While adding MySQL/MariaDB support, we noticed that the database queries currently performed are susceptible to InnoDB locking issues, see https://github.com/Icinga/icinga-notifications/pull/203#pullrequestreview-2180268854 and following comments. Currently, this happens to be prevented by the current locking, but if we improve this, this has to be addressed.

yhabteab:
This PR restructures the `Object` and `Incident` lookups and creations a bit to prevent some of the issues we were having as described in the referenced issues. Previously, the event processing logic goes like this:

1. With each ongoing request, we first sync the `Object` into the database in its own transaction. While this is happening, the entire objects cache is locked, so any other request targeting a completely different object will be blocked until the transaction is committed.
2. After the object is successfully synced, we then proceed with the rest of the processing, which includes syncing the  `Incident` and all the related records in a single and long transaction. And of course, while this is happening, the  incident object is locked, so any other request targeting the same incident will be blocked until the ongoing one completes.

This PR changes the logic to the following:

1. With each ongoing request, we first just retrieve the corresponding object from cache, and if it doesn't exist yet,  we simply create a new one in memory without saving it to the database. This way, the objects cache is never locked across database transactions, and any request targeting a completely different object can proceed promptly without being blocked.
2. After the object is retrieved or created in memory, we then use that object to lookup or create the corresponding incident in the incidents cache. Only when enter the `incident.ProcessEvent` function, and acquire the lock for that incident, we will then start a DB tx to sync the incident and all the related records just like before, but this time, the object syncing is included in this very same tx. When we receive another request targeting the same object, it will be blocked in `incident.ProcessEvent` until the ongoing one completes, but other requests can freely proceed without such blocking.

There is one difference between the old and new logic that is worth mentioning. Previously, if we fail to successfully sync the object into the database, it wasn't added to the cache at all, as the cache is only updated after the tx is committed. That's not the case anymore, but since incidents are created/cached in the same way (even with the main branch), I don't have any better way to handle this without introducing more complexity. If in doubt, we can always add some cleanup logic or the like to remove dead objects and incidents from the cache in a later/separate PR if we find it necessary.

resolves #250
resolves #266

## Blocked By
- https://github.com/Icinga/icinga-notifications/pull/423
- https://github.com/Icinga/icinga-notifications/pull/418 (this PR drops all extra tags related code, and would create a nasty merge conflicts. Plus when this PR is merged the failing GHAs will hopefully secceed 🤞🏼)

### DIFF
--- a/.github/workflows/tests_with_database.yml
+++ b/.github/workflows/tests_with_database.yml
@@ -71,12 +71,7 @@ jobs:
 
       - name: Run tests
         timeout-minutes: 10
-        # By default, multiple packages may be tested in parallel in different
-        # processes. Passing -p 1 reduces this to one process to prevent test
-        # cases in different packages from accessing the same database. Note
-        # that t.Parallel() only affects parallelism within one process, i.e.
-        # within the tests of one package.
-        run: go test -v -timeout 5m -p 1 ./...
+        run: go test -v -timeout 5m ./...
 
   postgresql:
     name: PostgreSQL ${{ matrix.version }}
@@ -130,9 +125,4 @@ jobs:
 
       - name: Run tests
         timeout-minutes: 10
-        # By default, multiple packages may be tested in parallel in different
-        # processes. Passing -p 1 reduces this to one process to prevent test
-        # cases in different packages from accessing the same database. Note
-        # that t.Parallel() only affects parallelism within one process, i.e.
-        # within the tests of one package.
-        run: go test -v -timeout 5m -p 1 ./...
+        run: go test -v -timeout 5m ./...
--- a/internal/incident/incident.go
+++ b/internal/incident/incident.go
@@ -153,6 +153,13 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event) error {
 		}
 
 		i.logger = i.logger.With(zap.String("incident", i.String()))
+	} else {
+		// For all existing incidents, we have to reload the recipients from the database to ensure that we have the
+		// most up-to-date recipient list, since the recipients or recipients roles might have changed since users
+		// are allowed to subscribe or manage incidents from within Icinga Notifications Web.
+		if err := i.restoreRecipients(ctx); err != nil {
+			return err
+		}
 	}
 
 	if ev.Type == baseEv.TypeState {
--- a/internal/incident/incident.go
+++ b/internal/incident/incident.go
@@ -145,6 +145,11 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event) error {
 	}
 	defer func() { _ = tx.Rollback() }()
 
+	if err := i.Object.SyncFromEvent(ctx, tx, ev); err != nil {
+		i.logger.Errorw("Cannot sync event object", zap.Error(err))
+		return fmt.Errorf("cannot sync event object: %w", err)
+	}
+
 	isNew := i.StartedAt.Time().IsZero()
 	if isNew {
 		err = i.processIncidentOpenedEvent(ctx, tx, ev)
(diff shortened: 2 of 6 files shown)


oxzi (dismissed):

yhabteab:
Sorry, I had to rebase this due to merge conflicts from #431.

oxzi (approved):

PR #431, opened 2026-05-29
Track muted state on incidents

### LINKED ISSUE #408: Track muted state on incidents rather than objects
The muted state is expected to be tracked for each individual incident instead.

Jan-Schuppik:
Previously the mute state (mute_reason) was stored on the object table, meaning all incidents for an object shared a single mute flag. This PR moves mute_reason onto the incident table so each incident independently owns its mute state.

  - incident table gains a mute_reason column; IsMuted() now reads MuteReason.Valid directly on the incident instead of delegating to the object
  - object table loses mute_reason; Object.IsMuted(), RestoreMutedObjects(), and the transient isMuted bool are all removed
  - Schema migrations (004.sql) copy active mute state from object to any open incident before dropping the column, preserving mute state across upgrades
  - handleMute/handleUnmute persist mute_reason via incident.Sync() in the same DB transaction as the history entry

Resolves:
- #408

resolves #408


yhabteab (dismissed):
Looks fine to me now, but as discussed offline with @nilmerg and @Jan-Schuppik, there's one bug that this PR now introduces or causes. Since the mute flag is now tracked on incident level and not on the object itself, the mute flag might get lost in the following scenario:

1. Host A goes down, and an incident is created
2. Someone puts Host A in downtime, which sets the mute flag on the incident
3. Host A now recovers while still in downtime, which automatically resolves the currently open incident
4. Host A goes down again while still in downtime, which creates a new incident, but it has no information about the host being still in downtime, so the new incident is not muted.

This is a problem because the new incident will trigger notifications, even though the host is still in downtime and should be muted. However, as this is not something that can be fixed in Icinga Notifications, but rather in Icinga DB, it's acceptable for now to merge this PR and then address the issue as part of https://github.com/Icinga/icinga-notifications/issues/407 and (some upcoming Icinga DB issue @nilmerg is going to create) in the future. I'm just mentioning it here for the sake of completeness and to make sure it's not forgotten.

internal/incident/incident.go
```diff
@@ -343,45 +343,54 @@ func (i *Incident) processIncidentOpenedEvent(ctx context.Context, tx *sqlx.Tx,
...
 func (i *Incident) handleMute(ctx context.Context, tx *sqlx.Tx, ev *event.Event) error {
-	if i.isMuted || !i.Object.IsMuted() {
+	if i.IsMuted() || !ev.Mute.Valid || !ev.Mute.Bool {
 		return nil
 	}
 
-	i.logger.Infow("Muting incident", zap.String("reason", i.Object.MuteReason.String))
+	i.logger.Infow("Muting incident", zap.String("reason", i.MuteReason.String))
+
+	i.MuteReason = types.MakeString(ev.MuteReason, types.TransformEmptyStringToNull)
```
oxzi:
```suggestion
	i.MuteReason = types.MakeString(ev.MuteReason, types.TransformEmptyStringToNull)

	i.logger.Infow("Muting incident", zap.String("reason", i.MuteReason.String))
```

I think those two lines should be switched as otherwise `MuteReason` might be empty at this point.

oxzi (commented):
At first glance, this looks fine. Furthermore, I have verified the schema upgrade.

internal/incident/incident.go
yhabteab: I think, this was my mistake as I've told Jan to use this (obviously wrong) version while resolving a nasty [merge conflicts](https://github.com/Icinga/icinga-notifications/compare/2ae5f1203b4cd6dc6381c2c0544a9be6a9491d26..b7be4dadcbce7af02e535131694d4ebaf573ab24#diff-30877d7ed04fba0471600a16c579278219727781285575c9fa048eba21945fc2R381-L411) from #423 and I didn't actually check the logs afterwards.

oxzi: @Jan-Schuppik: Could you please squash your fix in the original commit.

oxzi (approved):

PR #432, opened 2026-06-01
Schema: Add `locked` column to `source` table

### LINKED ISSUE #414: Ability for Notifications Web to lock the configuration of a particular source
Sources will be able to configure themselves and so their configuration should be read-only in Icinga Notifications Web. The database table `source` requires a new column `locked bool_enum` for this.

sukhwinder33445:
refs https://github.com/Icinga/icinga-notifications-web/issues/462
resolves #414

### DIFF
--- a/schema/mysql/upgrades/002.sql
+++ b/schema/mysql/upgrades/002.sql
@@ -1,0 +1,1 @@
+ALTER TABLE source ADD COLUMN locked enum('n', 'y') NOT NULL DEFAULT 'n';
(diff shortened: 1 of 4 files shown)


schema/mysql/upgrades/002.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE source add column locked enum('n', 'y') NOT NULL DEFAULT 'n';
```
yhabteab: s/add column/ADD COLUMN/. Same goes for the postgres upgrade script.

yhabteab:
I've updated the PR description (to resolve the correct notifications daemon issue) as this doesn't fully fix the web issue but just required as a dependency.

schema/mysql/upgrades/002.sql
sukhwinder33445: Done.

yhabteab (approved):

PR #438, opened 2026-06-03
Enhance event URL completion

### LINKED ISSUE #435: Enhance event url parsing
The current simple string concatenation to build the final event URL is buggy. It does not properly handle cases where the base URL may already contain query parameters or a trailing slash. This results in malformed URLs like `https://example.com/icingaweb2` and `/icingadb/host?name=host1` being concatenated to `https://example.com/icingaweb2//icingadb/host?name=host1` (note the double slash) or using `https://watchtower.devlab.com/icingaweb?controller=monitoring&action=host&host=watchtower.devlab.com` as base URL being completed to `https://watchtower.devlab.com/icingaweb?controller=monitoring&action=host&host=watchtower.devlab.com//icingadb/host?name=icinga2`

refs https://github.com/Icinga/icinga-notifications/issues/374

yhabteab:
Instead of using string concatenation to build the final event URL as it's currently done, we can use the `url` package to create a more robust and error-resistant URL construction. Thus, this PR refactors the code a bit to utilize that package, which should help prevent issues with malformed URLs as described in the linked issue. To accomplish this, the daemon `ConfigFile` struct has been updated to include a parsed `url.URL` instance of the Icinga Web 2 base URL, which is then used to construct the final event and incident URLs in a more structured way. I've also added a basic test cases for the `Event.CompleteURL` method to ensure that the URL construction works as expected.

fixes #435

### DIFF
--- a/internal/event/event.go
+++ b/internal/event/event.go
@@ -43,19 +43,25 @@ type Event struct {
 	evaluatedRelations map[string]jsonpath.NodeList
 }
 
-// CompleteURL prefixes the URL with the given Icinga Web 2 base URL unless it already carries a URL or is empty.
-func (e *Event) CompleteURL(icingaWebBaseUrl string) {
+// CompleteURL returns the complete URL for this event by combining the Icinga Web 2 URL with the event's own url field.
+//
+// If the event's url field is an absolute URL, is empty, or it can't be URL parsed, then this method is a no-op.
+// Otherwise, it is resolved against the provided Icinga Web 2 URL to form a complete URL.
+func (e *Event) CompleteURL(icingaWebBaseUrl *url.URL) {
 	if e.URL == "" {
 		return
 	}
 
-	if !strings.HasSuffix(icingaWebBaseUrl, "/") {
-		icingaWebBaseUrl += "/"
+	u, err := url.Parse(strings.TrimLeft(e.URL, "/"))
+	if err != nil {
+		return // leave it as is if it cannot be parsed as a URL
 	}
 
-	u, err := url.Parse(e.URL)
-	if err != nil || u.Scheme == "" {
-		e.URL = icingaWebBaseUrl + e.URL
+	if !u.IsAbs() {
+		// Actually, the Icinga Web 2 base url should always contain the trailing slash, but just in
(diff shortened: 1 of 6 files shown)


internal/event/event_test.go
```diff
@@ -13,6 +14,55 @@ import (
 func TestEvent(t *testing.T) {
 	t.Parallel()
 
+	t.Run("CompleteURL", func(t *testing.T) {
+		t.Parallel()
+
+		baseUrl, err := url.Parse("https://watchtower.devlab.com/icingaweb/")
```
oxzi: What's about this domain? Based on whois, I would guess we don't own this one and while it currently does not resolve, it might resolve to some nasty spam in the near future. What about the good old `example.com` or something based on RFC 2606?

internal/event/event.go
```diff
@@ -43,19 +43,22 @@ type Event struct {
...
+//
+// If the event's url field is an absolute URL, is empty, or it can't be URL parsed, then this method is a no-op.
+// Otherwise, it is resolved against the provided Icinga Web 2 URL to form a complete URL.
+func (e *Event) CompleteURL(icingaWebBaseUrl *url.URL) {
 	if e.URL == "" {
 		return
 	}
 
-	if !strings.HasSuffix(icingaWebBaseUrl, "/") {
-		icingaWebBaseUrl += "/"
```
oxzi, line 53: Removing this check and replacing it by `.JoinPath("/")` in `ConfigFile.Validate` splits the logic and removes the testability for "incomplete" URLs in `TestEvent`[^1].

[^1]: That's how I came to this comment, as I played with the `baseUrl` in the tests instead of changing the YAML configuration file all the time.

oxzi (commented):
PR looks good. Just two small comments.

internal/event/event_test.go
yhabteab: Changed it to `example.com`.

internal/event/event.go
yhabteab: Ok, just added an extra `JoinPath("/")` call within that method as well.

oxzi (approved):

PR #441, opened 2026-06-16
Support source connection via http over unix sockets

### LINKED ISSUE #412: Add support for HTTP over unix sockets as transport to deliver events
In addition to HTTP(S) over TCP, sources should be able to establish contact with Icinga Notifications via a Unix socket. It is assumed that the respective source is running on the same system. Therefore, sources do not need to authenticate, but only identify themselves. They can specify only the username; a password or certificate is not required.

We will generally recommend this architectural approach for our own sources.

Jan-Schuppik:
## Support HTTP over Unix sockets

Sources running on the same host as Icinga Notifications can now connect via a Unix domain socket instead of (or alongside) TCP. This removes the need to expose a network port for local sources and avoids password verification overhead, as the socket transport itself provides the trust boundary.

### Changes

**Configuration** (`internal/daemon/config.go`)

Three new fields are added to the `Listener` config block:

```yaml
listener:
  socket: /run/icinga-notifications/notifications.sock
  socket_mode: "0660"  # optional, defaults to 0600
  socket_group: icinga  # optional
```

`Validate()` is updated so the TCP address only defaults to `localhost:5680` when no socket is configured; setting `socket` without `address` starts the socket listener exclusively. It also removes a stale socket file at the configured path if one already exists.

**Listener** (`internal/listener/listener.go`)

`NewListener` gains a `useSocket bool` parameter. `Run()` dispatches to either `runTCP()` (the existing logic, renamed) or the new `runSocket()`, which:

- Listens on the configured Unix path via `net.Listen("unix", ...)`
- Sets group ownership via `os.Chown` when `socket_group` is configured
- Sets permissions via `os.Chmod` immediately after binding
- Shuts down gracefully on context cancellation, same as the TCP path

Authentication on the socket path uses OS peer credentials (`SO_PEERCRED`): the daemon reads the connecting process's UID from the kernel, resolves it to an OS username via `user.LookupId`, and matches it against the source's configured username; no password or HTTP Basic Auth is involved. The TCP path continues to use full credential verification. The `WWW-Authenticate` response header is omitted on the socket path since browser-style auth prompts are not applicable there.

**Runtime config** (`internal/config/runtime.go`)

`GetSourceFromUsername` is added alongside the existing `GetSourceFromCredentials`. It matches sources by `listener_username` without checking the password hash, and explicitly guards against matching sources that have no username configured (`ListenerUsername.Valid == false`).

**Main** (`cmd/icinga-notifications/main.go`)

Both listeners are started concurrently via `errgroup`. If only one is configured, only that one runs. Either listener failing propagates the error and cancels the other through the shared context.

### Tests

- `TestListenerValidate`: covers the `Validate()` address-defaulting logic: nothing configured, socket-only, explicit address, both, and invalid/empty `socket_mode` strings
- `TestGetSourceFromUsername`: known usernames, unknown username, and sources with no configured username not matched by empty string
- `TestSourceFromAuthOrAbort` : socket path: valid peer credentials matching a source username succeed, unknown UID and missing credentials both return 401 without `WWW-Authenticate`; TCP path: correct credentials pass, wrong password and unknown username both return 401 with `WWW-Authenticate`

resolves #412

### DIFF
--- a/cmd/icinga-notifications/main.go
+++ b/cmd/icinga-notifications/main.go
@@ -22,6 +22,7 @@ import (
 	"github.com/okzk/sdnotify"
 	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
+	"golang.org/x/sys/unix"
 )
 
 const (
@@ -34,6 +35,7 @@ func main() {
 }
 
 func run() int {
+	unix.Umask(0077) // Ensure Unix sockets are created with 0600 by default, denying group/other access.
 	daemon.ParseFlagsAndConfig()
 	conf := daemon.Config()
 
@@ -91,16 +93,26 @@ func run() int {
 		}
 	})
 
-	eg.Go(func() error {
-		err := listener.NewListener(db, runtimeConfig, logs).Run(ctx)
-		if err == nil || errors.Is(err, context.Canceled) {
-			logger.Info("Listener has finished")
-			return nil
-		} else {
-			logger.Errorf("Listener has finished with an error: %+v", err)
-			return err
-		}
-	})
+	listenerConf := daemon.Config().Listener
+	makeListeners := func(useSocket bool) {
+		eg.Go(func() error {
+			err := listener.NewListener(db, runtimeConfig, logs, useSocket).Run(ctx)
+			if err == nil || errors.Is(err, context.Canceled) {
+				logger.Info("Listener has finished")
+				return nil
+			} else {
+				logger.Errorf("Listener has finished with an error: %+v", err)
+				return err
+			}
+		})
+	}
+
+	if listenerConf.Socket != "" {
+		makeListeners(true)
+	}
+	if listenerConf.Addr != "" {
+		makeListeners(false)
+	}
 
 	eg.Go(func() error {
 		logger := logs.GetChildLogger("event-queue")
--- a/config.example.yml
+++ b/config.example.yml
@@ -14,6 +14,20 @@ listener:
   # Defaults to "localhost:5680".
   #address: "localhost:5680"
 
+  # Path to a Unix domain socket for local sources to submit events.
+  # When set, sources connecting via this socket identify themselves by username only; no password is required.
+  # This transport is recommended when the source and daemon run on the same host. Leave unset (the default) to disable.
+  #socket: "/run/icinga/icinga-notifications.sock"
+
+  # Permission bits for the Unix socket file, given as an octal. Defaults to "0660".
+  # To restrict access to a specific OS group, set socket_group and leave this at "0660".
+  # Note: each source requires its own OS user, so "0600" (owner-only) would limit the socket to a single source.
+  #socket_mode: "0660"
+
+  # OS group to assign to the Unix socket file. When socket_mode grants group access (e.g. "0660"),
+  # this controls which group gets that access. Defaults to the primary group of the daemon process.
+  #socket_group: "icinga-notifications"
+
   # Set credentials for all the HTTP debug endpoints. This can be set either directly or read from a file.
   # Setting both options is not allowed and will result in a configuration error. If both options are left
   # unset, the debug endpoints will be disabled and won't be accessible at all.
--- a/doc/03-Configuration.md
+++ b/doc/03-Configuration.md
@@ -43,19 +43,34 @@ It may also be `/usr/lib/icinga-notifications/channels`, depending on the operat
 ## HTTP API Configuration
 
 Configuration of the HTTP API listener for event submission and debugging endpoints.
+Icinga Notifications can run a TCP listener, a Unix socket listener, or both simultaneously.
+When neither `address` nor `socket` is configured, the TCP listener defaults to `localhost:5680`.
+TLS options apply only to the TCP listener.
+
+When a Unix socket is configured, Icinga Notifications identifies connecting processes by their OS user,
+and matches the OS username against the source's configured username. No password or HTTP Basic Auth is involved.
 
 For YAML configuration, the options are part of the `listener` section.
 For environment variables, each option is prefixed with `ICINGA_NOTIFICATIONS_LISTENER_`.
 
-| Option              | Description                                                                                    |
-|---------------------|------------------------------------------------------------------------------------------------|
-| address             | Address to bind to, port included. (Example: `localhost:5680`)                                 |
-| debug_password      | Password expected via HTTP Basic Authentication for debug endpoints.                           |
-| debug_password_file | `debug_password` in a file.                                                                    |
-| tls                 | **Optional.** Whether to require TLS for the listener. Defaults to `false`.                    |
-| cert                | **Optional.** Path to TLS server certificate. Required if `tls` is enabled.                    |
-| key                 | **Optional.** Path to the TLS private key. Required if `tls` is enabled.                       |
-| ca                  | **Optional.** Path to TLS CA cert/bundle to verify client certs. Required if `tls` is enabled. |
+| Option              | Description                                                                                                |
+|---------------------|------------------------------------------------------------------------------------------------------------|
+| address             | **Optional.** TCP address to bind to. Defaults to `localhost:5680` when `socket` is not set.               |
+| socket              | **Optional.** Path to a Unix domain socket for local event submission.                                     |
+| socket_mode         | **Optional.** Permission bits for the Unix socket file, as an octal. Defaults to `0660`.                   |
+| socket_group        | **Optional.** OS group to assign to the Unix socket file. Defaults to Icinga Notifications' primary group. |
+| debug_password      | Password expected via HTTP Basic Authentication for debug endpoints.                                       |
+| debug_password_file | `debug_password` in a file.                                                                                |
+| tls                 | **Optional.** Whether to require TLS for the TCP listener. Defaults to `false`.                            |
+| cert                | **Optional.** Path to TLS server certificate. Required if `tls` is enabled.                                |
+| key                 | **Optional.** Path to the TLS private key. Required if `tls` is enabled.                                   |
+| ca                  | **Optional.** Path to TLS CA cert/bundle to verify client certs. Required if `tls` is enabled.             |
+
+!!! important
+
+    The socket permissions restricts who can access the Unix socket.
+    A process can only submit events for sources whose configured listener_username matches the process's OS username.
+    So each source requires its own OS user; set `0600` (owner-only) would limit the socket to a single source.
 
 ## Database Configuration
 
--- a/doc/20-HTTP-API.md
+++ b/doc/20-HTTP-API.md
@@ -11,7 +11,18 @@ After creating a source in Icinga Notifications Web,
 the specified credentials can be used via HTTP Basic Authentication to submit a JSON-encoded
 [`Event`](https://github.com/Icinga/icinga-go-library/blob/main/notifications/event/event.go).
 
-The authentication is performed via HTTP Basic Authentication using the source's username and password.
+Authentication differs by transport:
+
+- **TCP:** HTTP Basic Authentication is used; both the source's username and password must match
+  the configured credentials.
+- **Unix socket:** The caller is identified automatically by their OS user. No HTTP Basic Auth or
+  password is involved.
+
+!!! important
+
+    A process connecting via the Unix socket can only submit events for sources whose configured
+    listener_username matches the process's OS username. Restrict socket access using `socket_mode` and
+    `socket_group` to limit which OS users can connect.
 
 !!! info
 
@@ -88,11 +99,19 @@ curl -v -u 'icingadb:insecureinsecure' -H 'X-Icinga-Reject-If-Relations-Incomple
 EOF
 ```
 
+To submit over a Unix socket instead, pass `--unix-socket /run/icinga/icinga-notifications.sock` to curl.
+No credentials are needed; the daemon identifies the caller by their OS user automatically.
+
+!!! info
+
+    curl must be executed as a user which is configured as listener_username of a source.
+
 ### Get Incidents
 
 A source can query the list of open incidents belonging to its objects using the `/incidents` HTTP API endpoint.
 
-The authentication is performed via HTTP Basic Authentication using the source's username and password.
+Authentication follows the same transport-specific rules as for event submission: TCP requires HTTP
+Basic Auth with username and password, while a Unix socket identifies the caller by their OS user.
 
 ```
 $ curl -u 'example:insecureinsecure' 'http://localhost:5680/incidents'
@@ -121,6 +140,8 @@ $ curl -u 'example:insecureinsecure' 'http://localhost:5680/incidents'
 There are multiple endpoints for dumping specific configurations.
 All of them are prefixed by `/debug`.
 To use those, the `debug-password` must be set and supplied via HTTP Basic Authentication next to an arbitrary username.
+Unlike event submission, the debug endpoints always require the password regardless of the transport; connecting via
+Unix socket does not bypass this check.
 
 ### Dump Config
 
--- a/internal/config/runtime.go
+++ b/internal/config/runtime.go
@@ -4,6 +4,9 @@ import (
 	"context"
 	"database/sql"
 	"errors"
+	"sync"
+	"time"
+
 	"github.com/icinga/icinga-go-library/database"
 	"github.com/icinga/icinga-go-library/logging"
 	"github.com/icinga/icinga-go-library/types"
@@ -13,8 +16,6 @@ import (
 	"github.com/icinga/icinga-notifications/internal/timeperiod"
 	"go.uber.org/zap"
 	"golang.org/x/crypto/bcrypt"
-	"sync"
-	"time"
 )
 
 // RuntimeConfig stores the runtime representation of the configuration present in the database.
@@ -190,6 +191,23 @@ func (r *RuntimeConfig) GetSourceFromCredentials(user, pass string, logger *logg
 	}
 
 	return src
+}
+
+// GetSourceByUsername looks up a Source by its listener username without verifying a password.
+//
+// This method is intended for connections over trusted transports such as Unix sockets, where the
+// transport itself provides the security guarantee that no password is required.
+func (r *RuntimeConfig) GetSourceByUsername(user string) *Source {
+	r.RLock()
+	defer r.RUnlock()
+
+	for _, src := range r.Sources {
+		if src.ListenerUsername.Valid && src.ListenerUsername.String == user {
+			return src
+		}
+	}
+
+	return nil
 }
 
 func (r *RuntimeConfig) fetchFromDatabase(ctx context.Context) error {
--- a/internal/config/runtime_test.go
+++ b/internal/config/runtime_test.go
@@ -1,0 +1,58 @@
+package config
+
+import (
+	"testing"
+
+	"github.com/icinga/icinga-go-library/types"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetSourceByUsername(t *testing.T) {
+	t.Parallel()
+
+	srcA := &Source{ListenerUsername: types.MakeString("icingadb")}
+	srcB := &Source{ListenerUsername: types.MakeString("icinga2")}
+	srcNoUser := &Source{} // ListenerUsername.Valid == false
+
+	rc := &RuntimeConfig{}
+	rc.Sources = map[int64]*Source{
+		1: srcA,
+		2: srcB,
+		3: srcNoUser,
+	}
+
+	tests := []struct {
+		name     string
+		username string
+		want     *Source
+	}{
+		{
+			name:     "KnownUsername1",
+			username: "icingadb",
+			want:     srcA,
+		},
+		{
+			name:     "KnownUsername2",
+			username: "icinga2",
+			want:     srcB,
+		},
+		{
+			name:     "UnknownUsername",
+			username: "unknown",
+			want:     nil,
+		},
+		{
+			name:     "EmptyUsername",
+			username: "",
+			want:     nil,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			assert.Same(t, tc.want, rc.GetSourceByUsername(tc.username))
+		})
+	}
+}
--- a/internal/daemon/config.go
+++ b/internal/daemon/config.go
@@ -4,8 +4,11 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"io/fs"
 	"net/url"
 	"os"
+	"os/user"
+	"strconv"
 	"time"
 
 	"github.com/creasty/defaults"
@@ -21,9 +24,27 @@ const (
 	ExitFailure = 1
 )
 
+// Mode is a Unix file permission mode parsed from an octal string in YAML or environment configuration.
+type Mode fs.FileMode
+
+// UnmarshalText implements [encoding.TextUnmarshaler] for Mode, parsing an octal string (e.g. "0660") into a file mode.
+func (m *Mode) UnmarshalText(text []byte) error {
+	parsedString, err := strconv.ParseUint(string(text), 8, 32)
+
+	if err != nil {
+		return fmt.Errorf("invalid socket_mode %q: expected an octal value like 0660: %w", text, err)
+	}
+	*m = Mode(parsedString)
+
+	return nil
+}
+
 // Listener defines the configuration for the Icinga Notifications API listener.
 type Listener struct {
-	Addr              string           `yaml:"address" env:"ADDRESS" default:"localhost:5680"`
+	Addr              string           `yaml:"address" env:"ADDRESS"`
+	Socket            string           `yaml:"socket" env:"SOCKET"`
+	SocketMode        *Mode            `yaml:"socket_mode" env:"SOCKET_MODE" default:"0660"`
+	SocketGroup       string           `yaml:"socket_group" env:"SOCKET_GROUP"`
 	DebugPassword     string           `yaml:"debug_password" env:"DEBUG_PASSWORD"`
 	DebugPasswordFile string           `yaml:"debug_password_file" env:"DEBUG_PASSWORD_FILE"`
 	TLSOptions        config.TLSCommon `yaml:",inline"`
@@ -32,6 +53,31 @@ type Listener struct {
 func (l *Listener) Validate() error {
 	if err := config.LoadPasswordFile(&l.DebugPassword, l.DebugPasswordFile); err != nil {
 		return err
+	}
+
+	if l.Socket != "" {
+		if mode := l.SocketMode; mode != nil && *mode > 0o777 {
+			return fmt.Errorf("the socket_mode \"%04o\" is too large (max 777)", mode)
+		} else if mode != nil && *mode&0o666 == 0 {
+			return fmt.Errorf(
+				"socket_mode \"%04o\" grants no read/write access; the socket cannot accept connections",
+				mode,
+			)
+		}
+
+		if groupName := l.SocketGroup; groupName != "" {
+			group, err := user.LookupGroup(groupName)
+			if err != nil {
+				return fmt.Errorf("cannot find group %q: %w", groupName, err)
+			}
+
+			if _, err = strconv.Atoi(group.Gid); err != nil {
+				return fmt.Errorf("cannot parse GID for group %q: %w", groupName, err)
+			}
+		}
+	} else if l.Addr == "" {
+		// Only set the default Address for TCP server if no socket path is provided
+		l.Addr = "localhost:5680"
 	}
 
 	if l.TLSOptions.Enable {
--- a/internal/daemon/config_test.go
+++ b/internal/daemon/config_test.go
@@ -1,0 +1,95 @@
+package daemon
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestListenerValidate(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name            string
+		listener        Listener
+		wantAddr        string
+		wantSocket      string
+		wantSocketMode  *Mode
+		wantSocketGroup string
+		wantErr         bool
+	}{
+		{
+			name:            "NoConnectionConfigured",
+			listener:        Listener{},
+			wantAddr:        "localhost:5680",
+			wantSocket:      "",
+			wantSocketGroup: "",
+		},
+		{
+			name:            "SocketOnly",
+			listener:        Listener{Socket: "/tmp/test.sock", SocketMode: new(Mode(0660))},
+			wantAddr:        "",
+			wantSocket:      "/tmp/test.sock",
+			wantSocketMode:  new(Mode(0660)),
+			wantSocketGroup: "",
+		},
+		{
+			name:            "AddrOnly",
+			listener:        Listener{Addr: "localhost:5681"},
+			wantAddr:        "localhost:5681",
+			wantSocket:      "",
+			wantSocketGroup: "",
+		},
+		{
+			name:            "AddrAndSocket",
+			listener:        Listener{Addr: "localhost:5681", Socket: "/tmp/test.sock", SocketMode: new(Mode(0660))},
+			wantAddr:        "localhost:5681",
+			wantSocket:      "/tmp/test.sock",
+			wantSocketMode:  new(Mode(0660)),
+			wantSocketGroup: "",
+		},
+		{
+			name:            "PermissiveSocketMode",
+			listener:        Listener{Socket: "/tmp/test.sock", SocketMode: new(Mode(0777))},
+			wantAddr:        "",
+			wantSocket:      "/tmp/test.sock",
+			wantSocketMode:  new(Mode(0777)),
+			wantSocketGroup: "",
+		},
+		{
+			name:     "SocketModeExceedsValidBits",
+			listener: Listener{Socket: "/tmp/test.sock", SocketMode: new(Mode(06660))},
+			wantErr:  true,
+		},
+		{
+			name:     "SocketModeNoReadOrWritePermission",
+			listener: Listener{Socket: "/tmp/test.sock", SocketMode: new(Mode(0100))},
+			wantErr:  true,
+		},
+		{
+			name:     "SocketGroupNotExists",
+			listener: Listener{Socket: "/tmp/test.sock", SocketMode: new(Mode(0600)), SocketGroup: "unknownGroup"},
+			wantErr:  true,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			err := tc.listener.Validate()
+
+			if tc.wantErr {
+				require.Error(t, err)
+				return
+			}
+
+			require.NoError(t, err)
+			assert.Equal(t, tc.wantAddr, tc.listener.Addr)
+			assert.Equal(t, tc.wantSocket, tc.listener.Socket)
+			assert.Equal(t, tc.wantSocketMode, tc.listener.SocketMode)
+			assert.Equal(t, tc.wantSocketGroup, tc.listener.SocketGroup)
+		})
+	}
+}
--- a/internal/listener/listener.go
+++ b/internal/listener/listener.go
@@ -4,8 +4,14 @@ import (
 	"context"
 	"crypto/subtle"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"io/fs"
+	"net"
 	"net/http"
+	"os"
+	"os/user"
+	"strconv"
 	"time"
 
 	"github.com/icinga/icinga-go-library/database"
@@ -37,21 +43,29 @@ func (rw *responseWriter) WriteHeader(status int) {
 	rw.ResponseWriter.WriteHeader(status)
 }
 
+// peerUserLookupKey is the context key under which a [peerUserLookupFunc] is stored for Unix socket connections.
+type peerUserLookupKey struct{}
+
+// peerUserLookupFunc resolves the OS username of the peer process connected via a Unix domain socket.
+type peerUserLookupFunc func() (string, error)
+
 type Listener struct {
 	db            *database.DB
 	logger        *logging.Logger
 	runtimeConfig *config.RuntimeConfig
 
-	logs *logging.Logging
-	mux  http.ServeMux
+	logs      *logging.Logging
+	mux       http.ServeMux
+	useSocket bool
 }
 
-func NewListener(db *database.DB, runtimeConfig *config.RuntimeConfig, logs *logging.Logging) *Listener {
+func NewListener(db *database.DB, runtimeConfig *config.RuntimeConfig, logs *logging.Logging, useSocket bool) *Listener {
 	l := &Listener{
 		db:            db,
 		logger:        logs.GetChildLogger("listener"),
 		logs:          logs,
 		runtimeConfig: runtimeConfig,
+		useSocket:     useSocket,
 	}
 
 	debugMux := http.NewServeMux()
@@ -88,34 +102,18 @@ func (l *Listener) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		zap.String("status", http.StatusText(crw.status)))
 }
 
-// Run the Listener's web server and block until the server has finished.
+// Run starts the HTTP listener and blocks until the server finishes.
 //
-// The web server either returns (early) when its ListenAndServe fails or when the given context is finished. After the
-// context is done, the web server shuts down gracefully with a hard limit of three seconds.
-//
-// An error is returned in every case except for a gracefully context-based shutdown without hitting the time limit.
+// An error is returned in all cases except for a graceful shutdown triggered by the context being done within a
+// hardcoded time limit.
 func (l *Listener) Run(ctx context.Context) error {
-	conf := daemon.Config().Listener
-	tlsConfig, err := conf.GetTlsConfig()
-	if err != nil {
-		return err
-	}
-
-	var https string
-	if conf.TLSOptions.Enable {
-		https = "s"
-	}
-	l.logger.Infof("Starting listener on http%s://%s", https, conf.Addr)
-
 	stdlogger, err := zap.NewStdLogAt(l.logger.Desugar(), zap.ErrorLevel)
 	if err != nil {
 		return err
 	}
 
 	server := &http.Server{
-		Addr:        conf.Addr,
 		Handler:     l,
-		TLSConfig:   tlsConfig,
 		ReadTimeout: 10 * time.Second,
 		IdleTimeout: 30 * time.Second,
 		// Redirect the standard library's HTTP server error log to our logger with error level because these
@@ -123,16 +121,19 @@ func (l *Listener) Run(ctx context.Context) error {
 		ErrorLog: stdlogger,
 	}
 
+	var listeningFunc func() error
+	if l.useSocket {
+		listeningFunc, err = l.initSocketServer(server)
+	} else {
+		listeningFunc, err = l.initTcpServer(server)
+	}
+	if err != nil {
+		return err
+	}
+
 	serverErr := make(chan error, 1)
 	go func() {
-		if conf.TLSOptions.Enable {
-			// We've already created the TLS config for the server, so we can pass empty strings
-			// for certFile and keyFile, which makes ListenAndServeTLS use the TLS config directly
-			// instead of trying to load certs from files.
-			serverErr <- server.ListenAndServeTLS("", "")
-		} else {
-			serverErr <- server.ListenAndServe()
-		}
+		serverErr <- listeningFunc()
 	}()
 
 	select {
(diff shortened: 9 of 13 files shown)


internal/listener/listener.go
```diff
@@ -146,19 +163,96 @@ func (l *Listener) Run(ctx context.Context) error {
...
+
+	listener, err := net.Listen("unix", path)
+	if err != nil {
+		return fmt.Errorf("cannot listen on unix socket %q: %w", path, err)
+	}
+
+	if err := os.Chmod(path, mode); err != nil {
+		_ = listener.Close()
+		return fmt.Errorf("cannot set permissions on unix socket %q: %w", path, err)
+	}
```
oxzi, line 206: Creating the socket first and adjusting the permissions later theoretically allows an attacker to race this. Please set a [`umask(2)`](https://www.man7.org/linux/man-pages/man2/umask.2.html) first and reset it afterwards. In Go, the system call is available in [`/x/sys/unix` as `unix.Umask`](https://pkg.go.dev/golang.org/x/sys/unix#Umask).

config.example.yml
```diff
@@ -14,6 +14,16 @@ listener:
   # Defaults to "localhost:5680".
   #address: "localhost:5680"
 
+  # Path to a Unix domain socket for local sources to submit events.
+  # When set, sources connecting via this socket identify themselves by username only; no password is required.
+  # This transport is recommended when the source and daemon run on the same host. Leave unset (the default) to disable.
+  #socket: "/run/icinga/icinga-notifications.sock"
+
+  # Permission bits for the Unix socket file, given as an octal string. Defaults to "0660".
+  # With "0660", the daemon and source processes must share an OS group (e.g., "icingadb").
```
oxzi: > With "0660", the daemon and source processes must share an OS group (e.g., "icingadb").

This is wrong. With `0660`, the socket's group will be the primary group of the Icinga Notifications daemon user and the connecting user needs to be part of this group. This is a relevant difference.

---

Btw, when allowing to set permissions next to `0600`, the socket's group should be configurable as well. Regarding to my earlier `umask(2)` comment, you might want to start with an owner-only umask, create the socket, `chown(2)` the user group (`chgrp(1)`), `chmod(2)` the socket to this `socket_mode`.

internal/daemon/config.go
```diff
@@ -33,6 +36,18 @@ func (l *Listener) Validate() error {
...
+		if _, err := strconv.ParseUint(l.SocketMode, 8, 32); err != nil {
+			return fmt.Errorf(
+				"invalid listen-unix-mode %q: expected an octal value like 0660: %w",
+				l.SocketMode,
+				err,
+			)
+		}
+	} else if l.Addr == "" {
+		l.Addr = "localhost:5680"
+	}
```
oxzi, line 81: While this change tries to keep consistency to the previous `default` struct tag, it introduces a conditional default. Prior, there was always an HTTP listener on `localhost:5680` when nothing was configured. Now, it is only there if one also has not configured an Unix domain socket. This is not the same - and not documented.

cmd/icinga-notifications/main.go
```diff
@@ -65,9 +68,29 @@ func main() {
...
+			}
+			return nil
+		})
+	}
+
+	if err := g.Wait(); err != nil {
+		logger.Errorf("A listener has finished with an error: %+v", err)
 	} else {
-		logger.Info("Listener has finished")
+		logger.Info("All listeners have finished")
```
oxzi: Consider a configuration where both `.Socket` and `.Addr` are unset. While technically correct, the info log would still be confusing, as Icinga Notifications would have not worked as expected.

Maybe consider this for the `Listener.Validate` method.

internal/listener/listener.go
```diff
@@ -146,19 +163,96 @@ func (l *Listener) Run(ctx context.Context) error {
...
+
+	l.logger.Infof("Starting listener on unix socket %s", path)
+
+	stdlogger, err := zap.NewStdLogAt(l.logger.Desugar(), zap.ErrorLevel)
```
oxzi: From here, the method is almost identical to `Listener.runTCP`. Consider refactoring this into a common method.

internal/listener/listener.go
```diff
@@ -146,19 +163,96 @@ func (l *Listener) Run(ctx context.Context) error {
...
 // sourceFromAuthOrAbort extracts a *config.Source from the HTTP Basic Auth. If the credentials are wrong, (nil, false) is
 // returned and 401 was written back to the response writer.
 func (l *Listener) sourceFromAuthOrAbort(w http.ResponseWriter, r *http.Request) (*config.Source, bool) {
-	if authUser, authPass, authOk := r.BasicAuth(); authOk {
-		src := l.runtimeConfig.GetSourceFromCredentials(authUser, authPass, l.logger)
-		if src != nil {
-			return src, true
+	var errorMsg string
+	if l.useSocket {
+		if user, _, ok := r.BasicAuth(); ok {
```
oxzi: Nitpick: In the branch below, the variables are named `authUser` and `authOk`.

internal/daemon/config.go
```diff
@@ -22,7 +23,9 @@ const (
 
 // Listener defines the configuration for the Icinga Notifications API listener.
 type Listener struct {
-	Addr              string           `yaml:"address" env:"ADDRESS" default:"localhost:5680"`
+	Addr              string           `yaml:"address" env:"ADDRESS"`
+	Socket            string           `yaml:"socket" env:"SOCKET"`
+	SocketMode        string           `yaml:"socket_mode" env:"SOCKET_MODE" default:"0660"`
```
oxzi: The `default` struct tag does not match the suggested value in the config file. I would suggest using the more secure `0600` instead.

oxzi (commented):
Thanks for creating this PR. In general, it looks good. Please address each of my comments from a first round of review.

As your PR changes the configuration the documentation in `./doc` needs to be updated as well.

Furthermore, please refer the original issue #412 in the PR description.

internal/listener/listener.go
julianbrost: > Please set a [`umask(2)`](https://www.man7.org/linux/man-pages/man2/umask.2.html) first and reset it afterwards.

Note that the umask is a process attribute, so dealing with it in multi-threaded processes is tricky. So if done as suggested, that implementation would be prone to a race condition if there was any other goroutine setting the umask as well (for example, opening two listening sockets at the same time). Unless Go provides some higher-level mechanism for setting the socket permissions, the safest option would probably to set a restrictive umask (0077) once for the whole process and keep it that way.

internal/daemon/config.go
Jan-Schuppik: As discussed offline, the default TCP listener should only be used if no socket listener is configured. Therefore, I added documentation about it.

cmd/icinga-notifications/main.go
```diff
@@ -12,12 +17,12 @@ import (
...
 	"github.com/okzk/sdnotify"
-	"os/signal"
-	"syscall"
-	"time"
+	"golang.org/x/sync/errgroup"
+	"golang.org/x/sys/unix"
 )
 
 func main() {
+	unix.Umask(0077)
```
oxzi: Please add a one line comment explaining why we change the process's umask. Otherwise, this might result in raised eyebrows in the near future.

internal/listener/listener.go
```diff
@@ -146,19 +141,118 @@ func (l *Listener) Run(ctx context.Context) error {
...
+			return nil, fmt.Errorf("cannot find group %q: %w", groupName, err)
+		}
+
+		gid, err := strconv.Atoi(group.Gid)
+		if err != nil {
+			return nil, err
+		}
+
+		if err := os.Chown(path, -1, gid); err != nil {
+			return nil, fmt.Errorf("cannot set permissions on unix socket %q: %w", path, err)
```
oxzi: This error message does not really reflect the cause and - at least your part - is indistinguishable from the chmod error below. How about: "cannot change ownership of unix socket"

doc/03-Configuration.md
```diff
@@ -43,19 +43,29 @@ It may also be `/usr/lib/icinga-notifications/channels`, depending on the operat
...
+|---------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| address             | **Optional.** TCP address to bind to, port included. Defaults to `localhost:5680` when `socket` is not set. Set to an empty string to disable the TCP listener when a Unix socket is configured.     |
+| socket              | **Optional.** Path to a Unix domain socket for local event submission. Sources authenticate with their username only — no password is required. Recommended when the source and daemon run on the same host. Leave unset to disable the Unix socket listener. |
+| socket_mode         | **Optional.** Permission bits for the Unix socket file, as an octal string. Defaults to `0600` (owner-only access). Set to `0660` to grant the group configured in `socket_group` access as well.   |
+| socket_group        | **Optional.** OS group to assign to the Unix socket file. Only meaningful when `socket_mode` grants group access (e.g. `0660`). Defaults to the primary group of the daemon process.                 |
```
oxzi: 1. There is lots of text describing how these options interact. While I like documentation, it quickly gets confusing in tables - especially in Markdown tables when being edited. Consider moving some parts to another paragraph outside the table. I would extract this in two new paragraphs:
    1. One describing the conditionality between the TCP and the Unix domain socket listener. This could be added to the top, right after your other added line.
    2. Another adding information about Unix domain sockets, such as the mode or group ownership.
2. "TCP address to bind to, port included." - The port is always port of a TCP address, pair of IP address and port. Thus, this additional information can be dropped.

doc/20-HTTP-API.md
```diff
@@ -11,12 +11,16 @@ After creating a source in Icinga Notifications Web,
...
+Authentication is performed via HTTP Basic Authentication and differs by transport:
+
+- **TCP:** Both the source's username and password must match the configured credentials.
+- **Unix socket:** Only the source's username is checked. The password field is ignored, as the socket's file permissions provide the access control boundary.
 
 !!! info
 
     Before Icinga Notifications version 0.2.0, the username was a fixed string based on the source ID, such as `source-${id}`.
     When upgrading a setup from an earlier version, these usernames are still valid, but can be changed in Icinga Notifications Web.
+    This applies to Unix socket authentication as well: a source that still uses the old `source-${id}` format authenticates over the socket with that identifier.
```
oxzi: As Unix domain socket support was just added now, this note does not really make sense.

doc/20-HTTP-API.md
```diff
@@ -88,11 +94,71 @@ curl -v -u 'icingadb:insecureinsecure' -H 'X-Icinga-Reject-If-Relations-Incomple
...
+    ],
+    "servicegroups": [
+      {
+        "name": "production-services",
+        "display_name": "Production Services"
+      }
+    ]
+  }
+}
+EOF
```
oxzi: Having the same verbose output twice is both verbose and does not serve any real purpose. I would just mention the `--unix-socket` flag of `curl(1)`, but not show a second example.

doc/20-HTTP-API.md
oxzi, line 142: As we changed the authentication model for Unix domain sockets, I would add a note here that the debugging endpoints still require the `debug_password` regardless of the listener listens on a TCP or an Unix domain socket.

doc/20-HTTP-API.md
```diff
@@ -11,12 +11,16 @@ After creating a source in Icinga Notifications Web,
 the specified credentials can be used via HTTP Basic Authentication to submit a JSON-encoded
 [`Event`](https://github.com/Icinga/icinga-go-library/blob/main/notifications/event/event.go).
 
-The authentication is performed via HTTP Basic Authentication using the source's username and password.
+Authentication is performed via HTTP Basic Authentication and differs by transport:
+
+- **TCP:** Both the source's username and password must match the configured credentials.
+- **Unix socket:** Only the source's username is checked. The password field is ignored, as the socket's file permissions provide the access control boundary.
```
oxzi: As the security model for the Unix domain socket is a bit special, I would consider adding a note stating that any system user allowed to connect to the socket can impersonate any source.

(As an idea for the future, we could authenticate Unix domain socket users based on their user name. But this is out of scope for this PR.)

internal/listener/listener.go
```diff
@@ -146,19 +141,118 @@ func (l *Listener) Run(ctx context.Context) error {
...
+
+	return func() error { return server.ListenAndServe() }, nil
+}
+
+// initSocketServer configures the given HTTP server to listen on the configured Unix socket path and returns a
+// function that starts serving. If a socket file already exists at the path, it is removed before binding.
+// The caller is responsible for actually calling that function and for graceful shutdown.
+func (l *Listener) initSocketServer(server *http.Server) (fn func() error, retErr error) {
+	conf := daemon.Config().Listener
+	modeVal, _ := strconv.ParseUint(conf.SocketMode, 8, 32)
```
oxzi: I know that you have already checked this in `Listener.Validate`, but please just return the error here anyway. Otherwise, this might lead to nasty errors after a few rounds of refactoring.

internal/listener/listener.go
```diff
@@ -146,19 +141,118 @@ func (l *Listener) Run(ctx context.Context) error {
...
+
+	if groupName := conf.SocketGroup; groupName != "" {
+		group, err := user.LookupGroup(groupName)
+		if err != nil {
+			return nil, fmt.Errorf("cannot find group %q: %w", groupName, err)
+		}
+
+		gid, err := strconv.Atoi(group.Gid)
+		if err != nil {
+			return nil, err
```
oxzi: Please wrap the error was well.

internal/listener/listener.go
oxzi: How about logging warnings in the following cases:
- The socket is world-accessible, `modeVal & 0o006 > 0`. I am sure people have reasons to do so, but in almost any cases these reasons are wrong.
- The socket is not readable at all, `modeVal & 0o666 == 0`. This cannot work and will only result in later errors. Honestly, this case could even be an error, as it cannot work.

internal/listener/listener.go
Jan-Schuppik: oh, i overlooked that

doc/20-HTTP-API.md
Jan-Schuppik: I like the idea. I've thought about system user based authentication, too

cmd/icinga-notifications/main.go
oxzi: @Jan-Schuppik: This comment is still open.

oxzi (commented):
Looks good so far, thanks!

Could you please:
1. Address the last open comment, https://github.com/Icinga/icinga-notifications/pull/441#discussion_r3425868857.
2. Afterward, squash all commits into one, including a descriptive commit message.

Then, it would be good for me :)

cmd/icinga-notifications/main.go
Jan-Schuppik: Now, in `Listener.Validate`, we'll set a default value for `Addr` in case neither `Socket` nor `Addr` is specified; so that's fine.

oxzi (dismissed):
LGTM, thanks!

@yhabteab: Do you want to take another look?

Jan-Schuppik:
As discussed with @nilmerg, I added a password check for socket authentication if a password is configured.

oxzi:
As we just discussed using system user names as a kind of authentication, as other Unix domain socket-based daemons have been doing for ages, and I said this would be easy with confidence, I just wrote a small PoC, as it is actually not very straightforward in Go.

The following diff on top of your PR branch extracts the peer's credentials, and logs them.

```
Source lookup request from Unix domain socket   {"creds": "&{Pid:37 Uid:1000 Gid:1000}"}
Source lookup request from Unix domain socket   {"creds": "&{Pid:40 Uid:23 Gid:0}"}
```

```diff
diff --git a/internal/listener/listener.go b/internal/listener/listener.go
index a289fbb..3904d89 100644
--- a/internal/listener/listener.go
+++ b/internal/listener/listener.go
@@ -25,6 +25,7 @@ import (
        "github.com/icinga/icinga-notifications/internal/incident"
        "github.com/icinga/icinga-notifications/internal/object"
        "go.uber.org/zap"
+       "golang.org/x/sys/unix"
 )

 // responseWriter is a wrapper around [http.ResponseWriter] that captures the status code written to the response.
@@ -43,6 +44,9 @@ func (rw *responseWriter) WriteHeader(status int) {
        rw.ResponseWriter.WriteHeader(status)
 }

+// listenerUnixDomainSocketCreds is the context key to retrieve the Unix Domain Socket credentials.
+type listenerUnixDomainSocketCreds struct{}
+
 type Listener struct {
        db            *database.DB
        logger        *logging.Logger
@@ -235,6 +239,13 @@ func (l *Listener) initSocketServer(server *http.Server) (fn func() error, retEr
                return nil, fmt.Errorf("cannot set permissions on unix socket %q: %w", path, err)
        }

+       server.ConnContext = func(ctx context.Context, c net.Conn) context.Context {
+               // TODO: Check errors and types :^)
+               f, _ := c.(*net.UnixConn).File()
+               creds, _ := unix.GetsockoptUcred(int(f.Fd()), unix.SOL_SOCKET, unix.SO_PEERCRED)
+               return context.WithValue(ctx, listenerUnixDomainSocketCreds{}, creds)
+       }
+
        l.logger.Infof("Starting listener on unix socket %s", path)

        return func() error { return server.Serve(listener) }, nil
@@ -243,6 +254,13 @@ func (l *Listener) initSocketServer(server *http.Server) (fn func() error, retEr
 // sourceFromAuthOrAbort extracts a *config.Source from the HTTP Basic Auth. If the credentials are wrong, (nil, false) is
 // returned and 401 was written back to the response writer.
 func (l *Listener) sourceFromAuthOrAbort(w http.ResponseWriter, r *http.Request) (*config.Source, bool) {
+       if l.useSocket {
+               // TODO: Check type and existence
+               creds := r.Context().Value(listenerUnixDomainSocketCreds{}).(*unix.Ucred)
+               l.logger.Infow("Source lookup request from Unix domain socket",
+                       zap.String("creds", fmt.Sprintf("%+v", creds)))
+       }
+
        if authUser, authPass, authOk := r.BasicAuth(); authOk {
                if l.useSocket {
                        src := l.runtimeConfig.GetSourceFromUsername(authUser, l.logger)
```

As @julianbrost mentioned, there is also [`.SyscallConn`](https://pkg.go.dev/net#UnixConn.SyscallConn), which might be safer in case of vanishing fds.

internal/config/runtime.go
```diff
@@ -192,6 +193,24 @@ func (r *RuntimeConfig) GetSourceFromCredentials(user, pass string, logger *logg
...
+	r.RLock()
+	defer r.RUnlock()
+
+	for _, src := range r.Sources {
+		if src.ListenerUsername.Valid && src.ListenerUsername.String == user {
+			return src
+		}
+	}
+
+	logger.Debugw("Cannot find source for username", zap.String("user", user))
```
yhabteab: This can actually be logged on the caller site instead of here! You wouldn't have to require a logger as parameter then.

internal/config/runtime_test.go
```diff
@@ -0,0 +1,63 @@
...
+		3: srcNoUser,
+	}
+
+	tests := []struct {
+		name     string
+		username string
+		want     *Source
+	}{
+		{
+			name:     "known username returns correct source",
```
yhabteab: Using such insanely verbose test names isn't necessary IMHO. You can just use `Known Username 1`, `Known ... 2` etc. for the valid cases, and `Unknown Username` etc. for the invalid cases instead.

internal/config/runtime_test.go
```diff
@@ -0,0 +1,63 @@
...
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			logger := logging.NewLogger(zaptest.NewLogger(t).Sugar(), time.Hour)
+			got := rc.GetSourceFromUsername(tc.username, logger)
+			assert.Same(t, tc.want, got)
```
yhabteab: Once the logger isn't required anymore, you can make this a simple and niche one liner.

```suggestion
			assert.Same(t, tc.want, rc.GetSourceFromUsername(tc.username))
```

internal/daemon/config.go
```diff
@@ -22,7 +23,10 @@ const (
 
 // Listener defines the configuration for the Icinga Notifications API listener.
 type Listener struct {
-	Addr              string           `yaml:"address" env:"ADDRESS" default:"localhost:5680"`
+	Addr              string           `yaml:"address" env:"ADDRESS"`
+	Socket            string           `yaml:"socket" env:"SOCKET"`
+	SocketMode        string           `yaml:"socket_mode" env:"SOCKET_MODE" default:"0600"`
```
yhabteab: As discussed yesterday offline, don't forget to change the type of this field.

internal/daemon/config_test.go
```diff
@@ -0,0 +1,76 @@
...
+	t.Parallel()
+
+	tests := []struct {
+		name     string
+		listener Listener
+		wantAddr string
+		wantErr  bool
+	}{
+		{
+			name:     "nothing set defaults TCP address",
```
yhabteab: Same here! Please try to use `Camel Case` for your test names where applicable and don't need to include every detail of the test case. Just using a shortened version of it enough!

internal/daemon/config_test.go
```diff
@@ -0,0 +1,76 @@
...
+			listener: Listener{},
+			wantAddr: "localhost:5680",
+		},
+		{
+			name:     "only socket set leaves addr empty",
+			listener: Listener{Socket: "/tmp/test.sock", SocketMode: "0660"},
+			wantAddr: "",
```
yhabteab: All the test cases just assert whether the TCP address works as intended or not, but asserting the expecting socket path and octal mode would be nice too (especially when the type of the `SocketMode` is changed to something else).

internal/listener/listener.go
```diff
@@ -88,51 +96,38 @@ func (l *Listener) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
 
-// Run the Listener's web server and block until the server has finished.
-//
-// The web server either returns (early) when its ListenAndServe fails or when the given context is finished. After the
-// context is done, the web server shuts down gracefully with a hard limit of three seconds.
+// Run starts the HTTP listener and blocks until the server finishes.
 //
-// An error is returned in every case except for a gracefully context-based shutdown without hitting the time limit.
+// An error is returned in all cases except for a graceful shutdown triggered by the context being done within the
+// hard time limit.
```
yhabteab: s/the hard/a hardcoded/.

internal/listener/listener.go
```diff
@@ -88,51 +96,38 @@ func (l *Listener) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
-		return err
-	}
-
-	var https string
-	if conf.TLSOptions.Enable {
-		https = "s"
-	}
-	l.logger.Infof("Starting listener on http%s://%s", https, conf.Addr)
-
+	var listeningFunc func() error
```
yhabteab: This variable is declared far away from its usage, so move it down a bit please.

internal/listener/listener.go
```diff
@@ -146,17 +141,131 @@ func (l *Listener) Run(ctx context.Context) error {
...
+func (l *Listener) initSocketServer(server *http.Server) (fn func() error, retErr error) {
+	conf := daemon.Config().Listener
+	modeVal, err := strconv.ParseUint(conf.SocketMode, 8, 32)
+	if err != nil {
+		return nil, fmt.Errorf("invalid socket_mode %q: %w", conf.SocketMode, err)
+	}
+
+	if modeVal&0o666 == 0 {
+		return nil, fmt.Errorf("socket_mode %q grants no read/write access; the socket cannot accept connections", conf.SocketMode)
+	}
```
yhabteab: This can be moved to the `Validate` method of the respective config file.

internal/listener/listener.go
```diff
@@ -146,17 +141,131 @@ func (l *Listener) Run(ctx context.Context) error {
...
+
+	if err == nil {
+		if info.Mode()&os.ModeSocket == 0 {
+			return nil, fmt.Errorf("the configured socket path already exists and is not a socket: %q", path)
+		}
+		if err := os.Remove(path); err != nil {
+			return nil, fmt.Errorf("cannot remove existing unix socket: %w", err)
+		}
+	}
```
yhabteab, line 201: Same with these validations! I would move them to the validate method too! They are all basic sanity checks on the configured socket path, so it would be nice to report any errors detected here as part of the config validations.

internal/listener/listener_test.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/icinga/icinga-go-library/types"
+	"github.com/icinga/icinga-notifications/internal/config"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap/zaptest"
+	"golang.org/x/crypto/bcrypt"
+)
+
+func makeTestListener(t *testing.T, useSocket bool) (*Listener, *config.Source) {
+	t.Helper()
```
yhabteab: Since we don't use plain `t.Fail...` function but the `testify` package for all the assertions, using the `t.Helper()` here is meaningless, and has no real value. The testify package will always include the callstack in the formatted error, and looks like this:

With `t.Helper()`:
```bash
--- FAIL: TestSourceFromAuthOrAbort (0.00s)
    --- FAIL: TestSourceFromAuthOrAbort/tcp (0.00s)
        --- FAIL: TestSourceFromAuthOrAbort/tcp/missing_auth_header_returns_401_with_WWW-Authenticate (0.02s)
            listener_test.go:204:
                	Error Trace:	/Users/yhabteab/Workspace/go/icinga-notifications/internal/listener/listener_test.go:22
                	            				/Users/yhabteab/Workspace/go/icinga-notifications/internal/listener/listener_test.go:204
                	Error:      	An error is expected but got nil.
                	Test:       	TestSourceFromAuthOrAbort/tcp/missing_auth_header_returns_401_with_WWW-Authenticate
```

Without `t.Helper()`:
```bash
--- FAIL: TestSourceFromAuthOrAbort (0.00s)
    --- FAIL: TestSourceFromAuthOrAbort/tcp (0.00s)
        --- FAIL: TestSourceFromAuthOrAbort/tcp/unknown_username_returns_401_with_WWW-Authenticate (0.02s)
            listener_test.go:20:
                	Error Trace:	/Users/yhabteab/Workspace/go/icinga-notifications/internal/listener/listener_test.go:20
                	            				/Users/yhabteab/Workspace/go/icinga-notifications/internal/listener/listener_test.go:187
                	Error:      	An error is expected but got nil.
                	Test:       	TestSourceFromAuthOrAbort/tcp/unknown_username_returns_401_with_WWW-Authenticate
```

So, please remove it!

internal/listener/listener_test.go
```diff
@@ -0,0 +1,215 @@
...
+		runtimeConfig: rc,
+		useSocket:     useSocket,
+		logger:        logging.NewLogger(zaptest.NewLogger(t).Sugar(), time.Hour),
+	}
+
+	return l, src
+}
+
+func makePasswordlessTestListener(t *testing.T, useSocket bool) (*Listener, *config.Source) {
+	t.Helper()
```
yhabteab: Same here! This function doesn't even perform any assertions. Also, I don't think this function is required at all, since `makeTestListener` does already a similar job. You can just parameterize `makeTestListener` whether to generate a password and use that everywhere instead.

internal/listener/listener_test.go
```diff
@@ -0,0 +1,215 @@
...
+		logger:        logging.NewLogger(zaptest.NewLogger(t).Sugar(), time.Hour),
+	}
+
+	return l, src
+}
+
+func TestSourceFromAuthOrAbort(t *testing.T) {
+	t.Parallel()
+
+	t.Run("socket", func(t *testing.T) {
```
yhabteab: See my previous comments about the test names!

yhabteab (commented):
Sorry, for hopping in, but I just wanted to submit my pending comments before you push the other changes for https://github.com/Icinga/icinga-notifications/pull/441#issuecomment-4801410098.

internal/listener/socket_creds_darwin.go
```diff
@@ -0,0 +1,29 @@
...
+package listener
+
+import (
+	"strconv"
+	"syscall"
+
+	"golang.org/x/sys/unix"
+)
+
+func socketPeerCreds(rawConn syscall.RawConn) (string, error) {
```
oxzi: The function name does not really reflect its purpose anymore.

internal/listener/socket_creds_darwin.go
```diff
@@ -0,0 +1,29 @@
...
+
```
oxzi: Please remove this empty line.

internal/listener/listener.go
```diff
@@ -145,19 +142,127 @@ func (l *Listener) Run(ctx context.Context) error {
...
+			return ctx
+		}
+
+		uid, err := socketPeerCreds(rawConn)
+		if err != nil {
+			l.logger.Warnw("Cannot obtain peer credentials", zap.Error(err))
+			return ctx
+		}
+
+		return context.WithValue(ctx, listenerUnixDomainSocketUid{}, uid)
```
oxzi: Instead of fetching the UID and storing this number as a string, how about returning the UID as an `uint` from the OS-specific lookup function, looking up the username for the UID here, and passing this value in the context? This way, the user lookup is at one point and not separated between the OS-specific function - currently named `socketPeerCreds` -, this `func` right here, and `Listener.sourceFromAuthOrAbort`.

internal/listener/listener.go
```diff
@@ -145,19 +142,127 @@ func (l *Listener) Run(ctx context.Context) error {
...
+		if err != nil {
+			l.logger.Warnw("Cannot extract RawConnection", zap.Error(err))
+			return ctx
+		}
+
+		uid, err := socketPeerCreds(rawConn)
+		if err != nil {
+			l.logger.Warnw("Cannot obtain peer credentials", zap.Error(err))
+			return ctx
+		}
```
oxzi, line 259: The idea behind using `.SyscallConn()` and check the FD from there instead of directly using the `UnixConn`s FD was to be able to get the latest FDs - in case of whatever happens - during lookup.

This `ConnContext` function is called upon connection and extracts the socket options at this point in time. When retrieving this information in the handlers via the `Listener.sourceFromAuthOrAbort` method, this information might be outdated. Thus, if you want to have the advantage of the `.SyscallConn`, don't store the UID at connection time, but a callback using the `syscall.RawConn.Control` function to get the UID using the latest FD.

internal/daemon/config.go
```diff
@@ -21,9 +24,26 @@ const (
 	ExitFailure = 1
 )
 
+type Mode fs.FileMode
+
+func (m *Mode) UnmarshalText(text []byte) error {
```
oxzi, line 31: Please add the API documentation, especially for exported types, functions, or methods.

internal/daemon/config.go
```diff
@@ -21,9 +24,26 @@ const (
...
+}
+
 // Listener defines the configuration for the Icinga Notifications API listener.
 type Listener struct {
-	Addr              string           `yaml:"address" env:"ADDRESS" default:"localhost:5680"`
+	Addr              string `yaml:"address" env:"ADDRESS"`
+	Socket            string `yaml:"socket" env:"SOCKET"`
+	SocketMode        *Mode  `yaml:"socket_mode" env:"SOCKET_MODE" default:"0600"`
+	SocketGroup       string `yaml:"socket_group" env:"SOCKET_GROUP"`
+	socketGid         string
```
oxzi: I don't really understand the need for `socketGid`.

In my understanding it caches the GID of `SocketGroup`, but as a string, not an `uint`. It is only used within `Listener.GetSocketGid` where it is either stored or decoded from string to int again, after being decoded prior in the other code branch. This method is also just called twice - once in validation, and once when initializing the server. Am I correct?

If so, I would just drop this.

internal/daemon/config.go
```diff
@@ -53,6 +104,33 @@ func (l *Listener) GetTlsConfig() (*tls.Config, error) {
...
+				return -1, fmt.Errorf("cannot parse GID for group %q: %w", groupName, err)
+			}
+
+			l.socketGid = group.Gid
+		} else {
+			return -1, nil
+		}
+	} else {
+		gid, err = strconv.Atoi(l.socketGid)
+		if err != nil {
```
oxzi: Please refer to my [prior comment](https://github.com/Icinga/icinga-notifications/pull/441/changes#r3512559007), asking why storing this as a string when you have already decoded it once, just to decode it once again.

When deleting this method and caching, as I have suggested, just ignore this comment :)

internal/listener/listener.go
```diff
@@ -145,19 +142,127 @@ func (l *Listener) Run(ctx context.Context) error {
...
-	if authUser, authPass, authOk := r.BasicAuth(); authOk {
-		src := l.runtimeConfig.GetSourceFromCredentials(authUser, authPass, l.logger)
-		if src != nil {
-			return src, true
+	var errMsg string
+	if l.useSocket {
+		if value := r.Context().Value(listenerUnixDomainSocketUid{}); value != nil {
+			if uid, ok := value.(string); ok && uid != "" {
+				if u, err := user.LookupId(uid); err == nil {
+					if src := l.runtimeConfig.GetSourceFromUsername(u.Username); src != nil {
```
oxzi: These are four `ifs` deep, where each must succeed. Consider changing this to an fail-early approach by negating each check. If it fails, return with an adequate error instead. Doing so allows you to get rid of this indentation depth and even allows including the exact reason for erroring out, as there are currently four `ifs` (one even with an ANDed expression) and one does not know what when wrong.

doc/03-Configuration.md
```diff
@@ -43,20 +43,40 @@ It may also be `/usr/lib/icinga-notifications/channels`, depending on the operat
 ## HTTP API Configuration
 
 Configuration of the HTTP API listener for event submission and debugging endpoints.
+The daemon can run a TCP listener, a Unix socket listener, or both simultaneously.
+When neither `address` nor `socket` is configured, the TCP listener defaults to `localhost:5680`.
+TLS options apply only to the TCP listener.
+
+When a Unix socket is configured, the daemon identifies connecting processes by their OS user:
+it reads the peer's UID from the kernel via `SO_PEERCRED`, resolves it to a username, and matches
+it against the source's configured username. No password or HTTP Basic Auth is involved; the OS
```
oxzi: There is no reason to include that many implementation details here (which are, btw, wrong on Darwin), explain the higher level functionality - that's enough.

For the first sentence, how about:

> When a Unix socket is configured, the daemon identifies connecting processes by their OS user, and matches the OS username against the source's configured username.

doc/03-Configuration.md
```diff
@@ -43,20 +43,40 @@ It may also be `/usr/lib/icinga-notifications/channels`, depending on the operat
...
 
 For YAML configuration, the options are part of the `listener` section.
 For environment variables, each option is prefixed with `ICINGA_NOTIFICATIONS_LISTENER_`.
 
 | Option              | Description                                                                                    |
 |---------------------|------------------------------------------------------------------------------------------------|
-| address             | Address to bind to, port included. (Example: `localhost:5680`)                                 |
+| address             | **Optional.** TCP address to bind to. Defaults to `localhost:5680` when `socket` is not set.  |
+| socket              | **Optional.** Path to a Unix domain socket for local event submission.                         |
+| socket_mode         | **Optional.** Permission bits for the Unix socket file, as an octal string. Defaults to `0600`.|
```
oxzi: With switching to only authenticating Unix domain socket users by their OS username, a mode of `0600` is a bit obstructive, as it would only allow one source, named exactly as the Icinga Notifications daemon user. Thus, with this changed architecture, consider using `0660` as a default.

doc/03-Configuration.md
```diff
@@ -43,20 +43,40 @@ It may also be `/usr/lib/icinga-notifications/channels`, depending on the operat
...
 | cert                | **Optional.** Path to TLS server certificate. Required if `tls` is enabled.                    |
 | key                 | **Optional.** Path to the TLS private key. Required if `tls` is enabled.                       |
 | ca                  | **Optional.** Path to TLS CA cert/bundle to verify client certs. Required if `tls` is enabled. |
 
+!!! warning
+
+    Access to the Unix socket is granted to the OS user whose username matches a configured source.
+    A process can only submit events for sources whose configured username matches the process's OS
+    username. Keep `socket_mode` at `0600` unless group access is explicitly required, and restrict
+    `socket_group` accordingly.
```
oxzi: This confuses me, sorry. The first sentence is a backwards, as the socket permissions restrict who can access the socket; the username is used internally to authenticate later on.

doc/20-HTTP-API.md
```diff
@@ -11,7 +11,19 @@ After creating a source in Icinga Notifications Web,
...
 [`Event`](https://github.com/Icinga/icinga-go-library/blob/main/notifications/event/event.go).
 
-The authentication is performed via HTTP Basic Authentication using the source's username and password.
+Authentication differs by transport:
+
+- **TCP:** HTTP Basic Authentication is used; both the source's username and password must match
+  the configured credentials.
+- **Unix socket:** The daemon identifies the caller via OS peer credentials (`SO_PEERCRED`). It
+  reads the connecting process's UID from the kernel, resolves it to an OS username, and matches
+  it against the source's configured username. No HTTP Basic Auth or password is involved.
```
oxzi: [Again, implementation details](https://github.com/Icinga/icinga-notifications/pull/441/changes#r3512637857). Keep it on higher level.

oxzi (commented):
Thanks for the updated PR. I have taken a look at the changed Unix Domain Socket UID lookup and only its implications.

internal/daemon/config.go
Jan-Schuppik: You're correct! 
The idea was to avoid looking up the GID again, since I wanted to validate it in `Listener.Validate()` without relying on that lookup. When using `uint` the empty value conflicts with a valid value (but i admit an `uint pointer` would be the better choice). 
But i guess this is kind of an overkil.

Would you say it's cleaner to simply place the GID lookup in the server initialization?

oxzi: As I wrote in my earlier comment, I would opt for dropping this field and just resolving the group name twice.

internal/listener/listener.go
```diff
@@ -145,20 +146,160 @@ func (l *Listener) Run(ctx context.Context) error {
...
+		if !ok {
+			l.logger.Warnw("expected *net.UnixConn", zap.String("connection_type", fmt.Sprintf("%T", c)), zap.Error(err))
+			return ctx
+		}
+
+		rawConn, err := unixConn.SyscallConn()
+		if err != nil {
+			l.logger.Warnw("Cannot extract RawConnection", zap.Error(err))
+			return ctx
+		}
```
oxzi, line 245: This is both allowed to fail hard. The code path explicitly creates an Unix Domain Socket listener. So, if, for whatever reasons, this is not an Unix Conn, let it crash, `log.Fatalw`. Otherwise, the callback will fail anyway and no user can authenticate themself.

internal/listener/listener.go
```diff
@@ -36,21 +42,29 @@ func (rw *responseWriter) WriteHeader(status int) {
 	rw.ResponseWriter.WriteHeader(status)
 }
 
+// listenerUnixDomainSocketUid is the context key to retrieve the Uid of the Unix Domain Socket user.
+type listenerUnixDomainSocketUid struct{}
```
oxzi: Please update the API doc, stating that a `peerUserLookupFunc` is being returned. Also consider renaming the type, as it no longer holds an UID, but a lookup function to get the current UID.

internal/listener/listener.go
```diff
@@ -36,21 +42,29 @@ func (rw *responseWriter) WriteHeader(status int) {
...
+
+// peerUserLookupFunc
+type peerUserLookupFunc func() (string, error)
```
oxzi: The API doc is incomplete.

internal/listener/listener.go
```diff
@@ -145,20 +146,160 @@ func (l *Listener) Run(ctx context.Context) error {
...
+		if !authOk {
+			w.Header().Set("WWW-Authenticate", `Basic realm="icinga-notifications source"`)
+			return errFunc("missing or malformed basic auth credentials")
+		}
+
+		src = l.runtimeConfig.GetSourceFromCredentials(authUser, authPass, l.logger)
+		if src == nil {
+			w.Header().Set("WWW-Authenticate", `Basic realm="icinga-notifications source"`)
+			return errFunc("expected valid icinga-notifications source basic auth credentials")
 		}
```
oxzi, line 312: As I [commented over at the other PR](https://github.com/Icinga/icinga-notifications/pull/454#discussion_r3518100398), please add a debug log for each branch indicating which authentication method was used.

internal/listener/listener.go
```diff
@@ -145,20 +146,160 @@ func (l *Listener) Run(ctx context.Context) error {
...
+	}
+
+	if err := os.Chmod(path, fs.FileMode(mode)); err != nil {
+		return nil, fmt.Errorf("cannot set permissions on unix socket %q: %w", path, err)
+	}
+
+	server.ConnContext = func(ctx context.Context, c net.Conn) context.Context {
+		unixConn, ok := c.(*net.UnixConn)
+		if !ok {
+			l.logger.Warnw("expected *net.UnixConn", zap.String("connection_type", fmt.Sprintf("%T", c)), zap.Error(err))
```
oxzi:
```suggestion
			l.logger.Warnw("expected *net.UnixConn", zap.String("connection_type", fmt.Sprintf("%T", c)))
```

`err` is from the outer scope and should be `nil` at this point.

internal/daemon/config.go
```diff
@@ -34,6 +55,44 @@ func (l *Listener) Validate() error {
...
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("cannot read socket path: %w", err)
+		} else if err == nil {
+			if info.Mode()&os.ModeSocket == 0 {
+				return fmt.Errorf("the configured socket path already exists and is not a socket: %q", path)
+			}
+			if err := os.Remove(path); err != nil {
+				return fmt.Errorf("cannot remove existing unix socket: %w", err)
+			}
+		}
```
oxzi: Sorry, I have missed this earlier.

This check does not validate the configuration, but does preparing work for the listener. When checking for an existing socket here (and potentially cleaning it up) and continuing in the `initSocketServer` method, we have a typical TOCTOU situation. Thus, move this to `initSocketServer` after `path := conf.Socket` (currently line 189).

internal/listener/listener.go
```diff
@@ -87,51 +101,38 @@ func (l *Listener) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
 		ReadTimeout: 10 * time.Second,
 		IdleTimeout: 30 * time.Second,
 		// Redirect the standard library's HTTP server error log to our logger with error level because these
 		// errors are usually unexpected and indicate a problem with the server that should be investigated.
 		ErrorLog: stdlogger,
 	}
 
+	var listeningFunc func() error
+	if l.useSocket {
+		listeningFunc, err = l.initSocketServer(server)
```
oxzi, line 126: Consider adding a deferred function to clean up the socket file afterwards. Otherwise, we are always ending up with an orphaned dead socket file to be removed during startup.

internal/listener/listener.go
```diff
@@ -145,20 +146,160 @@ func (l *Listener) Run(ctx context.Context) error {
...
+		}
+
+		username, err := lookup()
+		if err != nil {
+			return errFunc(err.Error())
+		}
+
+		src = l.runtimeConfig.GetSourceFromUsername(username)
+		if src == nil {
+			return errFunc("expected valid unix-user whose username is equal to an existing source username")
```
oxzi: Nitpick: Drop the word "valid" as, when we are at this point, we have already validated that we are dealing with an unix domain socket and have a user. There cannot be an invalid user.

As this is a user facing error, maybe make it more obvious.
(_Edit: consult the `curl` output in my [comment above](https://github.com/Icinga/icinga-notifications/pull/441#pullrequestreview-4624240261)_)


```suggestion
			return errFunc(fmt.Sprintf("system user %q is not registered as a source username", username))
```

oxzi (commented):
Only small things left. In general, I have successfully tested the changes.

<details>
<summary>Click here for lots of curl!</summary>

Two socket connection, once with a user name registered at a source (auth okay, error due to #455) and once with an unknown user (auth error).

```
$ doas -u icingadb curl -v -H 'X-Icinga-Reject-If-Relations-Incomplete: true' -d '@-' --unix-socket /tmp/sock 'http://localhost:5680/process-event' <<EOF
...
EOF
*   Trying /tmp/sock:0...
* Established connection to /tmp/sock (/tmp/sock port 0) from  port 0
* using HTTP/1.x
> POST /process-event HTTP/1.1
> Host: localhost:5680
> User-Agent: curl/8.21.0
> Accept: */*
> X-Icinga-Reject-If-Relations-Incomplete: true
> Content-Length: 970
> Content-Type: application/x-www-form-urlencoded
>
* upload completely sent off: 970 bytes
< HTTP/1.1 400 Bad Request
< Content-Type: text/plain; charset=utf-8
< Server: icinga-notifications/0.2.0-g6745724
< X-Content-Type-Options: nosniff
< Date: Fri, 03 Jul 2026 09:45:04 GMT
< Content-Length: 65
<
invalid event: at least one of 'incident' or 'muted' must be set
* Connection #0 to host /tmp/sock:0 left intact
```

```
$ doas -u kaffeemaschine curl -v -H 'X-Icinga-Reject-If-Relations-Incomplete: true' -d '@-' --unix-socket /tmp/sock 'http://localhost:5680/process-event' <<EOF
...
EOF
*   Trying /tmp/sock:0...
* Established connection to /tmp/sock (/tmp/sock port 0) from  port 0
* using HTTP/1.x
> POST /process-event HTTP/1.1
> Host: localhost:5680
> User-Agent: curl/8.21.0
> Accept: */*
> X-Icinga-Reject-If-Relations-Incomplete: true
> Content-Length: 970
> Content-Type: application/x-www-form-urlencoded
>
* upload completely sent off: 970 bytes
< HTTP/1.1 401 Unauthorized
< Server: icinga-notifications/0.2.0-g6745724
< Date: Fri, 03 Jul 2026 09:51:07 GMT
< Content-Length: 80
< Content-Type: text/plain; charset=utf-8
<
expected valid unix-user whose username is equal to an existing source username
* Connection #0 to host /tmp/sock:0 left intact
```

</details>

internal/listener/listener.go
oxzi: Sorry, it already does so. My oopsie. All fine!

oxzi (approved):
🛳️

PR #442, opened 2026-06-17
Let events affect incidents in multiple ways

### LINKED ISSUE #261: The process-event API could use stricter validation
The process-event API could use stricter validation. Here's what I mean:

```
cat 2.json 
{
}

curl -u 'source-2:whatisatypename123' -d '@2.json' 'http://localhost:5680/process-event'
invalid event: tags must not be empty
```

Ok good so far, but what about:

```
cat 3.json 
{
  "tags": {
    "": ""
  }
}

curl -i -u 'source-2:whatisatypename123' -d '@3.json' 'http://localhost:5680/process-event'
event processed successfully
```

Resulting in:
screenshot_2024-07-25-143104

### LINKED ISSUE #407: Let sources affect incidents in multiple ways with a single event

yhabteab:
Apart from the main changes, this PR also contains the following three additional commits that are not directly related to the referenced issue.

    incident: update incident table only once & avoid superfluous copies

    Previously, we've updated the incident table multiple times within the
    very same transaction, and was clearly highlighted by the Go profiler.
    Since the attributes aren't referenced anywhere else, we can just update
    them in memory and only persist them to the database once at the end of
    the transaction, which significantly eliminates a bunch of useless
    upsert queries.

---

    event: add early return in `ExtractMissingRelations`

    There's no need to retrieve a parser from pool to just immediately
    release it if no filter columns were provided

---

    object: don't blindly copy all fields in `SyncFromEvent`

    Instead of copying blindly all fields, we only need to update the fields
    that were actually changed. For example, the tags field doesn't need to
    be updated here, because if they've different tags, the ID would be
    different, and thus we would have a different object in the cache. This
    was highlighted by the Go profiler as one of the hotspots for memory
    allocations, so don't update the fields that don't need to be updated.

Also, about the memory problems I had in Icinga Notifications, it was not a memory leak and there's no memory problems at all ;). It was consuming too much memory simply because I've built my containers with the `-race` flag, and after removing it, the daemon never goes beyond the 200MiB mark.

## Requires

- https://github.com/Icinga/icinga-go-library/pull/212

fixes #261
resolves #407


yhabteab:
Rebased and fixed [some issues](https://github.com/Icinga/icinga-notifications/compare/0c0e5e941a239b4312ecb0cddf1f2572842de8d3..d5fd9aee43628f9206f8a6554248268f2b90cf3e) reported by Johannes offline.

yhabteab:
Added some new basic testing for the `ProcessEvent` function in https://github.com/Icinga/icinga-notifications/pull/442/changes/c69b2d73c3a8a757f287a0d3f2d95e901686489b. This doesn't cover notifications but it's still better than nothing, I guess.

internal/incident/sync.go
```diff
@@ -16,12 +16,13 @@ import (
...
 // Upsert implements the contracts.Upserter interface.
-func (i *Incident) Upsert() interface{} {
+func (i *Incident) Upsert() any {
 	return &struct {
 		Severity    baseEv.Severity `db:"severity"`
 		RecoveredAt types.UnixMilli `db:"recovered_at"`
 		MuteReason  types.String    `db:"mute_reason"`
-	}{Severity: i.Severity, RecoveredAt: i.RecoveredAt, MuteReason: i.MuteReason}
+		Message     types.String    `db:"message"`
+	}{}
```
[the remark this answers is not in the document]
yhabteab: Nope :)! That function is mainly used to extract the columns we want to update in case the incident already exists in the database. See https://github.com/Icinga/icinga-go-library/blob/main/database/db.go#L356-L360.

internal/incident/incident.go
```diff
@@ -210,6 +211,17 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event) error {
...
+	if ev.CloseIncident() {
+		if err := i.Close(ctx, tx, false); err != nil {
+			return err
+		}
+	}
+
+	if err := i.Sync(ctx, tx); err != nil {
+		i.logger.Errorw("Failed to update incident", zap.Error(err))
+		return err
+	}
```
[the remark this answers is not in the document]
yhabteab: > which partially defeats the “update incident table only once” optimization

No, it doesn't :). The `handleMute` and `handleUnmute` functions can mutate the incident in memory too, which don't directly call `i.Sync()` on their own.

yhabteab:
Fixed the broken new unittests after the newest changes in https://github.com/Icinga/icinga-go-library/pull/212 and `ProcessEvent` function accordingly.

schema/pgsql/upgrades/incident-message.sql
oxzi, line 1: While this started already with #448, but why did we stop using incrementing numbers?

internal/incident/incident.go
```diff
@@ -184,9 +177,15 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event) error {
...
-		if err := i.triggerEscalations(ctx, tx, ev, escalations); err != nil {
+		if err := i.triggerEscalations(ctx, tx, escalations); err != nil {
 			return err
 		}
+
+		// If we have managed to trigger any new escalations, we must trigger notifications as well,
+		// even if the event itself doesn't request it.
+		triggerNotifications = triggerNotifications || len(escalations) > 0
+		// Even if the severity didn't change, we want to update the message nonetheless.
+		i.Message = types.MakeString(ev.Message, types.TransformEmptyStringToNull)
```
oxzi: If `isNew` is set, the `i.Message` was already set in `i.processIncidentOpenedEvent`. While the `Message` does not change, this duplicate is not necessary.

oxzi (commented):
In general, LGTM. Thanks.

schema/pgsql/upgrades/incident-message.sql
yhabteab: Because @nilmerg didn't like the numeric names.

internal/incident/incident.go
yhabteab: Fixed.

oxzi (approved):
🚣

PR #443, opened 2026-06-17
Implement HA for Distributed Setups

### LINKED ISSUE #394: High Availability
- This should not pose a problem for generic sources
- Sources can communicate with any endpoint, even switching between them
- Events must therefore be deduplicated
  - For this purpose, a unique ID is required for each event
  - If one does not exist, one is generated based on the event information
  - A job queue is used to ensure that an event is processed only once
- History can also be retrieved via any endpoint, not necessarily the API endpoint that first received the original event

```mermaid
graph LR
    subgraph DB [Database]
        direction TB
        JQ[Job Queue]
        H[History]
    end

    Source((Source))

    INA[Icinga Notifications A]
    INB[Icinga Notifications B]

    %% Connections
    Source --> INA
    Source --> INB

    INA --> DB
    INB --> DB
```

oxzi:
incident: Synchronize Incidents with the database

    For the upcoming HA functionality, incidents will change during Icinga
    Notification's runtime. To reflect this architectural change, database
    synchronizations will now be performed when processing events.

---

    Implement HA for Distributed Setups

    With this change, Icinga Notifications can now be used in distributed
    setups with multiple Icinga Notifications nodes. The relational database
    is used as an event queue where submitted events are first placed, and
    picked up from later.

    The main burden of a distributed setup or high availability was placed
    on to the database. Similar as in Icinga DB, transactions are used to
    ensure only one node wins. However, for Notifications, no single leader
    was elected, but a work queue was constructed where multiple nodes might
    enqueue events and dequeue events. A lock is only hold for the oldest
    event, allowing other nodes to access subsequent events.

    A more minor change is that the HTTP API now returns a 202 on event
    submission, and not after the event was processed. By doing so,
    superfluous events will also return 202 and not 406, as the evaluation
    happens after the event was pulled from the queue.

---

Fixes #394.

### DIFF
--- a/cmd/icinga-notifications/main.go
+++ b/cmd/icinga-notifications/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"errors"
+	"os"
 	"os/signal"
 	"syscall"
 	"time"
@@ -14,13 +15,25 @@ import (
 	"github.com/icinga/icinga-notifications/internal/channel"
 	"github.com/icinga/icinga-notifications/internal/config"
 	"github.com/icinga/icinga-notifications/internal/daemon"
+	"github.com/icinga/icinga-notifications/internal/event"
 	"github.com/icinga/icinga-notifications/internal/incident"
 	"github.com/icinga/icinga-notifications/internal/listener"
 	"github.com/icinga/icinga-notifications/internal/retention"
 	"github.com/okzk/sdnotify"
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+)
+
+const (
+	ExitSuccess = 0
+	ExitFailure = 1
 )
 
 func main() {
+	os.Exit(run())
+}
+
+func run() int {
 	daemon.ParseFlagsAndConfig()
 	conf := daemon.Config()
 
@@ -65,19 +78,69 @@ func main() {
 		logger.Fatalf("Cannot load incidents from database: %+v", err)
 	}
 
-	ret := retention.New(db, logs.GetChildLogger("retention"))
-	go func() {
-		if err := ret.Run(ctx); err != nil && !errors.Is(err, context.Canceled) {
-			logger.Fatalf("Retention has finished with an error: %+v", err)
+	eg, ctx := errgroup.WithContext(ctx)
+
+	eg.Go(func() error {
+		err := retention.New(db, logs.GetChildLogger("retention")).Run(ctx)
+		if err == nil || errors.Is(err, context.Canceled) {
+			logger.Info("Retention has finished")
+			return nil
+		} else {
+			logger.Errorf("Retention has finished with an error: %+v", err)
+			return err
 		}
-	}()
+	})
+
+	eg.Go(func() error {
+		err := listener.NewListener(db, runtimeConfig, logs).Run(ctx)
+		if err == nil || errors.Is(err, context.Canceled) {
+			logger.Info("Listener has finished")
+			return nil
+		} else {
+			logger.Errorf("Listener has finished with an error: %+v", err)
+			return err
+		}
+	})
+
+	eg.Go(func() error {
+		logger := logs.GetChildLogger("event-queue")
+		err := event.ProcessQueue(
+			ctx,
+			db,
+			logger,
+			func(ctx context.Context, ev *event.Event) error {
+				err := incident.ProcessEvent(ctx, db, logs, runtimeConfig, ev)
+				if errors.Is(err, incident.ErrSeverityChangeWithoutIncidentFlag) ||
+					errors.Is(err, incident.ErrOpenIncidentWithoutSeverity) {
+					logger.Debugw("Skipping event processing",
+						zap.String("event_name", ev.Name),
+						zap.Error(err))
+					return nil
+				} else if err != nil {
+					logger.Errorw("Failed to successfully process event",
+						zap.String("event_name", ev.Name),
+						zap.Error(err))
+					return err
+				}
+
+				logger.Infow("Successfully processed event", zap.String("event_name", ev.Name))
+				return nil
+			})
+		if err == nil || errors.Is(err, context.Canceled) {
+			logger.Info("Event queue processor has finished")
+			return nil
+		} else {
+			logger.Errorf("Event queue processor has finished with an error: %+v", err)
+			return err
+		}
+	})
 
 	// When Icinga Notifications is started by systemd, we've to notify systemd that we're ready.
 	_ = sdnotify.Ready()
 
-	if err := listener.NewListener(db, runtimeConfig, logs).Run(ctx); err != nil {
-		logger.Errorf("Listener has finished with an error: %+v", err)
-	} else {
-		logger.Info("Listener has finished")
+	if err := eg.Wait(); err != nil {
+		return ExitFailure
 	}
+
+	return ExitSuccess
 }
--- a/doc/05-Distributed-Setups.md
+++ b/doc/05-Distributed-Setups.md
@@ -1,3 +1,38 @@
 # Distributed Setups
 
-To be done.
+Icinga Notifications supports running multiple daemon instances against a single shared database.
+This allows for high-availability setups where instances share the processing load
+and the remaining instances continue operating if one becomes unavailable.
+
+## Architecture
+
+All instances share the same database for both configuration and event processing.
+When an event is received via the [HTTP API](20-HTTP-API.md),
+the receiving instance writes it to a queue table in the database rather than processing it immediately.
+A background worker in each instance continuously polls this queue and claims events for processing.
+
+The database ensures that each event is processed exactly once across all running instances.
+There is no leader election; all instances are equal and contribute to the shared workload.
+
+Although database queries ensure that no two instances operate on identical objects, submission time is important.
+Therefore, ensure that all nodes have identical system times, e.g., via NTP.
+
+## Configuration
+
+To run additional instances, install Icinga Notifications on another host and point it at the same database.
+No additional configuration is required beyond what is described in [Configuration](03-Configuration.md).
+
+Each instance must be able to reach the database and requires any configured notification channels to be installed.
+The HTTP API of each instance operates independently.
+If you place a load balancer in front of multiple instances,
+incoming events will be distributed across them and processed from the shared queue.
+
+## Corrupt Queue Entries
+
+The event queue is designed to be an internal API, where no external process is allowed to insert events.
+Thus, it is expected to only contain valid data.
+
+Nevertheless, in the unlikely case that corrupted data crept into the `event_queue` table,
+Icinga Notifications will log an error and mark the event as failed in the database.
+The log contains a reference to the invalid `event_queue` row including its `id`.
+This allows inspecting the entry in the relational database and eventually fixing or deleting it.
--- a/internal/event/queue.go
+++ b/internal/event/queue.go
@@ -1,0 +1,227 @@
+package event
+
+import (
+	"bytes"
+	"context"
+	"crypto/sha256"
+	"database/sql"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+	"time"
+
+	"github.com/icinga/icinga-go-library/backoff"
+	"github.com/icinga/icinga-go-library/database"
+	"github.com/icinga/icinga-go-library/logging"
+	"github.com/icinga/icinga-go-library/retry"
+	"github.com/icinga/icinga-go-library/types"
+	"github.com/icinga/icinga-notifications/internal"
+	"github.com/jmoiron/sqlx"
+	"go.uber.org/zap"
+)
+
+const (
+	// The following consts are valid values for Queue.State.
+	//
+	// - queueStatePending is the initial state for submitting events.
+	// - queueStateProcessing is assigned when the ListenQueue picked this up for processing.
+	// - queueStateDone is set after ListenQueue successfully processed the entry.
+	// - queueStateError is assigned to invalid events for further processing.
+	//
+	// queueStateDone ensures that a source cannot enqueue the same event after it was already processed. If the event
+	// would have been already deleted from the database, but the source resubmits it due to some source-facing network
+	// issues, it would have been evaluated once again. By keeping evaluated events with the queueStateDone state in
+	// the database for a few minutes eliminates this potential issue.
+	queueStatePending    int16 = 0
+	queueStateProcessing int16 = 1
+	queueStateDone       int16 = 2
+	queueStateError      int16 = 64
+)
+
+// Queue describes an event.Event enqueued for processing in the relational database.
+type Queue struct {
+	// ID is the SHA256 hash based on the event.Event JSON representation.
+	ID types.Binary `db:"id"`
+
+	// Json serialization and Time from the event.Event. Time is required for database ordering.
+	Json string          `db:"json"`
+	Time types.UnixMilli `db:"time"`
+
+	// ObjectId is the Object ID of the event, used to exclude related queue entries on the database level.
+	ObjectId types.Binary `db:"object_id"`
+
+	// UserAgent describes the submitting client; allows migrations after upgrades.
+	UserAgent string `db:"user_agent"`
+
+	// State of the event. Check "queueState*" consts above.
+	State int16 `db:"state"`
+}
+
+// TableName implements [database.TableNamer].
+func (q *Queue) TableName() string {
+	return "event_queue"
+}
+
+// toEvent recreates an event.Event based on the JSON representation and the other fields stored in the database.
+func (q *Queue) toEvent() (*Event, error) {
+	var ev Event
+
+	// NOTE: There might be breaking changes with future updates. If this happens, this function should honor the
+	// UserAgent field and apply necessary migration steps.
+
+	err := json.NewDecoder(strings.NewReader(q.Json)).Decode(&ev)
+	if err != nil {
+		return nil, fmt.Errorf("cannot JSON decode event queue entry: %w", err)
+	}
+
+	ev.Time = q.Time.Time()
+
+	return &ev, nil
+}
+
+// Enqueue enqueues an event.Event into the event queue.
+//
+// The internally used database transaction is bound to the provided context. When calling this function from an HTTP
+// handler, pass the request's context to terminate the transaction if the client disconnects prematurely.
+func Enqueue(ctx context.Context, db *database.DB, ev *Event, objectID types.Binary) error {
+	jsonBuff := bytes.Buffer{}
+	idHash := sha256.New()
+
+	err := json.NewEncoder(io.MultiWriter(&jsonBuff, idHash)).Encode(ev)
+	if err != nil {
+		return fmt.Errorf("cannot JSON encode event.Event: %w", err)
+	}
+
+	q := &Queue{
+		ID:        idHash.Sum(nil),
+		Json:      jsonBuff.String(),
+		Time:      types.UnixMilli(ev.Time),
+		ObjectId:  objectID,
+		UserAgent: "icinga-notifications/" + internal.Version.Version,
+		State:     queueStatePending,
+	}
+	stmt, _ := db.BuildInsertIgnoreStmt(q)
+
+	return retry.WithBackoff(
+		ctx,
+		func(ctx context.Context) error {
+			return db.ExecTx(
+				ctx,
+				&sql.TxOptions{Isolation: sql.LevelSerializable},
+				func(ctx context.Context, tx *sqlx.Tx) error {
+					_, err := tx.NamedExecContext(ctx, stmt, q)
+					if err != nil {
+						return database.CantPerformQuery(err, stmt)
+					}
+
+					return nil
+				})
+		},
+		retry.Retryable,
+		backoff.DefaultBackoff,
+		db.GetDefaultRetrySettings())
+}
+
+// ProcessQueue fetches events from the database event queue and forwards them to the callback.
+//
+// This function blocks until either an error occurred or the passed context is finished. In either case, an error is
+// being returned.
+func ProcessQueue(
+	ctx context.Context,
+	db *database.DB,
+	logger *logging.Logger,
+	callback func(context.Context, *Event) error,
+) error {
+	selStmt := db.Rebind(`
+		SELECT q1.*
+		FROM event_queue q1
+		LEFT JOIN event_queue q2 ON
+			q2.state = ` + fmt.Sprintf("%d", queueStateProcessing) + ` AND
+			q1.object_id = q2.object_id
+		WHERE
+			q1.state = ` + fmt.Sprintf("%d", queueStatePending) + ` AND
+			q2.object_id IS NULL
+		ORDER BY q1.time
+		LIMIT 1`)
+	updStmt := `UPDATE event_queue SET state = :state WHERE id = :id`
+
+	for ctx.Err() == nil {
+		var q Queue
+
+		err := retry.WithBackoff(
+			ctx,
+			func(ctx context.Context) error {
+				return db.ExecTx(
+					ctx,
+					&sql.TxOptions{Isolation: sql.LevelSerializable},
+					func(ctx context.Context, tx *sqlx.Tx) error {
+						err := tx.GetContext(ctx, &q, selStmt)
+						if err != nil {
+							return database.CantPerformQuery(err, selStmt)
+						}
+
+						q.State = queueStateProcessing
+						_, err = tx.NamedExecContext(ctx, updStmt, q)
+						if err != nil {
+							return database.CantPerformQuery(err, updStmt)
+						}
+
+						return nil
+					})
+			},
+			func(err error) bool {
+				if errors.Is(err, sql.ErrNoRows) {
+					return true
+				}
+				return retry.Retryable(err)
+			},
+			backoff.NewExponentialWithJitter(time.Second, 3*time.Second),
+			retry.Settings{Timeout: retry.DefaultTimeout})
+		if errors.Is(err, sql.ErrNoRows) {
+			continue
+		} else if err != nil {
+			return fmt.Errorf("cannot fetch event from event queue: %w", err)
+		}
+
+		logger.Debugw("Claimed event queue entry for processing",
+			zap.Stringer("id", q.ID),
+			zap.Stringer("object_id", q.ObjectId))
+
+		if ev, err := q.toEvent(); err != nil {
+			logger.Errorw("Invalid event queue event cannot get decoded", zap.Stringer("id", q.ID), zap.Error(err))
+			q.State = queueStateError
+		} else if err = callback(ctx, ev); err != nil {
+			logger.Errorw("Event queue event cannot be processed", zap.Stringer("id", q.ID), zap.Error(err))
+			q.State = queueStateError
+		} else {
+			logger.Debugw("Processed event from event queue", zap.Stringer("id", q.ID))
+			q.State = queueStateDone
+		}
+
+		err = retry.WithBackoff(
+			ctx,
+			func(ctx context.Context) error {
+				return db.ExecTx(
+					ctx,
+					&sql.TxOptions{Isolation: sql.LevelSerializable},
+					func(ctx context.Context, tx *sqlx.Tx) error {
+						_, err = tx.NamedExecContext(ctx, updStmt, q)
+						if err != nil {
+							return database.CantPerformQuery(err, updStmt)
+						}
+
+						return nil
+					})
+			},
+			retry.Retryable,
+			backoff.DefaultBackoff,
+			db.GetDefaultRetrySettings())
+		if err != nil {
+			return fmt.Errorf("cannot update queue entry %q after processing: %w", q.ID, err)
+		}
+	}
+
+	return ctx.Err()
+}
--- a/internal/incident/incident.go
+++ b/internal/incident/incident.go
@@ -151,18 +151,20 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event) error {
 
 		i.logger = i.logger.With(zap.String("incident", i.String()))
 	} else {
+		if err := i.restoreState(ctx, tx); err != nil {
+			return err
+		}
+
+		if !i.RecoveredAt.Time().IsZero() {
+			RemoveCurrent(i.Object)
+			return ErrIncidentRecovered
+		}
+
 		if sevChanged, err := i.processSeverityChangedEvent(ctx, tx, ev); err != nil {
 			return err
 		} else {
 			// In case the severity didn't change, we need to check whether we can trigger notifications nonetheless.
 			triggerNotifications = sevChanged || ev.NotifyRecipients() || (ev.Muted.Valid && ev.IsMuted() != i.IsMuted())
-		}
-
-		// For all existing incidents, we have to reload the recipients from the database to ensure that we have the
-		// most up-to-date recipient list, since the recipients or recipients roles might have changed since users
-		// are allowed to subscribe or manage incidents from within Icinga Notifications Web.
-		if err := i.restoreRecipients(ctx); err != nil {
-			return err
 		}
 	}
 
@@ -241,6 +243,19 @@ func (i *Incident) RetriggerEscalations(ev *event.Event) {
 	i.runtimeConfig.RLock()
 	defer i.runtimeConfig.RUnlock()
 
+	ctx := context.Background()
+	tx, err := i.db.BeginTxx(ctx, nil)
+	if err != nil {
+		i.logger.Errorw("Cannot start a db transaction for escalation reevaluation", zap.Error(err))
+		return
+	}
+	defer func() { _ = tx.Rollback() }()
+
+	if err := i.restoreState(ctx, tx); err != nil {
+		i.logger.Errorw("Cannot restore incident state for escalation reevaluation", zap.Error(err))
+		return
+	}
+
 	if !i.RecoveredAt.Time().IsZero() {
 		// Incident is recovered in the meantime.
 		return
@@ -262,31 +277,33 @@ func (i *Incident) RetriggerEscalations(ev *event.Event) {
 		return
 	}
 
-	var notifications []*NotificationEntry
-	ctx := context.Background()
-	err = i.db.ExecTx(ctx, nil, func(ctx context.Context, tx *sqlx.Tx) error {
-		if err = i.triggerEscalations(ctx, tx, escalations); err != nil {
-			return err
-		}
+	if err := i.triggerEscalations(ctx, tx, escalations); err != nil {
+		i.logger.Errorw("Reevaluating time-based escalations failed", zap.Error(err))
+		return
+	}
 
-		channels := make(rule.ContactChannels)
-		for _, escalation := range escalations {
-			channels.LoadFromEscalationRecipients(escalation, ev.Time, i.isRecipientNotifiable)
-		}
+	channels := make(rule.ContactChannels)
+	for _, escalation := range escalations {
+		channels.LoadFromEscalationRecipients(escalation, ev.Time, i.isRecipientNotifiable)
+	}
 
-		notifications, err = i.generateNotifications(ctx, tx, ev, channels)
-		return err
-	})
+	notifications, err := i.generateNotifications(ctx, tx, ev, channels)
 	if err != nil {
 		i.logger.Errorw("Reevaluating time-based escalations failed", zap.Error(err))
-	} else {
-		if err = i.notifyContacts(ctx, ev, notifications); err != nil {
-			i.logger.Errorw("Failed to notify reevaluated escalation recipients", zap.Error(err))
-			return
-		}
-
-		i.logger.Info("Successfully reevaluated time-based escalations")
+		return
 	}
+
+	if err := tx.Commit(); err != nil {
+		i.logger.Errorw("Cannot commit transaction for escalation reevaluation", zap.Error(err))
+		return
+	}
+
+	if err := i.notifyContacts(ctx, ev, notifications); err != nil {
+		i.logger.Errorw("Failed to notify reevaluated escalation recipients", zap.Error(err))
+		return
+	}
+
+	i.logger.Info("Successfully reevaluated time-based escalations")
 }
 
 // Close closes the current incident if not already recovered.
@@ -324,6 +341,11 @@ func (i *Incident) Close(ctx context.Context, tx *sqlx.Tx, persist bool) error {
 // ErrSeverityChangeWithoutIncidentFlag is returned when an event tries to change the severity of an incident
 // but does not set the 'incident' flag.
 var ErrSeverityChangeWithoutIncidentFlag = errors.New("cannot change severity of an incident with an event that doesn't set the 'incident' flag")
+
+// ErrIncidentRecovered is returned when an incident was found to be already recovered by another node in HA mode.
+//
+// When Incident.Listener returns this error, the related object was removed from the current incidents.
+var ErrIncidentRecovered = errors.New("incident was recovered by another node")
 
 // processSeverityChangedEvent processes the given event as a severity changed event, if the severity has actually changed.
 //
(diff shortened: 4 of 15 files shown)


yhabteab:
Removed it from the project, remember we don't add PRs to the project if there's a corresponding issue for it :), and all PRs without an issue should be marked as side-projects or the like.

internal/event/event_queue.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/icinga/icinga-go-library/logging"
+	"github.com/icinga/icinga-go-library/retry"
+	"github.com/icinga/icinga-go-library/types"
+	"go.uber.org/zap"
+)
+
+// EventQueue describes an event.Event enqueued for processing in the relational database.
+//
+// It is represented by its event JSON representation, which is also used to derive an SHA256 hash used as the ID.
+type EventQueue struct {
```
yhabteab: I think, just using `Queue` as a name here would be fine (just as suggested by goland) since the package it lives in is called event. If it's then accessed via `event.Queue`, it would still be clear what that represents. Same goes with the package level function names!

internal/event/event_queue.go
```diff
@@ -0,0 +1,180 @@
...
+	ID       types.Binary    `db:"id"`
+	Json     string          `db:"json"`
+	Time     types.UnixMilli `db:"time"`
+	SourceID int64           `db:"source_id"`
```
yhabteab: Why not make the `Event.SourceId` field JSON [de]serializable instead of adding an extra column for it here? That field will always be populated by the listener after decoding the request body anyway, so adding it here as an extra column with the sole pupose of populating that very same field when loaded from the database doesn't make any sense to me.

yhabteab (commented):
> Similar as in Icinga DB, transactions are used to ensure only one node wins. However, for Notifications, no single leader was elected, but a work queue was constructed where multiple nodes might enqueue events and dequeue events.

It doesn't seem you're using the same transaction isolation level as Icinga DB though, which is `SERIALIZABLE`. To the parallelism of the work dequeueing, how do you prevent multiple nodes from processing two different events that affect the same object/incident at the same time? From quickly looking at the code, I don't see any mechanism to prevent this, so Johannes and I had a quick discussion about this and other things I want to share here:

1. Icinga Notifications should never enqueue the event into the event queue if it wasn't able to tell the source of the event about the success of the enqueueing. That is, if Icinga Notifications can't successfully write the respective header and response to the client, it should rollback the transaction and discard the event. Because from the perspective of the client, the request will be considered as failed, if it didn't either receive a response at all (e.g. due to timeout or network delays), or Icinga Notifications explicitly responded with an error code. In both cases, the client will have to retry the request, and if the event was already enqueued and picked up by some node, it will be processed twice, which is not what we want (remember, with the upcoming PR of mine, Icinga Notifications won't have any logic to detect superfluous events). For this to work, the event must be enqueued in a transaction setup with `SERIALIZABLE` isolation level, and Icinga Notifications must send its success response before committing that transaction but still after the actual INSERT statement.
2. Now, if we want to allow multiple nodes to process events in parallel from the queue, we need to make sure that they don't process events that affect the same object, as we can't guarantee that the processing of oldest event will be faster than the one queued after it, and if they affect the same object, we might end up with race conditions. To prevent this, we came up with the idea to add some more columns to the event queue table. The first one is `object_id`, which should be used to correlate the event with the object it affects. Since we always JSON decode the request body before queueing the event, the `object_id` can be easily extracted from it. Note however, that it must not reference to `object` table, as the object might not exist in the database at the time of enqueueing. The second column is `state` or `status` (pick whatever you like more) of type enum with values `pending`, `processing`, and `done`. When a node dequeues an event, it should immediately set the state to `processing` and once the processing is done, it should set it to `done`. Now, when a node tries to dequeue the next event, it should check if there are any events with state `processing` that have the same `object_id` as the event it wants to process. If there are, it should skip that event and try to dequeue the next one. This means, dequeued events shouldn't be deleted from the queue, but rather track their state, and instead, there should be some kind of cleanup goroutine that periodically deletes events with state `done`.
3. When point number 2 is implemented, the cleanup goroutine is going to run e.g. every 5-10 minutes and if the source, for whatever reason, submits the same event again, it won't be queued again, because the previous one is still in the queue (either already marked as done, processing or still pending). However, since the same event can have different meanings depending on the time it was submitted, Icinga Notifications should provide a way for the source to explicitly specify the event ID, but if it doesn't, it should be fine the way it is now, because the default ID generation strategy should be deterministic, and doesn't have to be time-based. This way, if the same event is submitted multiple times within a short period of time, but with different meanings, the source can explicitly specify the event ID to make sure they are treated as different events.

That's all I have for now, but if you object to any of the points I made, or if you have any questions, feel free to ask.

/cc @nilmerg please review if I missed something!

nilmerg:
It's summarized perfectly, thanks!

Though, one thing I noticed while reading your comment was that I'm not sure it's right to expect the source to provide a time based event id. Because said event id should be possible to be re-used to map history entries from notifications with history of e.g. Icinga DB, and I don't know if said history ids are also time based? The alternative would be a timestamp that's part of the event body and then needs to be incorporated into the id used in the job queue.

internal/event/event_queue.go
oxzi: In my mental picture, the excluding JSON struct tags for `Time` and `SourceId` were safety features, ensuring that these daemon-assigned properties cannot be forged by a submitting source - either fraudulent or by accident. At the moment, both fields are being overwritten after JSON deserialization in `Listener.ProcessEvent`, but this is only one minor refactoring away from introducing a severe bug. This assumption is also kinda supported by the API doc comment for `event.Event`.

Dunno, I am also not really proud about the current code, but it feels more safe for me.

oxzi:
Thanks for the extensive architecture review, @yhabteab. I have just pushed a new version with lots of changes. Most relevant:

1. The event queue went into its own package due to cyclic imports after importing the `object` package. Thus, the diff will be quite useless, sorry.
2. Two new columns were added: `object_id` and `processed`. The second is a NULLable timestamp, being set when a node works on this event. Other nodes are now skipping events where another event with the same `object_id` is being processed.

There is currently the assumption that all queued entries are valid. Of course, this is naive. How would you proceed with the listener callback? Of course, we can retry events later, but this will only block things.

oxzi:
Changed this once again after our offline discussion.

1. Moved back to `event` package and outsourced the object ID to a parameter.
2. Changed `processed` field to `state` as you have suggested earlier.
3. Removed source ID from ID hash, added `Event.ID` via IGL PR to enable sources to distinguish similar events.
4. Introduced an error state to assign to invalid entries for later inspections instead of crashing.

At the moment, there is no garbage collection for the `event_queue`. I would like to use your PR #446 for doing so. Thus, we should proceed there first.

internal/event/queue.go
```diff
@@ -0,0 +1,243 @@
...
+		Json:     jsonBuff.String(),
+		Time:     types.UnixMilli(ev.Time),
+		SourceId: ev.SourceId,
+		ObjectId: objectID,
+		Version:  internal.Version.Version,
+		State:    queueStatePending,
+	}
+
+	// The primary key, ID, is a unique hash based on the event's JSON representation. Thus, multiple submissions can
+	// get ignored. A UPSERT query would be invalid, as it resets the "processed" column.
```
yhabteab: This comment is outdated now! Also, I don't quite get what the second sentence is trying to say! Can you please rephrase it a bit? The ID being a unique hash of the event body isn't the reason why duplicated events get ignored.

internal/event/queue.go
```diff
@@ -0,0 +1,243 @@
...
+	stmt, _ := db.BuildInsertIgnoreStmt(q)
+	stmt = db.Rebind(stmt)
```
yhabteab: You don't need to explicitly rebind named SQL statements.

internal/event/queue.go
```diff
@@ -0,0 +1,243 @@
...
+	db *database.DB,
+	logs *logging.Logging,
+	callback func(context.Context, *logging.Logger, *Event) error,
+) error {
+	logger := logs.GetChildLogger("event-queue")
+
+	selStmtCols := db.BuildColumns(new(Queue))
+	for i := range selStmtCols {
+		selStmtCols[i] = `eq."` + selStmtCols[i] + `"`
+	}
```
yhabteab: Is there any reason that I don't currently see that you've to explicitly list all the columns like this instead of just using `*`?

internal/event/queue.go
```diff
@@ -0,0 +1,243 @@
...
+		FROM event_queue eq
+		WHERE
+			eq.state = ` + fmt.Sprintf("%d", queueStatePending) + ` AND
+			NOT EXISTS (
+				SELECT 1
+				FROM event_queue sub
+				WHERE
+					sub.state = ` + fmt.Sprintf("%d", queueStateProcessing) + ` AND
+					sub.object_id = eq.object_id)
+		ORDER BY eq.time LIMIT 1`)
```
yhabteab: I'm aware that sub-queries are way easier to read but did you consider using a `JOIN` based Anti-Join here instead? I know, that the Postgres optimizer can easily optimize this sub-select too but in MariaDB it makes a big difference:
```sql
EXPLAIN SELECT q1.*
  FROM event_queue q1
    LEFT JOIN event_queue q2 ON q2.state = 1 AND q1.object_id = q2.object_id
  WHERE q1.state = 0 AND t2.object_id IS NULL
  ORDER BY q1.time
  LIMIT 1;
```

internal/event/queue.go
```diff
@@ -0,0 +1,243 @@
...
+	updStmt, _ := db.BuildUpdateStmt(new(Queue))
```
yhabteab: You don't need to update the whole tuple but only the `state` column, so using this statement should be enough:

```suggestion
	updStmt := `UPDATE event_queue SET state = :state WHERE id = :id`
```

internal/event/queue.go
```diff
@@ -0,0 +1,243 @@
...
+						if err != nil {
+							return database.CantPerformQuery(err, updStmt)
+						}
+
+						return nil
+					})
+			},
+			func(err error) bool {
+				if errors.Is(err, sql.ErrNoRows) {
+					return true
```
yhabteab, line 176: Shouldn't here be some kind of sleeps to prevent from an excessive CPU usages?

internal/event/queue.go
```diff
@@ -0,0 +1,243 @@
...
+		logger.Debugw("Claimed event queue entry for processing",
+			zap.Stringer("id", q.ID),
+			zap.Stringer("object_id", q.ObjectId))
+
+		if ev, err := q.toEvent(); err != nil {
+			logger.Errorw("Invalid event queue event cannot get decoded", zap.Stringer("id", q.ID), zap.Error(err))
+			q.State = queueStateError
+		} else if err = callback(ctx, logger, ev); err != nil {
+			logger.Errorw("Event queue event cannot be processed", zap.Stringer("id", q.ID), zap.Error(err))
+			q.State = queueStateError
```
yhabteab, line 197: I don't understand why introduced the error state. What's supposed to happen with that event in this case? If `incident.ProcessEvent` returns an error for various reasons, then there's nothing that can be done to fix that error but mark the event as processed and delete it from the DB just like the others. The `queueStateError` is documented as _is assigned to invalid events for further processing._, but who is supposed to process it further?

internal/event/event_queue.go
yhabteab: > In my mental picture, the excluding JSON struct tags for `Time` and `SourceId` were safety features, ensuring that these daemon-assigned properties cannot be forged by a submitting source - either fraudulent or by accident.

That's fair but wouldn't it be better to just wrap the IGL `Event` type internally then? You could then make the source ID JSON [de]serializable there.

internal/listener/listener.go
```diff
@@ -222,20 +222,35 @@ func (l *Listener) ProcessEvent(w http.ResponseWriter, r *http.Request) {
...
-		l.abort(w, http.StatusInternalServerError, src, "event could not be processed successfully, see server logs for details")
+	// Enqueue the event restricted to the HTTP request context, and restrict it further to well-guessed three seconds.
+	//
+	// Doing so reduces the chance that event queueing works, while the HTTP client disappears. Unfortunately, this
+	// cannot be done in reverse - like with a callback, writing the HTTP response within the transaction -, as at this
+	// moment it is unknown if the transaction will succeed. This approach binds the transaction to the request context
+	// and has a minimum interval between COMMIT and writing back a response.
+	//
+	// Of course, the client might disappear without disconnecting. Unless we have some super aggressive TCP heartbeat,
+	// this will fall through. But, for the moment, I am totally fine with this.
```
yhabteab, line 233: Please explain the new behavior/workaround that should help to prevent such inconsistencies instead.

internal/listener/listener.go
```diff
@@ -222,20 +222,35 @@ func (l *Listener) ProcessEvent(w http.ResponseWriter, r *http.Request) {
...
+	ctx, cancel := context.WithTimeout(r.Context(), 3*time.Second)
+	defer cancel()
+
+	err := event.Enqueue(ctx, l.db, &ev, object.Get(l.db, &ev).ID)
```
yhabteab: You don't have to construct an object for this!

```suggestion
	err := event.Enqueue(ctx, l.db, &ev, object.ID(ev.SourceId, ev.Tags))
```

schema/mysql/upgrades/006.sql
```diff
@@ -0,0 +1,18 @@
...
+
+    version varchar(64) NOT NULL, -- From submitting client; allows migrations after upgrades.
+    state smallint NOT NULL DEFAULT 0, -- pendig (0), processing (1), done (2), or error (64).
+
+    CONSTRAINT pk_event_queue PRIMARY KEY (id)
+);
+
+CREATE INDEX idx_event_queue_time ON event_queue (time);
+CREATE INDEX idx_event_queue_object_id ON event_queue (object_id);
+CREATE INDEX idx_event_queue_state ON event_queue (state);
```
yhabteab: These two columns are used as combined where filter by the query, so adding an index for each of them separately makes little sense.

```suggestion
CREATE INDEX idx_event_queue_state_object_id ON event_queue (state, object_id);
```

schema/mysql/upgrades/006.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE event_queue (
+    id binary(32) NOT NULL, -- SHA256 of JSON representation.
+
+    json text NOT NULL COLLATE utf8mb4_unicode_ci,
```
yhabteab: Using `text` as type here won't be enough as Icinga DB's `output` and `long_output` have each `longtext` as a type and we combine both of them when submitting the events to Icinga Notifications. So, I don't know whether using `longtext` types would even be enough here and we might need to use `BLOB`, `BLOBMEDIUM` etc. instead. Also, do you really need to explicitly set the collation?

yhabteab (commented):
Since the nodes will process the events in parallel the incident state can be mutated by all of them as well. So, you must restore the state for each existing incident from the database before processing the event. See here for example!
https://github.com/Icinga/icinga-notifications/blob/f8477911c92e5cd699c5611abb0fe132a9cc2f8f/internal/incident/incident.go#L162-L167

This must be extended to restore all the other states as well!

internal/event/queue.go
oxzi: Honestly, no. I rebuilt `DB.BuildSelectStmt` and didn't really considered that `sqlx.Get` scans the struct. Changed.

internal/event/queue.go
oxzi: Haven't much thought about it as I expected the `event_queue` not grow too much. But a valid point.

As I was curious, I compared the query planner for both queries on a recent MySQL and MariaDB. There is a different strategy on MariaDB, but I am unsure how expensive this would be.

Nevertheless, changed to an antijoin. Thanks.

---

### MySQL 9.7.0

```
mysql> EXPLAIN SELECT eq.* FROM event_queue eq WHERE     eq.state = 0 AND     NOT EXISTS (         SELECT 1         FROM event_queue sub         WHERE             sub.state = 1 AND             sub.object_id = eq.object_id     ) ORDER BY eq.time LIMIT 1\G
*************************** 1. row ***************************
EXPLAIN: -> Limit: 1 row(s)  (cost=2.2 rows=1)
    -> Nested loop antijoin  (cost=2.2 rows=1)
        -> Sort: eq.`time`  (cost=1.1 rows=1)
            -> Index lookup on eq using idx_event_queue_state (state = 0)  (cost=1.1 rows=1)
        -> Filter: (sub.object_id = eq.object_id)  (cost=1.1 rows=1)
            -> Index lookup on sub using idx_event_queue_state (state = 1)  (cost=1.1 rows=1)

1 row in set, 1 warning (0.001 sec)

mysql> EXPLAIN SELECT q1.* FROM event_queue q1 LEFT JOIN event_queue q2     ON q1.object_id = q2.object_id     AND q2.state = 1 WHERE     q1.state = 0 AND     q2.object_id IS NULL ORDER BY q1.time LIMIT 1\G
*************************** 1. row ***************************
EXPLAIN: -> Limit: 1 row(s)  (cost=2.2 rows=0.1)
    -> Filter: (q2.object_id is null)  (cost=2.2 rows=0.1)
        -> Nested loop antijoin  (cost=2.2 rows=0.1)
            -> Sort: q1.`time`  (cost=1.1 rows=1)
                -> Index lookup on q1 using idx_event_queue_state (state = 0)  (cost=1.1 rows=1)
            -> Filter: (q2.object_id = q1.object_id)  (cost=1.01 rows=0.1)
                -> Index lookup on q2 using idx_event_queue_state (state = 1)  (cost=1.01 rows=1)

1 row in set (0.001 sec)
```

### MariaDB 12.3.2

```
MariaDB [notifications]> EXPLAIN SELECT eq.* FROM event_queue eq WHERE     eq.state = 0 AND     NOT EXISTS (         SELECT 1         FROM event_queue sub         WHERE             sub.state = 1 AND             sub.object_id = eq.object_id     ) ORDER BY eq.time LIMIT 1\G
*************************** 1. row ***************************
           id: 1
  select_type: PRIMARY
        table: eq
         type: index
possible_keys: idx_event_queue_state
          key: idx_event_queue_time
      key_len: 8
          ref: NULL
         rows: 1
        Extra: Using where
*************************** 2. row ***************************
           id: 2
  select_type: DEPENDENT SUBQUERY
        table: sub
         type: index_subquery
possible_keys: idx_event_queue_object_id,idx_event_queue_state
          key: idx_event_queue_object_id
      key_len: 32
          ref: func
         rows: 1
        Extra: Using where
2 rows in set (0.001 sec)

MariaDB [notifications]> EXPLAIN SELECT q1.* FROM event_queue q1 LEFT JOIN event_queue q2     ON q1.object_id = q2.object_id     AND q2.state = 1 WHERE     q1.state = 0 AND     q2.object_id IS NULL ORDER BY q1.time LIMIT 1\G
*************************** 1. row ***************************
           id: 1
  select_type: SIMPLE
        table: q1
         type: index
possible_keys: idx_event_queue_state
          key: idx_event_queue_time
      key_len: 8
          ref: NULL
         rows: 1
        Extra: Using where
*************************** 2. row ***************************
           id: 1
  select_type: SIMPLE
        table: q2
         type: ref
possible_keys: idx_event_queue_object_id,idx_event_queue_state
          key: idx_event_queue_object_id
      key_len: 32
          ref: notifications.q1.object_id
         rows: 1
        Extra: Using where; Not exists
2 rows in set (0.001 sec)
```

internal/event/queue.go
oxzi: It is in the next parameter of `retry.WithBackoff`, the `b backoff.Backoff`: `backoff.NewExponentialWithJitter(time.Second, 3*time.Second)`. As `sql.ErrNoRows` returns `true`, the `backoff.Backoff` is called, resulting in a `select` with a `time.After(b(attempt))`. So, empty queries will wait up to three seconds before being retried.

internal/event/queue.go
oxzi: > who is supposed to process it further

At the time of writing, it was a future consideration - and still is.

It felt wrong to delete failing events like they would have succeeded. By adding this error state, they are at least present, because nobody checks log files unless something is utterly wrong. By having this state, they could be made visible, for example, in Icinga Notifications Web via a banner indicating that there are $n$ failed events, similar to the banner Icinga DB Web shows for connection issues.

internal/event/queue.go
yhabteab: I just basically compared the index both query use, the number of rows they need to examine and what's included in the [`Extra` column](https://mariadb.com/docs/server/reference/sql-statements/administrative-sql-statements/analyze-and-explain-statements/explain#extra-column) (see the `Not exists` part). I think, the difference will only be very clear if you actually have some data in the table.

**Note** that I've reordered the JOIN conditions, so that the index I suggested in https://github.com/Icinga/icinga-notifications/pull/443#discussion_r3489904327 can fully be used.

internal/event/queue.go
yhabteab: Dong! Obviously, I completely missed that! Sorry.

internal/listener/listener.go
oxzi: Regarding the last paragraph: As we experimented the other day, there is no way to detect this scenario. If the client disappears w/o properly closing the connection, the TCP sessions stays alive, the context is not canceled and even writing to the socket succeeds, as the kernel buffers and retries this.

Or what are you referring to?

yhabteab: > Or what are you referring to?

Yes, I'm referring to the last paragraph (especially the last sentence). I'd either drop it entirely or explain why it's acceptable to still queue the event into the database.

schema/mysql/upgrades/006.sql
oxzi: Changed `text` type to `longtext` for the MySQL schema. I just thought about PostgreSQL, where `text` is of unlimited length. MySQL's `longtext` allows up to $2^{32}-1$ chars (or bytes, at least). If someone exceeds this in their output, I am pretty sure there are already other issues.

Regarding the `COLLATE`: I have just blindly copied it from an earlier entry, as I was not sure. But then we also have `text` (and `longtext`) fields without it, so I dropped it.

yhabteab: > If someone exceeds this in their output, I am pretty sure there are already other issues.

This isn't just about the output, the `Relations` field can become huge too. Remember, Icinga DB currently sends all the custom vars of a given checkable if you use a single one as an event rule filter. Unless, we want to change that, the limit will sooner or later be triggered on some large envs.

> Regarding the `COLLATE`: I have just blindly copied it from an earlier entry, as I was not sure. But then we also have `text` (and `longtext`) fields without it, so I dropped it.

Because all the existing tables define their encoding and collation on the table level and this new one doesn't.

internal/listener/listener.go
oxzi: Second thought: I have added a comment about the internal uniqueness due to the event ID. (If this was what you were thinking about.)

_Edit: Just saw your prior answer after reloading the tab. Honestly, GitHub comments are quite broken :/_

oxzi:
Thanks for the review. I have addressed all comments and pushed a rebased version. Furthermore, I changed the following:

- Synchronize incidents with the database, as [suggested](https://github.com/Icinga/icinga-notifications/pull/443#pullrequestreview-4589598401).
- Used `main` and `run` pattern from Icinga DB to run `defer` functions, which were ignored by `os.Exit(1)`.
- Integrated retention start in `errorgroup`.
- Changed `event_queue.version` to `user_agent`, including software name as Icinga Notifications Web considers populating this as well.
- Added optional additional `WHERE` clause to the Retention for usage in the event_queue retention.

cmd/icinga-notifications/main.go
```diff
@@ -65,19 +78,68 @@ func main() {
...
+			return err
+		}
+	})
+
+	eg.Go(func() error {
+		err := event.ListenQueue(
+			ctx,
+			db,
+			logs,
+			func(ctx context.Context, logger *logging.Logger, ev *event.Event) error {
```
yhabteab: Since the `logs` is needed just for doing `logs.GetChildLogger("event-queue")` call from within `ListenQueue`, why not make that call here and pass the `logger` as an argument to `ListenQueue` and drop the extra `*logging.Logger` param from the callback since you would've directly access to the `event-queue` logger here.

doc/05-Distributed-Setups.md
```diff
@@ -1,3 +1,38 @@
...
+
+## Configuration
+
+To run additional instances, install Icinga Notifications on another host and point it at the same database.
+No additional configuration is required beyond what is described in [Configuration](03-Configuration.md).
+
+Each instance must be able to reach the database and requires any configured notification channels to be installed.
+The HTTP API of each instance operates independently.
+If you place a load balancer in front of multiple instances,
+incoming events will be distributed across them and processed from the shared queue.
```
yhabteab, line 28: Sorry, but I failed to see how mentioning a load balancer here adds any value. You've explained how the queue is processed in the `Architecture` section, so mentioning a load balancer here is completely unnecessary IMHO.

doc/05-Distributed-Setups.md
```diff
@@ -1,3 +1,38 @@
...
+
+## Corrupt Queue Entries
+
+The event queue is designed to be an internal API, where no external process is allowed to insert events.
```
yhabteab, line 32: That's not true. You've even prepared the table to work with external users too.

doc/05-Distributed-Setups.md
```diff
@@ -1,3 +1,38 @@
...
+Thus, it is expected to only contain valid data.
+
+Nevertheless, in the unlikely case that corrupted data crept into the `event_queue` table,
+Icinga Notifications will result in an error and might even stop.
+The log contains a reference to the invalid `event_queue` row including its `id`.
+This allows inspecting the entry in the relational database and eventually fixing or deleting it.
```
yhabteab, line 38: I'm not yet convinced with all of this to be honest, but if the consensus (/cc @nilmerg) is to keep failed events in the database then please rephrase it a bit because as far I can see Icinga Notifications should not stop working just because you submitted an invalid event.

internal/event/queue.go
```diff
@@ -0,0 +1,222 @@
...
+
+			return nil
+		})
+}
+
+// ListenQueue fetches events from the database event queue and forwards them to the callback.
+//
+// This function blocks until either an error occurred or the passed context is finished. In either case, an error is
+// being returned.
+func ListenQueue(
```
yhabteab: Wouldn't `ProcessQueue` better fit as a name for this?

internal/event/queue.go
```diff
@@ -0,0 +1,222 @@
...
+
+	selStmt := db.Rebind(`
+		SELECT q1.*
+		FROM event_queue q1
+		LEFT JOIN event_queue q2 ON
+			q1.object_id = q2.object_id AND
+			q2.state = ` + fmt.Sprintf("%d", queueStateProcessing) + `
+		WHERE
+			q1.state = ` + fmt.Sprintf("%d", queueStatePending) + ` AND
+			q2.object_id IS NULL
```
yhabteab, line 145: Either change the order of the `idx_event_queue_object_id_state` INDEX columns to `(state, object_id)` and reorder the JOIN conditions accordingly or you've to reorder these two where conditions. I've already updated my [original comment](https://github.com/Icinga/icinga-notifications/pull/443#discussion_r3511295214) on the subquery to match the suggested [INDEX column oderders](https://github.com/Icinga/icinga-notifications/pull/443#discussion_r3489904327) but you didn't seem to have seen it.

internal/incident/incident.go
```diff
@@ -701,12 +713,45 @@ func (i *Incident) getRecipientsChannel(t time.Time) rule.ContactChannels {
...
-// restoreRecipients reloads the current incident recipients from the database.
-// Returns error on database failure.
-func (i *Incident) restoreRecipients(ctx context.Context) error {
-	contact := &ContactRow{}
+// restoreState reloads all incident state from the database.
+//
+// This method must be called for existing Incidents before processing an event, as another node in an HA setup may
+// have altered the state in the meantime.
+func (i *Incident) restoreState(ctx context.Context) error {
+	stmt := i.db.Rebind(i.db.BuildSelectStmt(new(Incident), new(Incident)) + ` WHERE "id" = ?`)
```
yhabteab:
```suggestion
	stmt := i.db.Rebind(i.db.BuildSelectStmt(i, i) + ` WHERE "id" = ?`)
```

internal/incident/incident.go
```diff
@@ -701,12 +713,45 @@ func (i *Incident) getRecipientsChannel(t time.Time) rule.ContactChannels {
...
+	if err := i.db.GetContext(ctx, i, stmt, i.Id); err != nil {
+		i.logger.Errorw("Failed to restore incident row from the database", zap.Error(err))
+		return err
+	}
+
+	var states []*EscalationState
+	err := i.db.SelectContext(
```
yhabteab: You could restore this and the contact rows below in the same way as in the `LoadOpenIncidents` via the `utils.ForEachRow` helper function.

internal/incident/incidents.go
```diff
@@ -199,6 +200,37 @@ func GetCurrentIncidentsForSource(sourceID int64) []*Incident {
...
+func loadOpenIncidentForObject(
+	ctx context.Context,
+	db *database.DB,
+	logger *logging.Logger,
+	runtimeConfig *config.RuntimeConfig,
+	obj *object.Object,
+) error {
+	i := NewIncident(db, obj, runtimeConfig, logger.With(zap.String("object", obj.DisplayName())))
+
+	stmt := db.Rebind(db.BuildSelectStmt(new(Incident), new(Incident)) + ` WHERE "recovered_at" IS NULL AND "object_id" = ?`)
```
yhabteab:
```suggestion
	stmt := db.Rebind(db.BuildSelectStmt(i, i) + ` WHERE "recovered_at" IS NULL AND "object_id" = ?`)
```

internal/incident/incidents.go
```diff
@@ -199,6 +200,37 @@ func GetCurrentIncidentsForSource(sourceID int64) []*Incident {
...
+	err := db.GetContext(ctx, i, stmt, obj.ID)
+	if err == sql.ErrNoRows {
```
yhabteab:
```suggestion
	if errors.Is(err, sql.ErrNoRows) {
```

internal/retention/prune.go
```diff
@@ -23,7 +23,8 @@ type TimeBoundPruner struct {
 	PK         string
 	TimeColumn string
 
-	Referrers []ReferencingRowPruner
+	Referrers      []ReferencingRowPruner
+	WhereCondition string
```
yhabteab: Did you consider using a composable `Condition` type instead of such a simple string? The condition could then for example, return its SQL stmt along with its respective arguments when finally assembled. That way you could provide any condition you want and also wouldn't be limited to just using constant values as arguments but also anything you want since you're going to provide the args separately they'll be prepared just like the other ones when finally executing it. The type could roughly look like this:

```go
type Condition struct {
	Stmt  string
	Args []any
}

func (c Condition) Get() (string, []any) { return r.Stmt, r.Args }
```

You can take this even further and allow to define multiple `Condition`s with a user provided `AND` or `OR` operators even if it's not necessary yet.

schema/mysql/upgrades/ha-init.sql
yhabteab, line 1: Pls use a descriptive name instead! We don't use such numeric names anymore :).

schema/mysql/schema.sql
```diff
@@ -436,6 +436,23 @@ CREATE TABLE incident_history (
...
 CREATE INDEX idx_incident_history_time_type ON incident_history(time, type) COMMENT 'Incident History ordered by time/type';
 
+CREATE TABLE event_queue (
+    id binary(32) NOT NULL, -- SHA256 of JSON representation.
+
+    json longtext NOT NULL,
+    time bigint NOT NULL,
+    object_id binary(32) NOT NULL, -- No foreign key, object might not exist at this point.
+
+    user_agent varchar(64) NOT NULL, -- From submitting client; allows migrations after upgrades.
```
yhabteab: I don't think 64 chars are enough for a user agent, see the `browser_session` table just few lines below for example.

schema/mysql/schema.sql
```diff
@@ -436,6 +436,23 @@ CREATE TABLE incident_history (
...
+    state smallint NOT NULL DEFAULT 0, -- pending (0), processing (1), done (2), or error (64).
+
+    CONSTRAINT pk_event_queue PRIMARY KEY (id)
+);
```
yhabteab: This table is missing ENGINE, encoding, collation etc. See all the existing tables for example.

internal/incident/incident.go
```diff
@@ -241,6 +243,11 @@ func (i *Incident) RetriggerEscalations(ev *event.Event) {
 	i.runtimeConfig.RLock()
 	defer i.runtimeConfig.RUnlock()
 
+	if err := i.restoreState(context.Background()); err != nil {
+		i.logger.Errorw("Cannot restore incident state for escalation reevaluation", zap.Error(err))
+		return
+	}
```
yhabteab: This also should check the before and after state of the incident. Normally, in a non-HA setup each ongoing event targeting the same incident cancels the scheduled timer (if any). However, since we now can't guarantee that an incident will always be bound to a specific node, there can be now a case, that subsequent events were processed on some other node just after we've scheduled the timer on node 1 for example. So, if the incident's state has changed here, this should be a no-op, otherwise we might send duplicate notifications.

oxzi: Just comparing if something has changed here does not fix the race. Consider the following with two nodes:

1. The `Incident.timer` on both nodes fires at round about the same time.
2. Both nodes would call `i.RetriggerEscalations` and `i.restoreState` within, where `i.restoreState` returns the same state on both nodes.
3. `i.evaluateEscalations` returns the same escalations on both nodes.
4. Both nodes are sending out notifications.

As the whole incident logic feels very locally bound, I moved this to a single transaction which locks an Incident via `SELECT ... FOR UPDATE`.

internal/incident/incident.go
oxzi: In general, good idea. However, after I [rebuilt this to use transactions](https://github.com/Icinga/icinga-notifications/pull/443/changes#r3513770513), `utils.ForEachRow` does not work anymore. And due to time constraints, I don't want to add this to the IGL first.

doc/05-Distributed-Setups.md
oxzi: This is called an example, and I placed it here to ensure sysadmins that they can route incoming requests through a round robin load balancer. Try to take the perspective of a sysadmin, not a developer. The _Architecture_ section hints that all nodes are equal, but does not directly answer the question: How to route packages through your network.

doc/05-Distributed-Setups.md
oxzi: You have to distinguish between different external users. After @nilmerg's request, Icinga Notifications Web is going to write to this queue as well. While this might be an alien from our core team's point of view, it is still part of the Icinga Notifications project. However, the event queue is still no free lunch for everybody and we certainly don't want users to put things there. It is an internal API; and this text is the user's documentation.

To clarify: The `user_agent` field is to distinguish different Icinga Notifications versions and to detect Icinga Notifications Web versions. Nothing else. It is there to allow migrations and since Icinga Notifications is no longer the only producer, `version` was no longer an adequate name.

doc/05-Distributed-Setups.md
oxzi: At the moment, there is no consensus. There is just a dissent between the two of us. And @nilmerg has not shared his opinion so far.

Last time, I stated my reasons in [this comment](https://github.com/Icinga/icinga-notifications/pull/443#discussion_r3507081236). Of course, I can also start dropping invalid events from the queue, but - as I wrote earlier - this will result in people to miss things, because a single error log entry mostly goes unnoticed.

For the time being, I have updated this section. Thanks for finding the outdated information.

internal/event/queue.go
oxzi: Good catch!

And no, I haven't seen the update. Most likely because GitHub does not reload comments while one has the page open.

internal/retention/prune.go
oxzi: I have chosen such a simple string because it does all I need. As the deadline for this PR is tomorrow, I didn't want to become overly artistic with it and just get it working without looking like a dirty hack. If in the future the need for more customisation arises, I am all in to refactor this into something looking like straight outta a design pattern book.

Nevertheless, I did a small refactoring to let the code look a bit less hacky.

schema/mysql/upgrades/ha-init.sql
oxzi: I was already expecting this comment :D

schema/mysql/schema.sql
oxzi: The `browser_session.user_agent` is a web browser user agent, which is usually ridiculously long. Here, it is the user agent of the submitting source, which should be either `icinga-notifications/$VERSION` or `icinga-notifications-web/$VERSION`.

- See my earlier comment: https://github.com/Icinga/icinga-notifications/pull/443/changes#r3514073861
- And the current code: https://github.com/Icinga/icinga-notifications/blob/847f4539522ee51fba8788471ea560374e162604/internal/event/queue.go#L102

Even if web would include the whole `git-describe(1)` information, there would still be space left. However, the whole SHA1 commit hash would be too long.

```
$ echo -n 'icinga-notifications-web/v0.2.0-80-g637d4b8f' | wc -c
44
$ echo -n 'icinga-notifications-web/637d4b8feb368c2f52996b93ba534b9824a7132a' | wc -c
65
```

So, I have enlarged it to 255 chars. If someone exceeds it, they are at least from "within" and we can talk to them.

schema/mysql/schema.sql
oxzi: That's happens when writing PostgreSQL first and adjusting types second.

internal/event/queue.go
yhabteab: > Either change the order of the `idx_event_queue_object_id_state` INDEX columns to `(state, object_id)` and reorder the JOIN conditions accordingly

Now, you only did the latter but you still didn't reorder the columns in the SQL scripts.

internal/retention/prune.go
yhabteab: > I am all in to refactor this into something looking like straight outta a design pattern book.

sarcasm? If you don't intent to change it that way, then please at least rename the field to something like `ExtraCondition` instead since this is not the only where condition used for the queries.

internal/incident/incident.go
yhabteab: Ty.

internal/retention/retention.go
```diff
@@ -103,4 +113,32 @@ var dbPruners = []TimeBoundPruner{
...
 	},
+	// Extra pruners for the event_queue.
+	{
+		// Events being processed too long - implies crashed daemon.
+		Table:            "event_queue",
+		PK:               "id",
+		TimeColumn:       "time",
+		WhereCondition:   "state = 1",
+		OverridePeriod:   15 * time.Minute,
+		OverrideInterval: time.Minute,
```
yhabteab: I woud set the interval here to the same value as for period. I don't think, the goroutine needs to run that frequently. Same goes for the below pruners.

schema/mysql/upgrades/ha-init.sql
```diff
@@ -0,0 +1,16 @@
...
+
+    user_agent varchar(255) NOT NULL, -- From submitting client; allows migrations after upgrades.
+    state smallint NOT NULL DEFAULT 0, -- pending (0), processing (1), done (2), or error (64).
+
+    CONSTRAINT pk_event_queue PRIMARY KEY (id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE INDEX idx_event_queue_time ON event_queue (time);
+CREATE INDEX idx_event_queue_time_state ON event_queue (time, state);
+CREATE INDEX idx_event_queue_object_id_state ON event_queue (object_id, state);
```
yhabteab: As commented in https://github.com/Icinga/icinga-notifications/pull/443#discussion_r3518294768, you need to change this index columns orders like this:

```suggestion
CREATE INDEX idx_event_queue_state_object_id ON event_queue (state, object_id);
```

yhabteab (commented):
Apart from the these two nitpicks, this PR seems to work. Johannes is not in office today, so I don't know how to proceed with the error state thing.

internal/event/queue.go
oxzi: Jikes. Thanks for double checking it.

internal/retention/prune.go
oxzi: > sarcasm?

Maybe a bit, sorry.

I have also renamed the field.

oxzi (commented):
Thanks for the feedback.

---

@yhabteab: As we discussed offline, I kept the `error` state for the moment and this PR. If @nilmerg disagrees, we can just remove it in a future PR.

I would kindly ask you to take another look.

yhabteab (approved):
Let's hope we didn't miss something 🤞! :shipit:

PR #446, opened 2026-06-22
Support a configurable retention policy

### LINKED ISSUE #391: Reduce data persistence to an absolute minimum
At the moment there are two growing datasets: *events* and *incidents*

- *events* are stored in the database table `event`
- *incidents* are stored in the database table `incident`

Events are transmitted by sources to Icinga Notifications in order to be processed by _event rules_ and be mapped to object specific incidents. There's no real need to store events in the database any longer. Incidents on the other hand need a configurable retention interval for automatic cleanup of closed incidents.

### LINKED ISSUE #393: Support a configurable retention policy for closed incidents
Closed incidents do not serve any significance once they're of a given age. It should be possible to configure a retention interval and have Icinga Notifications automatically clean up incidents accordingly. This interval must have a reasonable default configured.

Note that this setting should be configurable in the UI, as such it needs to be stored in the database.

yhabteab:
This PR adds a configurable retention policy to periodically cleanup historical data. It's designed to be as generic as possible to support all kind of tables (e.g. we might also want to purge all entries from everywhere marked as `deleted = y` in the future) even though we only need this for the `incident` and all its related tables right now. It's similar to Icinga DB's retention policy but not necessarily the same, so I didn't try to deduplicate it into IGL.

resolves #393
resolves #391

### DIFF
--- a/cmd/icinga-notifications/main.go
+++ b/cmd/icinga-notifications/main.go
@@ -2,6 +2,11 @@ package main
 
 import (
 	"context"
+	"errors"
+	"os/signal"
+	"syscall"
+	"time"
+
 	"github.com/icinga/icinga-go-library/database"
 	"github.com/icinga/icinga-go-library/logging"
 	"github.com/icinga/icinga-go-library/utils"
@@ -11,10 +16,8 @@ import (
 	"github.com/icinga/icinga-notifications/internal/daemon"
 	"github.com/icinga/icinga-notifications/internal/incident"
 	"github.com/icinga/icinga-notifications/internal/listener"
+	"github.com/icinga/icinga-notifications/internal/retention"
 	"github.com/okzk/sdnotify"
-	"os/signal"
-	"syscall"
-	"time"
 )
 
 func main() {
@@ -61,6 +64,13 @@ func main() {
 	if err != nil {
 		logger.Fatalf("Cannot load incidents from database: %+v", err)
 	}
+
+	ret := retention.New(db, logs.GetChildLogger("retention"))
+	go func() {
+		if err := ret.Run(ctx); err != nil && !errors.Is(err, context.Canceled) {
+			logger.Fatalf("Retention has finished with an error: %+v", err)
+		}
+	}()
 
 	// When Icinga Notifications is started by systemd, we've to notify systemd that we're ready.
 	_ = sdnotify.Ready()
--- a/doc/03-Configuration.md
+++ b/doc/03-Configuration.md
@@ -103,6 +103,57 @@ For environment variables, each option is prefixed with `ICINGA_NOTIFICATIONS_DA
 | max_rows_per_transaction       | **Optional.** Maximum number of rows Icinga Notifications is allowed to `SELECT`,`DELETE`,`UPDATE` or `INSERT` in a single transaction. Defaults to `8192`. |
 | wsrep_sync_wait                | **Optional.** Enforce [Galera cluster](#galera-cluster) nodes to perform strict cluster-wide causality checks. Defaults to `7`.                             |
 
+## Retention Configuration
+
+You can configure the data retention and cleanup behavior of Icinga Notifications to automatically remove old data
+from the database after a certain period. This helps to manage the size of the database and ensure that it does not
+grow indefinitely.
+
+For YAML configuration, the options are part of the `retention` dictionary.
+For environment variables, each option is prefixed with `ICINGA_NOTIFICATIONS_RETENTION_`.
+
+| Option     | Description                                                                                                                                                                                                        |
+|------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| period     | **Optional.** Retention period defined as [duration string](#duration-string). Defaults to `0s` (disabled). Data older than this period will be automatically deleted from the database.                           |
+| interval   | **Optional.** Interval for periodic retention checks defined as [duration string](#duration-string). Defaults to `1h`.                                                                                             |
+| batch_size | **Optional.** Maximum number of rows to delete in a single transaction during retention cleanup. Defaults to `5000`.                                                                                               |
+| options    | **Optional.** Map of component name to retention period in order to set a different retention period for each component instead of the default one. See [retention components](#retention-components) for details. |
+
+!!! info
+
+    By default, Icinga Notifications does not delete any data from the database, since the default retention period is
+    `0s`, which means that all data is retained indefinitely. You can set the `period` option to a non-zero value to
+    enable automatic data cleanup after the specified period has passed.
+
+If you don't want to enable automatic data cleanup for all components, you can let the `period` option be `0s` and set
+non-zero retention periods for individual components in the `options` as [described below](#retention-components).
+This goes also for the other way around, if you want to enable automatic data cleanup for all components, you can set
+the `period` option to a non-zero value and set `0s` retention periods for a specific component to disable automatic
+data cleanup for that component.
+
+### Retention Components
+
+In addition to the default retention period, you can set a different retention period for each component of
+Icinga Notifications. This allows you to keep certain types of data for a longer or shorter period than others,
+depending on their importance or relevance.
+
+For YAML configuration, the options are part of the `retention.options` dictionary.
+For environment variables, each option is prefixed with `ICINGA_NOTIFICATIONS_RETENTION_OPTIONS_` and expects
+a [duration string](#duration-string) as value.
+
+Currently, the following components are available:
+
+| Component | Description                                                           |
+|-----------|-----------------------------------------------------------------------|
+| incident  | Incidents and all related data, such as their history, contacts, etc. |
+
+!!! info
+
+    Incidents that are still open are not eligible for retention cleanup, even if they are open for months or years.
+    The retention policy only applies to closed incidents, and the retention period is computed from the time they were
+    closed, not from the time they were opened. If you want to clean up some incidents that are still open too, you need
+    to close them first, and then wait for the retention period to pass.
+
 ## Logging Configuration
 
 Configuration of the logging component used by Icinga Notifications.
@@ -151,6 +202,7 @@ ICINGA_NOTIFICATIONS_LOGGING_OPTIONS=database:error,listener:debug
 | database        | Database connection status and queries.                                   |
 | incident        | Incident management and changes.                                          |
 | listener        | HTTP listener for event submission and debugging.                         |
+| retention       | Data retention and cleanup.                                               |
 | runtime-updates | Configuration changes through Icinga Notifications Web from the database. |
 
 ## Appendix
--- a/internal/retention/prune.go
+++ b/internal/retention/prune.go
@@ -1,0 +1,169 @@
+package retention
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/icinga/icinga-go-library/backoff"
+	"github.com/icinga/icinga-go-library/com"
+	"github.com/icinga/icinga-go-library/database"
+	"github.com/icinga/icinga-go-library/retry"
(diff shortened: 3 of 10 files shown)


yhabteab:
@cla-bot check

doc/03-Configuration.md
```diff
@@ -152,6 +184,7 @@ ICINGA_NOTIFICATIONS_LOGGING_OPTIONS=database:error,listener:debug
 | incident        | Incident management and changes.                                          |
 | listener        | HTTP listener for event submission and debugging.                         |
 | runtime-updates | Configuration changes through Icinga Notifications Web from the database. |
+| retention       | Data retention and cleanup.                                               |
```
oxzi:
```suggestion
| retention       | Data retention and cleanup.                                               |
| runtime-updates | Configuration changes through Icinga Notifications Web from the database. |
```

Let's keep it sorted. Same for the yaml file.

doc/03-Configuration.md
```diff
@@ -103,6 +103,38 @@ For environment variables, each option is prefixed with `ICINGA_NOTIFICATIONS_DA
...
+You can configure the data retention and cleanup behavior of Icinga Notifications to automatically remove old data
+from the database after a certain period. This helps to manage the size of the database and ensure that it does not
+grow indefinitely.
+
+For YAML configuration, the options are part of the `retention` dictionary.
+For environment variables, each option is prefixed with `ICINGA_NOTIFICATIONS_RETENTION_`.
+
+| Option     | Description                                                                                                                                                                                                        |
+|------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| period     | **Optional.** Retention period defined as [duration string](#duration-string). Defaults to `"30d"`. Data older than this period will be automatically deleted from the database.                                   |
```
oxzi: `30d` is not a valid Go duration, as you have documented in the YAML's comment. However, people might get tricked into using the `d` unit based on this string, especially when using environment variables, where the extra documentation is lacking.

internal/retention/prune.go
```diff
@@ -0,0 +1,168 @@
...
+			db.GetDefaultRetrySettings(),
+		)
+		if err != nil {
+			return 0, err
+		}
+
+		counter.Add(affected)
+		if affected < limit {
+			break
+		}
```
oxzi, line 165: At least for `ReferencingRowPruner.Exec`, this comparison will result in one unnecessary `DELETE` statement which will affect no entries.

https://github.com/Icinga/icinga-notifications/blob/5b53719084450582b9bbd3334dd400a1bd2f47f5/internal/retention/prune.go#L119

By setting `limit` to the length of `args`, `affected` should be equal to `limit`, resulting in the second `DELETE` statement with no more affected rows.

internal/retention/retention.go
```diff
@@ -0,0 +1,100 @@
...
+	errs := make(chan error, 1)
+
+	for _, pruner := range dbPruners {
+		period := conf.Period
+		if pruner.Table == "incident" && conf.Options.Incident != nil {
+			period = *conf.Options.Incident
+		}
+
+		if period == 0 {
+			r.logger.Debugf("Skipping retention for table %s because retention days is set to 0", pruner.Table)
```
oxzi:
```suggestion
			r.logger.Debugf("Skipping retention for table %s because retention period is set to 0", pruner.Table)
```

`period` is a Go `time.Duration`.

internal/retention/retention.go
```diff
@@ -0,0 +1,100 @@
...
+			}
+			r.logger.Logf(level, "Removed %d old items from table %s", deleted, pruner.Table)
+		}, periodic.Immediate())
+	}
+
+	select {
+	case err := <-errs:
+		return err
+	case <-ctx.Done():
+		return ctx.Err()
```
oxzi: When no pruner is running, this `select` waits here until the main SIGINT context is done. Not really harmful, but feels a bit wasteful, as we could just `return nil`.

doc/03-Configuration.md
```diff
@@ -103,6 +103,38 @@ For environment variables, each option is prefixed with `ICINGA_NOTIFICATIONS_DA
 | max_rows_per_transaction       | **Optional.** Maximum number of rows Icinga Notifications is allowed to `SELECT`,`DELETE`,`UPDATE` or `INSERT` in a single transaction. Defaults to `8192`. |
 | wsrep_sync_wait                | **Optional.** Enforce [Galera cluster](#galera-cluster) nodes to perform strict cluster-wide causality checks. Defaults to `7`.                             |
 
+## Retention Configuration
+
+You can configure the data retention and cleanup behavior of Icinga Notifications to automatically remove old data
+from the database after a certain period. This helps to manage the size of the database and ensure that it does not
```
oxzi, line 109: This first sentence reads like this is an optional feature one can enable. However, it is enabled by default with `period`'s default value of `720h`. Furthermore, it is not described how to disable this feature, which should work by setting `period` to `0` and to ensure that either `options` is empty or all `options` have a zero duration as well.

I would consider going with an opt-in instead, as I would still prefer a filling up disk to lost entries - and, btw, that's how we also do it for Icinga DB.

cmd/icinga-notifications/main.go
```diff
@@ -62,6 +63,13 @@ func main() {
 		logger.Fatalf("Cannot load incidents from database: %+v", err)
 	}
 
+	ret := retention.New(db, logs.GetChildLogger("retention"))
+	go func() {
+		if err := ret.Run(ctx); err != nil {
+			logger.Fatalf("Retention has finished with an error: %+v", err)
```
oxzi, line 71: While I was verifying if `logger.Fatalf` works from within a goroutine (it does, thanks), I stumbled about another thing.

The context is bound to a SIGINT signal, indicating that the user wants to stop the daemon. However, both this line as well as the already existing `listener.NewListener` will handle the `context.Canceled` error as a normal error, instead of the expected behavior. While the prior code only logged a confusing error, this one will exit with code `1`, which indicate an error or even result in a restart from some service manager.

internal/retention/retention.go
```diff
@@ -0,0 +1,100 @@
...
+
+// dbPruners defines the list of tables and their corresponding time columns to be pruned, along with any related
+// tables that need to be pruned in a cascading manner to maintain referential integrity. Each entry specifies the
+// main table to prune, its primary key, the time column used for determining which rows are old enough to delete,
+// and any referrer tables that have foreign key relationships with the main table.
+var dbPruners = []TimeBoundPruner{
+	{
+		Table:      "incident",
+		PK:         "id",
+		TimeColumn: "recovered_at",
```
oxzi, line 97: This might be worthy of documentation. At the moment, the docs are sounding like all incidents are going to be purged. However, only those where `recovered_at IS NOT NULL` are being taken into account.

doc/03-Configuration.md
yhabteab: Done.

doc/03-Configuration.md
yhabteab: Changed.

internal/retention/prune.go
yhabteab: Fixed.

internal/retention/retention.go
yhabteab: Fixed.

internal/retention/retention.go
yhabteab: Changed.

cmd/icinga-notifications/main.go
oxzi: As I was just reworking #443, feel free to take a look at the changes I made over there to the `main` method.

https://github.com/Icinga/icinga-notifications/blob/10fb93362d8f03b779064405706853c9cbaab4d5/cmd/icinga-notifications/main.go#L74-L119

doc/03-Configuration.md
yhabteab: I'm not sure whether you want me to still rephrase that sentence or not, but I only changed the retention to be disabled by default.

cmd/icinga-notifications/main.go
yhabteab: Fixed only for this new code part.

internal/retention/retention.go
yhabteab: Done.

yhabteab:
Addressed all your requested changes and also added an index for the `recovered_at` column on the `incident` table.

yhabteab:
Just rebased it to get the latest changes from the main branch.

doc/03-Configuration.md
```diff
@@ -103,6 +103,51 @@ For environment variables, each option is prefixed with `ICINGA_NOTIFICATIONS_DA
...
+grow indefinitely.
+
+For YAML configuration, the options are part of the `retention` dictionary.
+For environment variables, each option is prefixed with `ICINGA_NOTIFICATIONS_RETENTION_`.
+
+| Option     | Description                                                                                                                                                                                                        |
+|------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| period     | **Optional.** Retention period defined as [duration string](#duration-string). Defaults to `0s`. Data older than this period will be automatically deleted from the database.                                      |
```
oxzi:
```suggestion
| period     | **Optional.** Retention period defined as [duration string](#duration-string). Defaults to `0s` (disabled). Data older than this period will be automatically deleted from the database.                                      |
```

internal/retention/prune.go
```diff
@@ -0,0 +1,169 @@
...
+			return total, nil
+		}
+
+		for _, referrer := range tbp.Referrers {
+			if _, err := referrer.Exec(ctx, db, ids...); err != nil {
+				return 0, err
+			}
+		}
+
+		if affected, err := exec(ctx, db, deleteStmt, limit, ids...); err != nil {
```
oxzi: This `exec` call has the same [unnecessary `DELETE` bug I mentioned earlier](https://github.com/Icinga/icinga-notifications/pull/446/changes#r3453094174) - sorry, missed it back then.

Here, `ids` are populated by `selectStmt`, which is `tbp.assembleSelect(limit)`, a `SELECT ... LIMIT __limit__` query.


```suggestion
		if affected, err := exec(ctx, db, deleteStmt, 0, ids...); err != nil {
```

internal/retention/retention.go
```diff
@@ -0,0 +1,106 @@
...
+	for _, pruner := range dbPruners {
+		period := conf.Period
+		if pruner.Table == "incident" && conf.Options.Incident != nil {
+			period = *conf.Options.Incident
+		}
+
+		if period == 0 {
+			r.logger.Debugf("Skipping retention for table %s because retention period is set to 0", pruner.Table)
+			continue
+		}
```
oxzi, line 46: Overwriting the main _period_ with the _options_ value circumvents the logic to disable the retention. The current documentation and YAML comments states that a zero value for `retention.period` disables the retention altogether. However, when setting `retention.options.incident`, the retention is enabled.

So, I would suggest to either add a `conf.Period == 0` check before the loop or to document that one can enable certain retention jobs by setting `retention.options` regardless of `retention.period`. I am fine with both and, at least at the moment, with only one option, it does not really make a difference.

oxzi (commented):
In general, this PR looks fine - just two small things. Nevertheless, I have successfully tested this PR branch with a working retention configuration.

doc/03-Configuration.md
yhabteab: Changed.

internal/retention/prune.go
yhabteab: Fixed.

internal/retention/retention.go
yhabteab: I've documented the behavior.

oxzi (approved):

PR #447, opened 2026-06-22
Schema: add `incident_contact.changed_at` column

### LINKED ISSUE #445: Add `changed_at` to `incident_contact`
For the implementation of https://github.com/Icinga/icingadb-web/issues/1358#issuecomment-4682061584, it makes sense to add a `changed_at` column to the `incident_contact` table, otherwise the required timestamps can only be accesed via the `incident_history` table.

yhabteab:
resolves #445

### DIFF
--- a/internal/incident/db_types.go
+++ b/internal/incident/db_types.go
@@ -2,6 +2,7 @@ package incident
 
 import (
 	"context"
+
 	"github.com/icinga/icinga-go-library/database"
 	"github.com/icinga/icinga-go-library/notifications/event"
 	"github.com/icinga/icinga-go-library/types"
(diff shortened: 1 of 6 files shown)


nilmerg (changes requested):
It's fine as it is, but maybe go a little further?:

- Extract the latest change from the incident history during migration
- Add a comment in the schema explaining that the column is only necessary for the UI and named to match other occurrences

yhabteab:
> Extract the latest change from the incident history during migration

Is that really necessary?

yhabteab:
> > Extract the latest change from the incident history during migration
> 
> Is that really necessary?

I don't think this is really necessary, but changed it anyway, so pls have a look at it again.

nilmerg (approved):
It's *necessary* in terms of usefulness in development environments to not require a full wipe due to such a minimal change ;)

Thanks!

yhabteab:
> It's _necessary_ in terms of usefulness in development environments to not require a full wipe due to such a minimal change ;)

The initial version didn't require you to wipe out all your existing data, instead it just used now as a changed_at timestamp for all existing entries, which is not 100% correct but would've perfectly fine for that use case too.

PR #463, opened 2026-07-09
object, incident: Remove runtime cache

### LINKED ISSUE #457: Incidents: Move to relational database
As foreshadowed by the HA changes, incidents now exist in a mixed state between memory and resynchronization with the relational database. This is the worst of both worlds, as one still has to synchronize and runs the risk of missing it at some point. Currently, synchronization is lacking for some incident-related HTTP endpoints, for example. Therefore, we should transition to a database-only approach.

---

I have directly assigned myself, as I did the HA changes until last week. Will start later today with this.

oxzi:
With switching from a single node to a multi-node HA setup, storing both objects and incidents locally and only synchronizing them once from the database results in a diverged state between different nodes. Now, the local cache is gone and both objects and incidents are directly fetched from the database.

This required further changes, most prominently moving the incident reevaluation from a local timer to the database as well. So, a new next_escalation_check_at column was added to the incident table, which is scanned routinely.

Another change is switching mutexes for transactions, using the "SELECT ... FOR UPDATE" pattern to lock the current incident.

---

Refs #394. Fixes #457.


internal/incident/incident.go
```diff
@@ -34,39 +34,30 @@ type Incident struct {
...
-	Object *object.Object `db:"-"`
+	// NextEscalationCheckAt stores when this Incident's time-based escalations should be reevaluated next. It is used
+	// in the ReevaluateEscalations function to reevaluate incident escalations.
+	//
+	// For example, if there are escalations configured for incident_age>=1h and incident_age>=2h.
+	//   - If the incident is less than an hour old, this field contains the incident start time plus one hour.
+	//   - If the incident is between 1h and 2h old, this field contains the incident start time plus two hours.
+	//   - If the incident is already older than 2h, no future escalations can be reached solely based on the incident
+	//     aging, so the field is left nil.
+	NextEscalationCheckAt types.UnixMilli `db:"next_escalation_check_at"`
```
yhabteab, line 47: Incident age filter supports only [the `>` operator](https://github.com/Icinga/icinga-notifications-web/pull/500) now :).

internal/incident/incident.go
```diff
@@ -75,8 +66,23 @@ func NewIncident(
...
+	i.db = db
+	i.logger = logger
+	i.runtimeConfig = runtimeConfig
+	i.EscalationState = map[escalationID]*EscalationState{}
+	i.Rules = map[ruleID]struct{}{}
+	i.Recipients = map[recipient.Key]*RecipientState{}
+}
+
+// IncidentObject fetches the object.Object this incident belongs to from the database.
+func (i *Incident) IncidentObject(ctx context.Context) (*object.Object, error) {
```
yhabteab: Since there is no `Object` incident field anymore, I would use that name here instead. It was called before that way due to naming conflict with the now removed field.

internal/incident/incident.go
```diff
@@ -124,9 +135,6 @@ func (i *Incident) IsNotifiable(role ContactRole) bool {
 
 // ProcessEvent processes the given event for the current incident in an own transaction.
 func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event) error {
-	i.Lock()
-	defer i.Unlock()
-
 	i.runtimeConfig.RLock()
 	defer i.runtimeConfig.RUnlock()
```
yhabteab: After removing the incident lock, I think, we can also greatly reduce the scope of this remaining mutex by moving it down right before the `if ev.OpenOrEscalate() {` statement at line 194. Everything else performed before that if statement don't make use of this runtime config, so we're acquiring the lock here way too early.

internal/incident/incident.go
```diff
@@ -137,14 +145,30 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event) error {
...
+	obj := object.New(ev)
+	if err := obj.SyncFromEvent(ctx, i.db, tx, ev); err != nil {
 		i.logger.Errorw("Cannot sync event object", zap.Error(err))
 		return fmt.Errorf("cannot sync event object: %w", err)
 	}
 
+	stmt := i.db.Rebind(i.db.BuildSelectStmt(i, i) + ` WHERE "recovered_at" IS NULL AND "object_id" = ?`)
+	if err := tx.GetContext(ctx, i, stmt, obj.ID); err != nil && !errors.Is(err, sql.ErrNoRows) {
+		return fmt.Errorf("cannot get active incident: %w", err)
+	}
```
yhabteab: With this new query performed here, the `i.restoreState(ctx, tx)` call in the below else case doesn't make any sense to me now. Why don't you just use the same query performed in `i.restoreState(ctx, tx)` here and call `i.restoreRelatedState` explicitly if that query yields an existing incident? You just need to replace the where clause used there with this one, but everything else should just workk fine.

internal/incident/incident.go
```diff
@@ -156,7 +180,6 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event) error {
 		}
 
 		if !i.RecoveredAt.Time().IsZero() {
-			RemoveCurrent(i.Object)
 			return ErrIncidentRecovered
 		}
```
yhabteab: When my previous comment is accepted, then this if check can also be moved up to near that query as well.

internal/incident/incident.go
```diff
@@ -232,78 +255,85 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event) error {
...
-		channels.LoadFromEscalationRecipients(escalation, ev.Time, i.isRecipientNotifiable)
+		notifications, err = i.generateNotifications(ctx, tx, ev, channels)
+		if err != nil {
+			return fmt.Errorf("cannot generate notifications for reevaluated escalations: %w", err)
+		}
+
+		o, err = i.IncidentObject(ctx)
+		if err != nil {
+			return err
+		}
```
yhabteab, line 313: I would move this down before the final `i.notifyContacts` call. Currently, my IDE flags the `o` usage below as a _Potential nil dereference when passing 'o' to 'notifyContacts'_.

internal/incident/incident.go
```diff
@@ -735,9 +748,34 @@ func (i *Incident) restoreState(ctx context.Context, tx *sqlx.Tx) error {
...
+	if err != nil {
+		i.logger.Errorw("Failed to restore incident rules from the database", zap.Error(err))
+		return err
+	}
+
+	rules := make(map[ruleID]struct{})
+	for _, ruleRow := range ruleRows {
+		rules[ruleRow.RuleID] = struct{}{}
+	}
+
```
yhabteab, line 790: Sorry, but after adding such a query again, I could nothing but give it a try with `utils.ForEachRow`. The only changes needed to add tx support for that function look like this, which I think are way better than this.

<details><summary>Diff</summary>

```diff
diff --git a/internal/utils/utils.go b/internal/utils/utils.go
index 1899f5b..2ec094c 100644
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -13,7 +13,7 @@ import (

 // ExecAndApply applies the provided restoreFunc callback for each successfully retrieved row of the specified type.
 // Returns error on any database failure or fails to acquire the table semaphore.
-func ExecAndApply[Row any](ctx context.Context, db *database.DB, stmt string, args []interface{}, restoreFunc func(*Row)) error {
+func ExecAndApply[Row any](ctx context.Context, db *database.DB, executer database.TxOrDB, stmt string, args []any, restoreFunc func(*Row)) error {
        table := database.TableName(new(Row))
        sem := db.GetSemaphoreForTable(table)
        if err := sem.Acquire(ctx, 1); err != nil {
@@ -22,7 +22,7 @@ func ExecAndApply[Row any](ctx context.Context, db *database.DB, stmt string, ar
        defer sem.Release(1)

        //nolint:sqlclosecheck // False positive, does not detect deferred close: https://github.com/ryanrolds/sqlclosecheck/issues/43
-       rows, err := db.QueryxContext(ctx, db.Rebind(stmt), args...)
+       rows, err := executer.QueryxContext(ctx, db.Rebind(stmt), args...)
        if err != nil {
                return err
        }
@@ -46,7 +46,7 @@ func ExecAndApply[Row any](ctx context.Context, db *database.DB, stmt string, ar
 // ForEachRow applies the provided restoreFunc callback for each successfully retrieved row of the specified type.
 // It will bulk SELECT the data from the database scoped to the specified ids and scans into the provided Row type.
 // Returns error on any database failure or fails to acquire the table semaphore.
-func ForEachRow[Row, Id any](ctx context.Context, db *database.DB, idColumn string, ids []Id, restoreFunc func(*Row)) error {
+func ForEachRow[Row, Id any](ctx context.Context, db *database.DB, executer database.TxOrDB, idColumn string, ids []Id, restoreFunc func(*Row)) error {
        subject := new(Row)
        query := fmt.Sprintf("%s WHERE %q IN (?)", db.BuildSelectStmt(subject, subject), idColumn)
        stmt, args, err := sqlx.In(query, ids)
@@ -54,7 +54,7 @@ func ForEachRow[Row, Id any](ctx context.Context, db *database.DB, idColumn stri
                return errors.Wrapf(err, "cannot build placeholders for %q", query)
        }

-       return ExecAndApply(ctx, db, stmt, args, restoreFunc)
+       return ExecAndApply(ctx, db, executer, stmt, args, restoreFunc)
 }

 // PrefixWithJSONPathRootSelector ensures that the provided JSONPath expression starts with the root
```

</details>

With that diff applied, you can make use of that and get rid of this ugly code. Also, nice catch with the missing explicit `incident_rule` restoration as it never seem to have been restored explicitly that way until now.

schema/mysql/upgrades/incident-next-escalation-check-at.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE incident ADD COLUMN next_escalation_check_at bigint;
+CREATE INDEX idx_incident_next_escalation_check_at ON incident(next_escalation_check_at);
+
+-- As the internal timer has vanished, force a reevaluation for each active incident. Depending on the amount of
+-- incidents, this might result in a temporary flood, limited by the "LIMIT 1000" in ReevaluateEscalations and spread
+-- across multiple 30s intervals.
+UPDATE incident SET next_escalation_check_at = started_at WHERE recovered_at IS NULL;
```
yhabteab: Why? :)

internal/incident/incident.go
oxzi: Good to know. I have briefly updated the comment.

internal/incident/incident.go
oxzi: Good idea, thanks. But it seems like we cannot move the lock that far down, as `i.restoreState` in the `if isNew`'s `else` branch accesses the runtime config.

https://github.com/Icinga/icinga-notifications/blob/d9a6cf6d1f491e0d3e06298352e02935ee29390d/internal/incident/incident.go#L178
https://github.com/Icinga/icinga-notifications/blob/d9a6cf6d1f491e0d3e06298352e02935ee29390d/internal/incident/incident.go#L744
https://github.com/Icinga/icinga-notifications/blob/d9a6cf6d1f491e0d3e06298352e02935ee29390d/internal/incident/incident.go#L751
https://github.com/Icinga/icinga-notifications/blob/d9a6cf6d1f491e0d3e06298352e02935ee29390d/internal/incident/incident.go#L757
https://github.com/Icinga/icinga-notifications/blob/d9a6cf6d1f491e0d3e06298352e02935ee29390d/internal/incident/incident.go#L790

So, I have moved it above the `if isNew` branching.

internal/incident/incident.go
oxzi: Because they do different things. At the point of this query, the `Incident` `i` does not even know its own `ID` or its `StartedAt` time, used in `i.IsNew()` below. It is not even certain if there is such an incident, that's why `sql.ErrNoRows` is ignored.

https://github.com/Icinga/icinga-notifications/blob/d9a6cf6d1f491e0d3e06298352e02935ee29390d/internal/incident/incidents.go#L152-L154

The `restoreState` method uses a `SELECT ... FOR UPDATE` on `i.ID`, locking the incident.

https://github.com/Icinga/icinga-notifications/blob/d9a6cf6d1f491e0d3e06298352e02935ee29390d/internal/incident/incident.go#L745

So, they do different things: Once populating and checking if there is already an incident, and once locking an existing one.

internal/incident/incident.go
oxzi: Good idea, changed it and the method looks much cleaner now.

schema/mysql/upgrades/incident-next-escalation-check-at.sql
oxzi: As discussed offline, I wanted to ensure a migration path, as otherwise all these incidents would never escalate again. But removed again, as more things are going to change.

oxzi (commented):
Thanks for the review, @yhabteab. I have tried to address all of your comments and pushed a rebased version.

---

I have rewritten the `incident.ReevaluateEscalations` function to separate the work into two goroutines: a producer and a consumer.

Unfortunately, I could not use `db.YieldAll` as it requires the `database.Entity` interface - requiring `database.IDer`, which requires `func ID() ID`, where `ID` requires `String()` -, while `Incident` already implements `contracts.Incident` - requiring `func ID() int64`. This contradicts and working around this would be more boilerplate than saving otherwise.

internal/incident/incident.go
oxzi: After rebuilding it as you have suggested with a `WHERE "recovered_at" IS NULL AND "object_id" = ? FOR UPDATE` query, I tried to verify that this is actually safe. The current PR state uses `SELECT FOR UPDATE` on a primary key, while this would use a gap lock referencing `recovered_at` and `object_id`.

MySQL's [documentation on InnoDB Locking](https://dev.mysql.com/doc/refman/8.0/en/innodb-locking.html) states the following for gap locks (emphasis mine):

>  A gap lock is a lock on a gap between index records, or a lock on the gap before the first or after the last index record. For example, SELECT c1 FROM t WHERE c1 BETWEEN 10 and 20 FOR UPDATE; prevents other transactions from inserting a value of 15 into column t.c1, **whether or not there was already any such value in the column, because the gaps between all existing values in the range are locked**.

In my limited understanding and after a bit of greenfield testing, this change might actually lock all `recovered_at IS NULL` rows. Thus, I would not further process this change and keep this as it is.

oxzi: Nevermind. Pushed an updated version according to our discussion last Friday: Using repeatable read isolation level by default and locking `FOR UPDATE` here.

oxzi:
Pushed an updated version, that addresses the `FOR UPDATE` query discussed above and reduces the runtime config locking scope. Please take another look, @yhabteab.

internal/incident/incident.go
```diff
@@ -232,78 +258,87 @@ func (i *Incident) ProcessEvent(ctx context.Context, ev *event.Event) error {
...
 	escalations, err := i.evaluateEscalations(ev.Time)
 	if err != nil {
-		i.logger.Errorw("Reevaluating time-based escalations failed", zap.Error(err))
-		return
+		return fmt.Errorf("cannot reevaluate escalations: %w", err)
 	}
 
-	if len(escalations) == 0 {
+	var notifications []*NotificationEntry
+	var o *object.Object
```
yhabteab: I think, this variable pre-declaration isn't need anymore.

internal/incident/incident.go
```diff
@@ -626,15 +639,26 @@ func (i *Incident) triggerEscalations(ctx context.Context, tx *sqlx.Tx, escalati
...
 		if contact == nil {
+			i.runtimeConfig.RUnlock()
 			i.logger.Debugw("Incident refers unknown contact, might got deleted", zap.Int64("contact_id", notification.ContactID))
 			continue
 		}
 
-		if i.notifyContact(contact, ev, notification.ChannelID) != nil {
+		contactName := contact.String()
+		err := i.notifyContact(obj, contact, ev, notification.ChannelID)
+		i.runtimeConfig.RUnlock()
```
yhabteab, line 661: Instead of passing `notifications.ChannelID` as an argument and accessing the `runtimeConfig` there again, I would just pass the channel instead and release config lock before notifying.

internal/incident/incidents.go
```diff
@@ -246,45 +160,20 @@ func ProcessEvent(
...
+	o := object.New(ev)
+	i := NewIncident(db, o, runtimeConfig, logs.GetChildLogger("incident").With(zap.String("object", o.DisplayName())))
+
+	// Runs in a single transaction. Under concurrent processing (HA setup), this might get aborted by the database.
+	err := retry.WithBackoff(
+		ctx,
+		func(ctx context.Context) error { return i.ProcessEvent(ctx, ev) },
+		retry.Retryable,
+		backoff.DefaultBackoff,
+		db.GetDefaultRetrySettings())
```
yhabteab: I don't think wrapping the entire function call here in a retry is a good idea, since `ProcessEvent` can fail for a number of reasons including notifying users.

yhabteab:
As discussed offline, I've pushed the changes in a [separate branch](https://github.com/Icinga/icinga-notifications/compare/incidents-and-objects-only-in-the-database...yh-ha-ha-ha), please have a look at it and incorporate into this PR as you like.

internal/incident/incident.go
oxzi: You have already addressed this in your wip diff :)

yhabteab:
My bad, I've accidentally pushed into your PR 🙈 .

oxzi (commented):
Worked on your feedback, slurped your commit, added one or two additional locks, and tested it once again.

internal/incident/incidents.go
```diff
@@ -3,239 +3,145 @@ package incident
...
+	ch := make(chan *Incident, 1)
+
+	eg.Go(func() error {
+		defer close(ch)
+
+		stmt := db.Rebind(`
+			SELECT "id", "object_id" FROM "incident"
+			WHERE "recovered_at" IS NULL AND "next_escalation_check_at" IS NOT NULL AND "next_escalation_check_at" <= ?
+			ORDER BY "next_escalation_check_at"
+			LIMIT 1000`)
```
yhabteab:
```suggestion
			ORDER BY "next_escalation_check_at"`)
```

We don't limit the results anymore since they are processed row-wise.

yhabteab (approved):
LFTM! As already discussed offline, some of the functions could be converted to return iterators instead of slice of incidents and objects, but I'll do that in a follow-up PR since that's not required for this PR.

PR #466, opened 2026-07-13
HA: Reset stalled processing queue entries & use updatable last_update timestamp

oxzi:
retention: Reset stalled processing queue entries

    Instead of deleting event queue entries in a stalled processing state,
    reset them to the pending state.

    Initially, there was only one kind of retention, which deleted affected
    rows. With the last refactoring[0], the logic became easier to adjust.
    This enabled this change, introducing the ResetPruner, which performs a
    UPDATE query instead of a DELETE.

    As old event queue entries in the processing state imply a crashed
    daemon or another fatal error, they should be reset and retried.
    Otherwise, we would lose events, which we accept prior to this commit.

    As a drive by change, the queue state variables from the event package
    were exported. This allowed getting rid of magic numbers in the static
    dbPruner configuration.

    [0]: fb63975c2178fa64e7d492869bff6db62f3f8ebc

---

    HA: Dedicated last_update timestamp

    Use a dedicated last_update timestamp instead of the event time for the
    event queue. While this timestamp is used where the prior time - now
    renamed to event_time - is used, it can be updated for each event queue
    object modification.

    This ensures that the retention uses the defined durations instead of
    the duration minus an unknown delta, depending on when the event was
    processed or how long it waited in the queue.

---

Refs #394.


oxzi:
I have pushed a second commit, introducing a second update-able timestamp used for retention - as discussed yesterday. More details are in the commit message and in the updated top post.

yhabteab (approved):
LFTM! Ty.

oxzi:
Just verified the schema upgrade: no diff.

PR #469, opened 2026-07-16
Add `/health` API endpoint

yhabteab:
This endpoint is used by Icinga DB to periodically check a) if the listener is still alive and b) if its database connection is still healthy. Though, like the other endpoints, it also requires an authenticated source.

refs https://github.com/Icinga/icingadb/issues/1141


yhabteab:
Just discussed with Johannes offline and decided to not introduce this endpoint as Icinga DB will have kinda active state once [this issue](https://github.com/Icinga/icingadb/issues/1140) is resolved. So, closing it.

Issue #415, opened 2026-04-13
New API endpoint to fetch notification history


nilmerg:
Sources may want to correlate sent notifications with their own history. To allow this, the HTTP API must provide a new endpoint by which it is possible to fetch notifications sent since a given time:

`GET /notifications?since=<timestamp>`

A source is always expected to pass a timestamp, even if that's just the current time.

The API responds with a single line of JSON for each entry. Each entry is separated by a new line.

Each entry has the property `event_id` which is the unique identifier provided by the source when delivering the event. (See #394)
There may be notifications sent due to an escalation that isn't itself triggered by an event. Such must have the _event id_ of the most recent severity change.

Only _sent_ notifications. i.e. unless the notification has been attempted to deliver, no entry is returned.

```json
{
  "event_id": "a554df8b…",
  "sent_at": 123456,
  "contact_name": "John Doe",
  "contactgroup_name": "The Does",
  "schedule_name": "Everyday",
  "channel_name": "Mail",
  "message": "Something gone wrong."
}
```

julianbrost:
> `GET /notifications?since=<timestamp>`

The word "notifications" is already overloaded in this project, so I'd rather pick a more descriptive path here.

> A source is always expected to pass a timestamp, even if that's just the current time.

What would be the expected behavior in that case? An empty response given that there are probably no events? Or is this implicitly asking for some long running response stream similar to the Icinga 2 event streams?

nilmerg:
> What would be the expected behavior in that case? An empty response given that there are probably no events? Or is this implicitly asking for some long running response stream similar to the Icinga 2 event streams?

I didn't go that much into the detail as I don't have the expertise to do so. I'd welcome your thoughts, or leave it up to the implementor(s) of this and the icingadb counterpart.

nilmerg:
As noted in https://github.com/Icinga/icingadb/issues/1112, the way the event queue ID is constructed is currently incompatible with the intended usage in this API result's body. It is now part of this task to change this:

The event's ID must be the sole attribute to de-/duplicate entries in the job queue. Only if it is missing, the entire event's body is hashed to construct a suitable ID. This requires full responsibility to ensure proper uniqueness from a source if it passes an ID, but also gives it full control over what should deliberately be a duplicate.

It makes absolute sense to require this in form of a UUID. It's also assumed that `event_id` mentioned in the OP returns this as-is.

nilmerg:
[...]

PR #471, opened 2026-07-17
Yield incidents from DB & NDJSON stream HTTP response

yhabteab:
This PR refactors the incident retrieval logic to yield incidents from both the database and an NDJSON stream HTTP response. I initially wanted to implement only the former, but I realized that without support for streaming in our HTTP handlers, the benefits would be limited. Therefore, this PR does both, allowing for more efficient incident retrieval and processing.

As a result, the HTTP handlers HTTP status codes have been updated to reflect the new behavior. The handlers now return a 202 Accepted status code when the request passes its respective validation checks. From there, the incidents are streamed one by one in [NDJSON](http://jsonlines.org/) format, allowing clients to process them as they arrive, but also allowing Icinga Notifications to fetch incidents from the database in a more efficient manner (aka cursor-based retrieval).

I already discussed this with Johannes, and he agreed that the changed behavior is acceptable, especially since an [upcoming new handler](https://github.com/Icinga/icinga-notifications/issues/415) will definitely benefit from/require this streaming behavior.

## Requires

- https://github.com/Icinga/icinga-go-library/pull/222


internal/listener/listener.go
```diff
@@ -500,110 +502,92 @@ func (l *Listener) modifyIncidents(
...
+						IncidentStatus: source.IncidentStatus{
+							Code:   http.StatusInternalServerError,
+							Status: "failed to modify incident, see server logs for details",
+						},
+					}, nil
+				default:
+					return &source.ModifiedIncidentResp{
+						ObjectTags: pair.Object.Tags,
+						IncidentStatus: source.IncidentStatus{
+							Code:   http.StatusOK,
```
oxzi: With the switch to 202 Accepted, having a `"code": 200` JSON response feels a bit strange. How about dropping the `code` field altogether and replace it with an optional `error` field only to be populated in case of an error?

internal/listener/listener.go
```diff
@@ -500,110 +502,92 @@ func (l *Listener) modifyIncidents(
...
-	if err := enc.Encode(results); err != nil {
-		l.logger.Errorw("Failed to serialize modify incidents response", zap.String("source", src.Name), zap.Error(err))
+	pairCh, errCh := incident.YieldForSource(r.Context(), l.db, l.logs, l.runtimeConfig, src.ID)
+	if err := StreamJsonResults(r.Context(), w, pairCh, errCh, opts...); err != nil {
+		l.logger.Debugw("Error streaming modify incidents response", zap.String("source", src.Name), zap.Error(err))
 	}
 }
 
-// getIncidents handles GET requests to the /incidents endpoint.
+// getIncidentsHandler handles GET requests to the /incidents endpoint.
```
oxzi: While you are already doing great changes here, could you please move the `getIncidentsHandler` method before the `modifyIncidentsHandler` method? In my head, it just makes more sense the other way around and I already found myself multiple times scrolling the code in the wrong direction. And than it would mirror the order of the docs.

doc/20-HTTP-API.md
```diff
@@ -122,40 +122,34 @@ Basic Auth with username and password, while a Unix socket identifies the caller
...
-| is_muted    | A boolean indicating whether the incident is muted or not.                 |
-| object_tags | A dictionary containing the object ID tags associated with the incident.   |
-| severity    | The severity level of the incident (e.g., `crit`, `err`, `warning`, etc.). |
+| Attribute   | Description                                                                                                  |
+|-------------|--------------------------------------------------------------------------------------------------------------|
+| is_muted    | A boolean indicating whether the incident is muted or not.                                                   |
+| object_tags | A dictionary containing the object ID tags associated with the incident.                                     |
+| severity    | The severity level of the incident (e.g., `crit`, `err`, `warning`, etc.).                                   |
+| code        | The HTTP status code indicating the result of the request for this incident. Excluded from successful cases. |
+| status      | A string describing the result of the request for this incident. Excluded from successful cases.             |
```
oxzi: The behavior regarding `code` and `status` differs significantly between getting incidents and modifying them. This is not explicitly documented in the modifying section, and would even then be leading to confusions.

How about the following:
- Dropping `code` altogether.
- Renaming `status` to `error`, as it is only present for a success. This would play well together with https://github.com/Icinga/icinga-notifications/pull/471/changes#r3612412403

Afterward, the attributes would be identical for getting and modifying, and we would even have dropped one field.

internal/listener/response_utils.go
```diff
@@ -0,0 +1,122 @@
...
+		case <-ctx.Done():
+			return ctx.Err()
+		case err := <-errCh:
+			if err != nil {
+				with.onError(enc, &wroteHeader, err)
+			}
+			return err
+		case result, ok := <-inCh:
+			if !ok {
+				return <-errCh
```
oxzi: When `yield` emits an error to `errCh`, it always `returns` afterwards, resulting in a deferred close of both `errCh` and `inCh` (`pairCh` named there). Thus, both this case with `!ok` and the previous case `<-errCh` became ready. Since there is no guaranteed order, we might end up here for an error. So, you might also want to do the `with.onError(enc, &wroteHeader, err)` thing here.

internal/listener/listener.go
```diff
@@ -444,48 +442,52 @@ func (l *Listener) IncidentsHandler(w http.ResponseWriter, r *http.Request) {
...
+		l.logger.Warnw("Error processing incident request", zap.String("source", src.Name), zap.Error(err))
+
+		var status source.IncidentStatus
+		if errors.Is(err, ErrFilterEval) {
+			status.Code = http.StatusBadRequest
+			status.Status = err.Error()
+		} else {
+			status = source.IncidentStatus{
+				Code:   http.StatusInternalServerError,
+				Status: "some incidents could could not be modified due to an internal error, see server logs for details",
```
oxzi:
```suggestion
				Status: "some incidents could not be modified due to an internal error, see server logs for details",
```

internal/incident/incidents.go
```diff
@@ -144,14 +60,102 @@ var ErrOpenIncidentWithoutSeverity = errors.New("cannot open or escalate an inci
...
-	o := object.New(ev)
-	i := NewIncident(db, o, runtimeConfig, logs.GetChildLogger("incident").With(zap.String("object", o.DisplayName())))
-	return i.ProcessEvent(ctx, ev)
+	l *logging.Logging,
+	rc *config.RuntimeConfig,
+	srcID int64,
+) (<-chan Pair, <-chan error) {
+	query := `
+		SELECT "incident".* FROM "incident"
+			JOIN "object" ON "incident"."object_id" = "object"."id"
```
oxzi:
```suggestion
		JOIN "object" ON "incident"."object_id" = "object"."id"
```

doc/20-HTTP-API.md
yhabteab: > Afterward, the attributes would be identical for getting and modifying

I've dropped the code and renamed status to error as suggested but the response for retrieving and modifying an incident still aren't identical.

internal/incident/incidents.go
yhabteab: Done.

internal/listener/listener.go
yhabteab: Fixed.

internal/listener/response_utils.go
yhabteab: Fixed.

internal/listener/listener.go
yhabteab: Reordered.

internal/listener/listener.go
yhabteab: Done.

oxzi (commented):
LGTM. After addressing the IGL feedback, merging the IGL PR, and bumping the IGL go.mod reference, this would be mergeable for me.

yhabteab:
> bumping the IGL go.mod reference, this would be mergeable for me.

Done.

oxzi (approved):
Thanks!

Issue #409, opened 2026-03-19
Also send notifications outside or irrespective of any incident


nilmerg:
We have currently only so-called _event rules_ which handle incoming events to map them to such. This needs to be extended in order to be able to handle incoming events, that may not be related to an incident, let alone cause one. To achieve this in Icinga Notifications, the user will be able to set up _escalation rules_ (what _event rules_ were) and _notification rules_. This is about what the latter are and how they' re handled.

## Database schema

_Notification rules_ differ to _escalation rules_ only in that they do not have _escalations_. So, I'm inclined to suggest to add an additional `enum` column to table `rule` and be done with it. Though, recipients remain, and `rule_escalation_recipient` is bound to `rule_escalation`, so either decouple this somewhat or introduce a new table.

## How it should work

As mentioned, _notification rules_ should be able to trigger notifications without any relation to an incident. To differentiate them from the others, the flag `incident` must be set to `false` and thus becomes a boolean flag.

```mermaid

stateDiagram-v2
    direction LR
    [*] --> Event
    Event --> EscalationRule : incident:true
    Event --> NotificationRule : incident:false
    EscalationRule --> Recipient
    NotificationRule --> Recipient
    Recipient --> [*]
```

Events are to be matched using a separate property called `properties`. The value is the same as for the `relations` as described in #406. Though, its purpose is to describe the _event_, not the _object_ and hence, Icinga DB may include `type: Downtime` or `previous_hard_state: ok`. `properties` are also expected to take part in _attribute negotiation_.

In contrast to _escalation rules_, _notification rules_ may match based on `properties` **and** `relations`. Both should be merged before filter evaluation, but preferably only after validation when the _negotiation_ is finished.

_notification rules_ must be set up in order for such events to be processed. If no _notification rule_ matches, an event is dropped.

nilmerg:
Updated the OP to better describe the difference to incident related events.

PR #472, opened 2026-07-22
New api endpoint notification history

### LINKED ISSUE #415: New API endpoint to fetch notification history

Jan-Schuppik:
Sources currently have no way to ask Icinga Notifications which notifications it actually attempted for their objects. This PR adds a GET /notification-history HTTP endpoint that streams every notification attempt (sent or failed) recorded since a given point in time, along with the database schema and internal plumbing needed to record that history.

### What's included

Two new tables, notification_history and skipped_notification_history, record every notification attempt and, for cases where multiple escalation recipients would have notified the same contact through the same channel, the additional origins that were skipped as duplicates. To support this, rule.ContactChannels was reworked from a per-contact/channel boolean into a map of []ChannelOrigin (rule, escalation, contact group, schedule, role), so internal/incident can tell which origin a pending notification is attributed to and which ones were superfluous.

Every event now carries a stable event_id: a UUID either supplied by the source (via the bumped icinga-go-library event.Event.ID field) or, if missing, derived by hashing the event's JSON body in event.Enqueue (internal/event/queue.go). This ID is persisted on incident_history and used to correlate notification_history rows with the event that triggered them. As part of this, the event_queue/job_queue deduplication key was migrated from a raw SHA256 bytea hash to the same UUID.

ContactRole and NotificationState moved out of internal/incident: ContactRole (renamed to recipient.Role) now lives in internal/recipient/contact_role.go, and the local NotificationState type was replaced by icinga-go-library's notifications/source.NotificationState, which required bumping the icinga-go-library dependency in go.mod.

The new endpoint reuses the same source-authentication and NDJSON-streaming machinery as the existing /incidents endpoint; the shared error-handling logic in listener.go was factored out into createStreamErrFunc so both endpoints use it. OrphanRowPruner was also extended to accept multiple ReferencedBy relations instead of a single referencing table, since orphaned object rows now need to check both the incident and notification_history tables before being pruned.

Schema changes (see schema/{mysql,pgsql}/upgrades/notification-history.sql): adds the notification_history and skipped_notification_history tables, adds incident_history.event_id, and changes event_queue.id/job_queue.id from a binary hash column to a UUID column. A running deployment needs the migration applied before upgrading. The MySQL migration also adds a CHECK constraint, which MySQL 5.7 silently ignores instead of enforcing, so this PR also drops mysql:5.7 from CI and raises the documented minimum to MySQL ≥8.0.

### Testing

Added a "Notification History" test case to the existing internal/incident/incidents_test.go, covering both fetching all entries and filtering by since (requires a test database). Extended internal/listener/listener_test.go with a NotificationHistoryHandler subtest covering auth, method, and query-parameter validation, and added internal/listener/response_utils_test.go to cover the shared NDJSON streaming helper. internal/event/queue_test.go was updated to cover UUID-based event deduplication.

To verify manually, run the daemon, trigger some events, then query the endpoint as documented in the new doc/20-HTTP-API.md section:

`curl -Gu 'example:insecureinsecure' 'http://localhost:5680/notification-history?since=<unix-ms>'  --data-urlencode 'filter={"host":"test-host"}'`

Depends on:
- https://github.com/Icinga/icinga-go-library/pull/229
- https://github.com/Icinga/icingadb/pull/1164

resloves #415


doc/20-HTTP-API.md
```diff
@@ -232,6 +232,70 @@ while the incident for the `database` host and `load` service failed to be modif
 {"object_tags":{"environment":"08434a503ec43bb67cd380c5d0b6217a1ebf924b","host":"database","service":"load"},"error":"failed to modify incident, see server logs for details"}
 ```
 
+## Notification History
+
+The `/notification-history` Icinga Notifications HTTP API endpoint allows sources to retrieve a stream of
+notification attempts recorded for its objects, i.e. every time Icinga Notifications tried to notify a contact
+through a channel. Entries can be retrieved by sending a `GET` request to the endpoint. The endpoint requires a
```
yhabteab:
```suggestion
through a channel. Entries can be retrieved by sending a `GET` request to the endpoint. It requires a
```

Reads nicer than the repetitive _the endpoint_.

doc/20-HTTP-API.md
```diff
@@ -232,6 +232,70 @@ while the incident for the `database` host and `load` service failed to be modif
...
+`since` query parameter to specify the earliest point in time to retrieve entries for.
+
+Authentication follows the same transport-specific rules as for event submission: TCP requires HTTP Basic Auth
+with username and password, while a Unix socket identifies the caller by their OS user.
```
yhabteab: The Basic Auth with username and password part is no longer true since we also cert auth as well. I think having an own **Authentication** section explaining all the auth methods would be nice. We could just reference that section instead of having to explain the auth methods for each endpoint possibly going out of sync at some point.

doc/20-HTTP-API.md
```diff
@@ -232,6 +232,70 @@ while the incident for the `database` host and `load` service failed to be modif
...
+
+### Getting Notification History
```
yhabteab: I know, you just wanted to make this consistent with the `/incidents` endpoint but as opposed to that, this handler doesn't and will never support `POST` requests, so I would remove this subheading and join the section with above one instead.

doc/20-HTTP-API.md
```diff
@@ -232,6 +232,70 @@ while the incident for the `database` host and `load` service failed to be modif
...
+
+In order to retrieve notification history entries, send a `GET` request to the `/notification-history` endpoint
+with a `since` query parameter set to a Unix timestamp in milliseconds. Only entries whose `triggered_at` is
+greater than or equal to this value, and that belong to the requesting source, are returned. This endpoint will
+include the attributes listed below in the response for each entry:
+
+| Attribute         | Description                                                                                    |
+|-------------------|--------------------------------------------------------------------------------------------------|
+| event_id          | The hex-encoded ID of the event that triggered the notification.                                 |
```
yhabteab:
```suggestion
| event_id          | Hex-encoded ID of the event that caused the notification to be triggered.                  |
```

yhabteab: I would remove the _The_ prefix from all (except `state` and `message`) the below attr descriptions as well. Instead, you can just use _Name of the contact group ..._.

doc/20-HTTP-API.md
```diff
@@ -232,6 +232,70 @@ while the incident for the `database` host and `load` service failed to be modif
...
+2026-01-01T00:00:00Z (`1767225600000`):
+
+```
+$ curl -v \
+    -u 'example:insecureinsecure' \
+    'http://localhost:5680/notification-history?since=1767225600000'
+```
+```
+...
+{"event_id":"3fa85f645717456bb3fc2c963f66afa6","triggered_at":1767225654000,"contact_name":"Jane Doe","contactgroup_name":"","schedule_name":"On-Call","channel_name":"email","incident_id":42,"message":"Something went somewhere very wrong.","state":"sent"}
```
yhabteab: Joining the request and response looks way more nicer.

```suggestion
$ curl -vu 'example:insecureinsecure' 'http://localhost:5680/notification-history?since=1767225600000'
...
{"event_id":"3fa85f645717456bb3fc2c963f66afa6","triggered_at":1767225654000,"contact_name":"Jane Doe","contactgroup_name":"","schedule_name":"On-Call","channel_name":"email","incident_id":42,"message":"Something went somewhere very wrong.","state":"sent"}
```

doc/20-HTTP-API.md
```diff
@@ -232,6 +232,70 @@ while the incident for the `database` host and `load` service failed to be modif
...
+```
+
+Each line is a single, compact JSON object; the example below shows the same object pretty-printed for readability:
```
yhabteab:
```suggestion
Each line is a single and compact JSON object with the fields described above. The example below shows the same object pretty-printed for readability:
```

Apart from that, does it really make sense to have a pretty-printed of the response here? Since the JSON is the same pretty-printed or not, I wouldn't care that much adding a prettified version of it here. So, I would just remove the below one.

internal/event/event.go
```diff
@@ -34,6 +38,43 @@ type Event struct {
 	evaluatedRelations map[string]jsonpath.NodeList
 }
 
+// CreateEvent creates a new Event from the given source ID and base event.Event.
+// It ensures that the event has a valid ID, generating one if necessary.
+func CreateEvent(sourceId int64, baseEvent baseEv.Event) (Event, error) {
```
yhabteab: The function name should just be `Create` as it's clear what it creates when invoked via `event.Create()`. Also, `baseEv.Event` read from requests in the API listener and from the job queue can be huge, so passing by value that around is costly, so please use pointers for both the parameter type and the return type, so that you only have to copy `baseEv.Event` once when assigning it to the event construction below.

internal/event/event_id_test.go
yhabteab, line 1: Please move this to the existing `event_test.go` file. This doesn't need its own one.

internal/event/event_id_test.go
```diff
@@ -0,0 +1,100 @@
...
+			return Event{
+				Time:     time.Unix(1700000000, 0),
+				SourceId: 1,
+				Event: baseEv.Event{
+					Name:    "dummy: random fortune",
+					Message: "Something went somewhere very wrong.",
+					Tags:    map[string]string{"host": "dummy"},
+				},
+			}
+		}
```
yhabteab: Why not use `CreateEvent` here and everywhere for your tests?

internal/event/event_id_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+		a := newEvent()
+		b := newEvent()
+
+		require.NoError(t, a.EnsureID())
+		require.NoError(t, b.EnsureID())
+
+		assert.NotEqual(t, types.UUID{}, a.ID, "EnsureID should generate a non-zero ID")
```
yhabteab:
```suggestion
		assert.NotZero(t, a.ID, "EnsureID should generate a non-zero ID")
```

internal/event/queue.go
```diff
@@ -42,8 +41,9 @@ const (
 
 // Queue describes an event.Event enqueued for processing in the relational database.
 type Queue struct {
-	// ID is the SHA256 hash based on the event.Event JSON representation.
-	ID types.Binary `db:"id"`
+	// ID is the event.Event ID, which is a UUID provided by the source or generated by the listener.
+	// It is used to deduplicate events and ensure that the same
```
yhabteab: This seems to be missing something.

internal/recipient/contact_role.go
```diff
@@ -0,0 +1,69 @@
+package recipient
+
+import (
+	"database/sql/driver"
+	"fmt"
+)
+
+type ContactRole int
```
yhabteab: This should be just `Role` now so that it makes sense when used like `recipient.Role` from elsewhere.

yhabteab (changes requested):
I haven't got the time to review this thoroughly, but I left some inline comments below. I'll try to get back to this tomorrow and do a more thorough review.

internal/incident/incident.go
```diff
@@ -643,12 +648,31 @@ func (i *Incident) notifyContacts(
...
+			State:             notification.State,
+			TriggeredAt:       notification.SentAt,
+		}
+		stmtNotificationHistory, _ := i.db.BuildUpdateStmt(notificationHistoryUpdateRow)
+		if result, err := i.db.NamedExecContext(ctx, stmtNotificationHistory, notificationHistoryUpdateRow); err != nil {
+			i.logger.Errorw(
+				"Failed to update state and time of notification history", zap.String("contact", contactName),
+				zap.Error(err),
+			)
+		} else if rowsAffected, _ := result.RowsAffected(); rowsAffected == 0 {
```
yhabteab: Please do not ignore errors.

internal/incident/notification_history.go
```diff
@@ -0,0 +1,178 @@
...
+	ChannelID         int64                    `db:"channel_id"`
+	IncidentID        types.Int                `db:"incident_id"`
+	Message           types.String             `db:"message"`
+	State             source.NotificationState `db:"state"`
+}
+
+// TableName implements the contracts.TableNamer interface.
+func (n *NotificationHistory) TableName() string {
+	return "notification_history"
+}
```
yhabteab: This shouldn't be needed. The type is already called exactly like the table name and `database.TableName()` will transform that type name into `notification_history` without this function as well.

internal/incident/notification_history.go
```diff
@@ -0,0 +1,178 @@
...
+    nh.message,
+    nh.state
+    FROM notification_history nh
+    JOIN "contact" c ON nh.contact_id = c.id
+    JOIN "channel" ch ON nh.channel_id = ch.id
+	LEFT JOIN "contactgroup" cg ON nh.contactgroup_id = cg.id
+	LEFT JOIN "schedule" s ON nh.schedule_id = s.id
+    WHERE nh.triggered_at >= ? AND nh.source_id = ?`
+
+	entryCh, entryErrCh := yieldQuery[source.NotificationHistory](ctx, db, query, since, srcID)
```
yhabteab: Please inline this function here. You don't have to out source it.

internal/incident/notification_history.go
```diff
@@ -0,0 +1,178 @@
...
+	srcID int64,
+) (<-chan source.NotificationHistory, <-chan error) {
+	// TODO: which sources should be included?
+	query := `SELECT 
+    nh.event_id,
+    nh.triggered_at,
+    c.full_name AS contact_name,
+    COALESCE(cg.name, '') AS contactgroup_name,
+    COALESCE(s.name, '') AS schedule_name,
+    ch.name AS channel_name,
```
yhabteab: If you change the type of all `*Name` attributes in the `source.NotificationHistory` type to `types.String` instead, you wouldn't have to coalesce anything here.

internal/incident/notification_history.go
```diff
@@ -0,0 +1,178 @@
...
+    WHERE nh.triggered_at >= ? AND nh.source_id = ?`
+
+	entryCh, entryErrCh := yieldQuery[source.NotificationHistory](ctx, db, query, since, srcID)
+
+	return entryCh, entryErrCh
+}
+
+// NotificationSkippedHistoryEntry represents a single skipped_notification_history database entry, recording the path
+// a notification was skipped for a given incident history entry.
+type NotificationSkippedHistoryEntry struct {
```
yhabteab: The table is called `skipped_notification_history` and I would name this type exactly like that as well and you wouldn't need to implement the `TableName` interface method below.

internal/incident/sync.go
```diff
@@ -163,13 +170,57 @@ func (i *Incident) generateNotifications(
...
+						IncidentID:        types.MakeInt(i.Id, types.TransformZeroIntToNull),
+						Message:           types.MakeString(ev.Message, types.TransformEmptyStringToNull),
+						State:             source.NotificationStateFailed,
+						SourceID:          ev.SourceId,
+					}
+
+					if err := notificationHistory.Sync(ctx, i.db, tx); err != nil {
+						i.logger.Errorw("Failed to insert notification history",
+							zap.String("contact", contact.FullName),
+							zap.Bool("incident_muted", i.IsMuted()),
```
yhabteab: `i.IsMuted()` is always false at this point, so you don't need to include it in the logs. Same goes for the else case below.

internal/incident/sync.go
```diff
@@ -163,13 +170,57 @@ func (i *Incident) generateNotifications(
...
+
+					notificationHistory = NotificationHistory{
+						IncidentHistoryID: hr.ID,
+						EventID:           ev.ID,
+						TriggeredAt:       types.UnixMilli(time.Now()),
+						ContactID:         contact.ID,
+						ContactgroupID:    types.MakeInt(origin.ContactGroupID, types.TransformZeroIntToNull),
+						ScheduleID:        types.MakeInt(origin.ScheduleID, types.TransformZeroIntToNull),
+						ChannelID:         chID,
+						IncidentID:        types.MakeInt(i.Id, types.TransformZeroIntToNull),
```
yhabteab: `i.Id` can never be zero here.

```suggestion
						IncidentID:        types.MakeInt(i.Id),
```

internal/incident/sync.go
```diff
@@ -131,7 +132,13 @@ func (i *Incident) AddRuleMatched(ctx context.Context, tx *sqlx.Tx, r *rule.Rule
 //
 // This function will just insert NotificationStateSuppressed incident histories and return an empty slice if
 // the incident is muted, otherwise a slice of pending *NotificationEntry(ies) that can be used to update
-// the corresponding histories after the actual notifications have been sent out.
+// the corresponding histories after the actual notifications have been sent out. The given eventID correlates
+// the generated rows to the event that caused them (or, for time-based escalation re-evaluation without a live
+// event, to the incident's last severity-changing event).
```
yhabteab:
```suggestion
// the corresponding histories after the actual notifications have been sent out. The given event's ID correlates
// with the generated notification histories so that they can be traced back to the event that triggered them.
```

internal/incident/sync.go
```diff
@@ -163,13 +170,57 @@ func (i *Incident) generateNotifications(
...
-					HistoryRowID: hr.ID,
-					ContactID:    contact.ID,
-					State:        NotificationStatePending,
-					ChannelID:    chID,
-				})
+			if suppress {
+				continue
+			}
+
+			var notificationHistory NotificationHistory
```
yhabteab: You don't have to cache the entire type here since you only its ID for the skipped ones.

internal/listener/listener.go
```diff
@@ -424,10 +425,9 @@ func (l *Listener) ProcessEvent(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	ev := event.Event{
-		Time:     time.Now(),
-		SourceId: src.ID,
-		Event:    innerEv,
+	ev, err := event.CreateEvent(src.ID, innerEv)
+	if err != nil {
+		l.abort(w, http.StatusBadRequest, src, "cannot create event.Event: %v", err)
```
yhabteab: As far I can tell, any error returned by `event.CreateEvent` is not something caused by a _bad request_ but simply due to other internal reasons, thus I wouldn't use bad request as http error here.

internal/listener/listener.go
```diff
@@ -655,6 +626,80 @@ func (l *Listener) modifyIncidentsHandler(w http.ResponseWriter, r *http.Request
 	}
 }
 
+// GetNotificationHistory handles GET requests to the /notification-history endpoint.
+func (l *Listener) GetNotificationHistory(w http.ResponseWriter, r *http.Request) {
```
yhabteab:
```suggestion
func (l *Listener) NotificationHistoryHandler(w http.ResponseWriter, r *http.Request) {
```

internal/listener/listener.go
```diff
@@ -655,6 +626,80 @@ func (l *Listener) modifyIncidentsHandler(w http.ResponseWriter, r *http.Request
...
+	}
+
+	onStreamErr := l.createStreamErrFunc(r, w, src, "notification history entries")
+	onStreamResult := func(entry source.NotificationHistory) (any, error) {
+		return entry, nil
+	}
+	opts := []StreamOpt[source.NotificationHistory]{
+		WithOnError[source.NotificationHistory](onStreamErr),
+		WithOnResult(onStreamResult),
+	}
```
yhabteab:
```suggestion
	opts := []StreamOpt[source.NotificationHistory]{
		WithOnError[source.NotificationHistory](l.createStreamErrFunc(r, w, src, "notification histories")),
		WithOnResult(func(entry source.NotificationHistory) (any, error) {
			return entry, nil
		}),
	}
```

internal/listener/listener.go
```diff
@@ -655,6 +626,80 @@ func (l *Listener) modifyIncidentsHandler(w http.ResponseWriter, r *http.Request
...
+}
+
+// createStreamErrFunc returns an [OnErrFunc] that handles errors during streaming of results to the client.
+//
+// It logs the error and sends an appropriate HTTP response to the client. If the error is due to the client
+// disconnecting prematurely, it logs a debug message and does not send a response. If the error is due to a
+// filter evaluation failure, it sends a 400 Bad Request response. For other errors, it sends a 500 Internal
+// Server Error response.
+func (l *Listener) createStreamErrFunc(r *http.Request, w http.ResponseWriter, src *config.Source, namePlural string) OnErrFunc {
+
```
yhabteab: Redundant empty line.

internal/listener/notification_history_test.go
yhabteab, line 1: Please move this to the existing `listener_test.go` file.

internal/rule/rule.go
```diff
@@ -86,31 +86,54 @@ func (r *Rule) Eval(filterable filter.Filterable) (bool, error) {
...
+	ScheduleID       int64 // Non-zero if the contact was resolved from a schedule.
+	Role             recipient.ContactRole
+}
+
+// ContactChannels stores, per contact and channel ID, the origins that selected this channel.
+//
+// When multiple escalation recipients resolve to the same contact and channel, all their origins are
+// recorded: the first origin is the one the notification is attributed to, any further ones denote
+// duplicates that would have notified the same contact via the same channel.
+type ContactChannels map[*recipient.Contact]map[int64][]ChannelOrigin
```
yhabteab: This type has become quite complex now. How about making the channel ID as part of the `ChannelOrigin` type and get rid of the second map here? This type could then look like this instead:
```go
type ContactChannels map[*recipient.Contact][]ChannelOrigin
```

internal/utils/contact_role.go
yhabteab, line 1: Is this a left over? You have already moved this to the recipient package, so this should be gone completely now.

schema/mysql/upgrades/notification-history.sql
```diff
@@ -0,0 +1,45 @@
+ALTER TABLE incident_history ADD COLUMN event_id binary(16); -- used for external references, lower case
```
yhabteab: lower case?

schema/mysql/upgrades/notification-history.sql
```diff
@@ -0,0 +1,45 @@
...
+
+    CONSTRAINT pk_notification_history PRIMARY KEY (id),
+    CONSTRAINT ck_notification_history_state_notnull CHECK (state IS NOT NULL),
+    CONSTRAINT fk_notification_history_incident_history FOREIGN KEY (incident_history_id) REFERENCES incident_history(id),
+    CONSTRAINT fk_notification_history_contact FOREIGN KEY (contact_id) REFERENCES contact(id),
+    CONSTRAINT fk_notification_history_contactgroup FOREIGN KEY (contactgroup_id) REFERENCES contactgroup(id),
+    CONSTRAINT fk_notification_history_schedule FOREIGN KEY (schedule_id) REFERENCES schedule(id),
+    CONSTRAINT fk_notification_history_channel FOREIGN KEY (channel_id) REFERENCES channel(id),
+    CONSTRAINT fk_notification_history_incident FOREIGN KEY (incident_id) REFERENCES incident(id),
+    CONSTRAINT fk_notification_history_source FOREIGN KEY (source_id) REFERENCES source(id)
```
yhabteab: All these foreign keys must be removed as that would not allow to remove contact, contactgroup etc. while it's referenced by this table.

schema/mysql/upgrades/notification-history.sql
```diff
@@ -0,0 +1,45 @@
...
+    rule_escalation_id bigint NOT NULL,
+    contactgroup_id bigint,
+    schedule_id bigint,
+
+    CONSTRAINT pk_skipped_notification_history PRIMARY KEY (id),
+    CONSTRAINT fk_skipped_notification_history_notification FOREIGN KEY (notification_history_id) REFERENCES notification_history(id),
+    CONSTRAINT fk_skipped_notification_history_rule FOREIGN KEY (rule_id) REFERENCES rule(id),
+    CONSTRAINT fk_skipped_notification_history_rule_escalation FOREIGN KEY (rule_escalation_id) REFERENCES rule_escalation(id),
+    CONSTRAINT fk_skipped_notification_history_contactgroup FOREIGN KEY (contactgroup_id) REFERENCES contactgroup(id),
+    CONSTRAINT fk_skipped_notification_history_schedule FOREIGN KEY (schedule_id) REFERENCES schedule(id)
```
yhabteab: Same here.

schema/mysql/upgrades/notification-history.sql
```diff
@@ -0,0 +1,45 @@
...
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE INDEX idx_skipped_notification_history_incident_id ON skipped_notification_history(notification_history_id);
```
yhabteab: What's this index used for? Also the index name doesn't describe what's currently indexing.

yhabteab (changes requested):

schema/mysql/upgrades/notification-history.sql
Jan-Schuppik: To speed up looking up skipped entries for a given notification

doc/40-Authentication.md
yhabteab, line 1: This doesn't need to be in a separate file. With my previous comment, I was meant to introduce a new section within the `20-HTTP-API.md` file itself.

doc/20-HTTP-API.md
yhabteab: This must be also as part of the new authentication section.

doc/20-HTTP-API.md
```diff
@@ -123,8 +104,8 @@ by sending a `POST` request with the desired changes. The endpoint requires a `f
 incidents the source wants to retrieve or modify. Please refer to the [API Filtering](#api-filtering) section for more
 details on how to construct the filter.
 
-Authentication follows the same transport-specific rules as for event submission: TCP requires HTTP
-Basic Auth with username and password, while a Unix socket identifies the caller by their OS user.
+See [Authentication](40-Authentication.md) for the transport-specific rules that apply here, which are the
+same as for event submission.
```
yhabteab: I would just remove the last part of this sentence. Same goes for the notification history endpoint.
```suggestion
See [Authentication](40-Authentication.md) for the transport-specific rules that apply here.
```

doc/20-HTTP-API.md
```diff
@@ -232,6 +213,50 @@ while the incident for the `database` host and `load` service failed to be modif
 {"object_tags":{"environment":"08434a503ec43bb67cd380c5d0b6217a1ebf924b","host":"database","service":"load"},"error":"failed to modify incident, see server logs for details"}
 ```
 
+## Notification History
+
+The `/notification-history` Icinga Notifications HTTP API endpoint allows sources to retrieve a stream of
+notification attempts recorded for its objects, i.e. every time Icinga Notifications tried to notify a contact
+through a channel. Entries can be retrieved by sending a `GET` request to the endpoint. It requires a
+`since` query parameter to specify the earliest point in time to retrieve entries for.
```
yhabteab: Either try to merge the third paragraph and this one together or remove this entirely as this states in both of them:

```suggestion
through a channel.
```

doc/20-HTTP-API.md
```diff
@@ -232,6 +213,50 @@ while the incident for the `database` host and `load` service failed to be modif
...
+
+```
+$ curl -v -u 'example:insecureinsecure' 'http://localhost:5680/notification-history?since=1767225600000'
+...
+{"event_id":"3fa85f645717456bb3fc2c963f66afa6","triggered_at":1767225654000,"contact_name":"Jane Doe","contactgroup_name":"","schedule_name":"On-Call","channel_name":"email","incident_id":42,"message":"Something went somewhere very wrong.","state":"sent"}
+```
+
+Each line is a single and compact JSON object with the fields described above.
+
+Like the `/incidents` endpoint, Icinga Notifications streams the response as a series of JSON objects, one per
```
yhabteab: Same here. You could merge these for example like this:

```suggestion
Like the `/incidents` endpoint, Icinga Notifications streams the response as a series of JSON objects with the fields described above, one per
```

schema/mysql/schema.sql
```diff
@@ -449,8 +450,39 @@ CREATE TABLE incident_history (
...
+    schedule_id bigint,
+
+    CONSTRAINT pk_skipped_notification_history PRIMARY KEY (id),
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE INDEX idx_skipped_notification_history_notification_history_id ON skipped_notification_history(notification_history_id);
+
 CREATE TABLE event_queue (
-    id binary(32) NOT NULL, -- SHA256 of JSON representation.
+    id binary(16) NOT NULL, -- used for external references, lower case
```
yhabteab: What does lower case actually mean in this case? Also, this isn't just for external references but has a real  and significant meaning to Icinga Notifications too. So, just keep its previous comment please.

internal/incident/incidents.go
```diff
@@ -33,11 +33,25 @@ func ReevaluateEscalations(
 	pairCh, errCh := yield(ctx, db, logger, runtimeConfig, false, query, types.UnixMilli(time.Now()))
 	var errs []error
 	for pair := range pairCh {
-		err := pair.Incident.RetriggerEscalations(ctx, pair.Object, &event.Event{
-			Time:  time.Now(),
-			Event: baseEv.Event{Incident: types.MakeBool(true)},
-		})
```
yhabteab, line 39: Please leave this function unchanged. Instead, fetch the corresponding event ID within `i.RetriggerEscalations` just before calling the `i.generateNotifications` function. In case there are no newly triggered escalations or `i.RetriggerEscalations` fails due to other reasons prematurely, you're wasting a database query on a kind large table for nothing here.

internal/incident/incident.go
```diff
@@ -716,8 +745,9 @@ func (i *Incident) getRecipientsChannel(t time.Time) rule.ContactChannels {
 
 				for _, contact := range contacts {
 					if contactChs[contact] == nil {
-						contactChs[contact] = make(map[int64]bool)
-						contactChs[contact][contact.DefaultChannelID] = true
+						contactChs[contact] = []rule.ChannelOrigin{{
+							Role: state.Role,
+						}}
```
yhabteab, line 756: The comment above the outer loop says this:
> 	// Check whether all the incident recipients do have an appropriate contact channel configured.
	// When a recipient has subscribed/managed this incident via the UI, fallback to the default contact channel.

And you're adding only the state here which is quite useless without the actual channel ID.

internal/incident/notification_history.go
```diff
@@ -0,0 +1,148 @@
...
+    WHERE nh.triggered_at >= ? AND nh.source_id = ?`
+
+	valueCh := make(chan source.NotificationHistory)
+	errCh := make(chan error, 1) // buffered to avoid goroutine leak if the receiver is not ready to receive errors.
+
+	go func() {
+		defer close(valueCh)
+		defer close(errCh)
+
+		tableName := new(source.NotificationHistory).TableName()
```
yhabteab: Why make it that complicated? You already know which types you're processing here, so why use this indirection instead of just using, for example, `notification history`?

internal/listener/listener.go
```diff
@@ -464,7 +461,7 @@ func (l *Listener) ProcessEvent(w http.ResponseWriter, r *http.Request) {
 	ctx, cancel := context.WithTimeout(r.Context(), 10*time.Second)
 	defer cancel()
 
-	err := event.Enqueue(ctx, l.db, &ev, object.ID(ev.SourceId, ev.Tags))
+	err := event.Enqueue(ctx, l.db, ev, object.ID(ev.SourceId, ev.Tags))
```
yhabteab: Please revert all these changes.

internal/event/queue.go
```diff
@@ -94,15 +94,14 @@ func (q *Queue) toEvent() (*Event, error) {
...
 
-	err := json.NewEncoder(io.MultiWriter(&jsonBuff, idHash)).Encode(ev)
+	err := json.NewEncoder(io.MultiWriter(&jsonBuff)).Encode(ev)
 	if err != nil {
 		return fmt.Errorf("cannot JSON encode event.Event: %w", err)
 	}
 
 	q := &Queue{
-		ID:         idHash.Sum(nil),
+		ID:         ev.ID,
```
yhabteab, line 183: Now, you're using `ev.ID` unconditionally regardless whether it was or not. Please let the `io.MultiWriter` write to `idHash` as well and use it conditionally, i.e., if `ev.ID` is not set.

internal/incident/incident.go
```diff
@@ -643,12 +648,26 @@ func (i *Incident) notifyContacts(
...
 		}
 
 		notification.SentAt = types.UnixMilli(time.Now())
+
+		notificationHistoryUpdateArgs := map[string]any{
+			"ihi":   notification.HistoryRowID,
+			"state": notification.State,
+			"tt":    notification.SentAt,
+		}
+		stmtNotificationHistory := i.db.Rebind(`UPDATE "notification_history" SET "state" = :state, "time" = :tt WHERE "id" = :ihi`)
```
yhabteab: It's a named query, you don't need to explicitly rebind it. Also, there's no such as `time` column in the `notification_history` table and the where clause is wrong since it's referencing the `notification_history.ID` column while using `incident_history.ID` as an argument.

```suggestion
		stmtNotificationHistory := `UPDATE "notification_history" SET "state" = :state, "triggered_at" = :tt WHERE "incident_history_id" = :ihi`
```

internal/incident/incidents_test.go
```diff
@@ -483,3 +484,8 @@ func withMsg(msg string) eventOption              { return func(ev *event.Event)
 func withSeverity(sev baseEv.Severity) eventOption {
 	return func(ev *event.Event) { ev.Severity = sev }
 }
+func withID(id uuid.UUID) eventOption {
+	return func(ev *event.Event) {
+		ev.ID = types.MakeUUID(id)
+	}
+}
```
yhabteab: Do you really need this for the tests? If yes, then you're only using it once but there also a bunch of other places in this file that create and process new incidents/events that still don't make use of this. Alternatively, you can always set the ID in `makeEvent` without having to provide this option if the ID is really needed for something.

internal/incident/notification_history.go
```diff
@@ -0,0 +1,127 @@
...
+// Each yielded entry has its SkippedHistory field populated with the skipped_notification_history rows recorded
+// for it, if any.
+func YieldNotificationHistoryForSource(
+	ctx context.Context,
+	db *database.DB,
+	since string,
+	srcID int64,
+) (<-chan source.NotificationHistory, <-chan error) {
+	// TODO: which sources should be included?
+	query := `SELECT
```
yhabteab: Where are the changes we made to this query and function? Also, the TODO should be gone when you implement what Johannes requested in https://github.com/Icinga/icinga-notifications/issues/415#issuecomment-5339789460. It's still missing! When doing that, please drop the `ForSource` part from the function name.

yhabteab (changes requested):

internal/rule/rule.go
```diff
@@ -86,31 +86,56 @@ func (r *Rule) Eval(filterable filter.Filterable) (bool, error) {
...
+			RuleEscalationID: er.EscalationID,
+			ContactGroupID:   er.GroupID.Int64,
+			ScheduleID:       er.ScheduleID.Int64,
+			Role:             recipient.RoleRecipient,
+		}
 		for _, c := range er.Recipient.GetContactsAt(t) {
 			if ch[c] == nil {
-				ch[c] = make(map[int64]bool)
+				ch[c] = []ChannelOrigin{}
 			}
```
yhabteab: You don't need to do this anymore since it's perfectly fine to just append to a nil slice as well.

internal/testutils/testutils.go
```diff
@@ -66,3 +69,11 @@ func MakeRandomString(t *testing.T) string {
 
 	return fmt.Sprintf("%x", buf)
 }
+
+// MakeRandomUUID returns a random UUID string.
+func MakeRandomUUID(t *testing.T) types.UUID {
```
yhabteab: As discussed this in the other PR, you don't need a random UUID as you can just use `uuid.New()` isntead that does't require such helper function. So, please remove this.

internal/incident/sync.go
```diff
@@ -131,45 +132,99 @@ func (i *Incident) AddRuleMatched(ctx context.Context, tx *sqlx.Tx, r *rule.Rule
...
+					RuleEscalationID: origin.RuleEscalationID,
+					ContactgroupID:   types.MakeInt(origin.ContactGroupID, types.TransformZeroIntToNull),
+					ScheduleID:       types.MakeInt(origin.ScheduleID, types.TransformZeroIntToNull),
+				}
+				if err := skippedHistoryEntry.Sync(ctx, i.db, tx); err != nil {
+					i.logger.Errorw("Failed to insert notification skipped history",
+						zap.String("contact", contact.FullName),
+						zap.Error(err))
+					return nil, err
+				}
```
yhabteab, line 201: After taking a closer look at this code snippet, it doesn't make sense to me at all. You're inserting something into the `notification_history` table with state `failed` here, even though the notification at this point has not actually failed yet, not even attempted to be sent. So, I think, this function just needs to return all the necessary information to the caller, and actually perform the insertion into the `notification_history` table with the actual state of the notification at the point of sending it (right where you're performing an update to the `notification_history` table).

internal/rule/rule.go
```diff
@@ -86,31 +86,56 @@ func (r *Rule) Eval(filterable filter.Filterable) (bool, error) {
...
 			if er.ChannelID.Valid {
-				ch[c][er.ChannelID.Int64] = true
+				ch[c] = append(ch[c], origin)
 			} else {
-				ch[c][c.DefaultChannelID] = true
+				ch[c] = append(ch[c], origin)
```
yhabteab: That's not quite correct. As you can see in the diff context, previously the if and else case were different but now has become the same without any difference. You've to override `origin.ChannelID` if `!er.ChannelID.Valid` and append to `ch[c]` only once please.

schema/mysql/upgrades/notification-history.sql
```diff
@@ -0,0 +1,33 @@
+ALTER TABLE incident_history ADD COLUMN event_id char(36) AFTER incident_id; -- used for external references
+ALTER TABLE event_queue MODIFY COLUMN id char(36) NOT NULL;
```
yhabteab: Please change the type. We don't use `char` for UUIDs anymore. Also, the `incident_history` can get quite big, so it might a good idea to add an idex on the `event_id` column as well since you'll be using that column in your where clause.

schema/mysql/upgrades/notification-history.sql
```diff
@@ -0,0 +1,33 @@
...
+    event_id char(36) NOT NULL,
+    triggered_at bigint NOT NULL,
+    contact_id bigint NOT NULL,
+    contactgroup_id bigint,
+    schedule_id bigint,
+    channel_id bigint NOT NULL,
+    incident_id bigint NOT NULL,
+    message text NOT NULL,
+    state enum('sent', 'failed'),
+    source_id bigint NOT NULL,
```
yhabteab: If this column is still required (please sort this with Johannes out), move it to above `incident_history_id` and move the `triggered_at` column to after the `state` column just like how you did it in the postgres schema (keep them consistent please).

yhabteab (changes requested):
Another round of review comments :). Sorry for the chunked review comments but the PR is quite big, so I need to do something else to get my head right again :). Apart from the tests, I've now reviewed all at least once.

doc/20-HTTP-API.md
```diff
@@ -101,19 +82,35 @@ curl -v -u 'icingadb:insecureinsecure' -H 'X-Icinga-Reject-If-Relations-Incomple
 EOF
 ```
 
-To submit over a Unix socket instead, pass `--unix-socket /run/icinga/icinga-notifications.sock` to curl.
-No credentials are needed; the daemon identifies the caller by their OS user automatically.
+## Authentication
```
yhabteab: This must be above the `Process Event` section as well.

doc/20-HTTP-API.md
```diff
@@ -101,19 +82,35 @@ curl -v -u 'icingadb:insecureinsecure' -H 'X-Icinga-Reject-If-Relations-Incomple
...
+
+    A process connecting via the Unix socket can only submit events for sources whose configured
+    listener_username matches the process's OS username. Restrict socket access using `socket_mode` and
+    `socket_group` to limit which OS users can connect.
 
 !!! info
 
-    The daemon identifies the source by the certificate's Subject. If no matching source is found, the request is rejected.
+    Before Icinga Notifications version 0.2.0, the username was a fixed string based on the source ID, such as `source-${id}`.
+    When upgrading a setup from an earlier version, these usernames are still valid, but can be changed in Icinga Notifications Web.
```
yhabteab: It's way too old now and can be removed entirely :).

doc/20-HTTP-API.md
yhabteab: Still not changed.

internal/event/queue.go
```diff
@@ -94,22 +96,26 @@ func (q *Queue) toEvent() (*Event, error) {
...
 		LastUpdate: types.UnixMilli(time.Now()),
 		Json:       jsonBuff.String(),
 		EventTime:  types.UnixMilli(ev.Time),
 		ObjectId:   objectID,
 		UserAgent:  "icinga-notifications/" + internal.Version.Version,
 		State:      QueueStatePending,
 	}
+
```
yhabteab: Please revert.

internal/incident/incident.go
```diff
@@ -664,6 +685,22 @@ func (i *Incident) notifyContacts(
...
 
+		if err := notification.HistoryEntry.Sync(ctx, i.db); err != nil {
+			i.logger.Errorw(
+				"Failed to update incident history with contact notified", zap.String("contact", contactName),
+				zap.Error(err),
+			)
+		}
+		for _, skippedEntry := range notification.SkippedHistoryEntries {
+			skippedEntry.NotificationID = notification.HistoryEntry.ID
+			_, err := database.InsertObtainID(
```
yhabteab: You don't need to use `database.InsertObtainID` here. You can just use plain inserts.

internal/incident/notification_history.go
```diff
@@ -0,0 +1,130 @@
...
+
+	n.ID = historyId
+
+	return nil
+}
+
+// YieldNotificationHistory returns a channel of [Pair] for all active incidents of the given source (see yield docstring).
+//
+// Each yielded entry has its SkippedHistory field populated with the skipped_notification_history rows recorded
+// for it, if any.
```
yhabteab: The doc string is outdated.

internal/incident/sync.go
```diff
@@ -131,45 +135,92 @@ func (i *Incident) AddRuleMatched(ctx context.Context, tx *sqlx.Tx, r *rule.Rule
...
-					zap.Error(err))
-				return nil, err
-			}
+	var notifications []*NotificationEntry
+	for contact, channelOrigins := range contactChannels {
+		slices.SortFunc(channelOrigins, func(a, b rule.ChannelOrigin) int {
+			return cmp.Compare(a.ChannelID, b.ChannelID)
+		})
+
+		var lastOrigin rule.ChannelOrigin
```
yhabteab: You only need the the channel ID, so please cache only that instead.

internal/incident/sync.go
```diff
@@ -131,45 +135,92 @@ func (i *Incident) AddRuleMatched(ctx context.Context, tx *sqlx.Tx, r *rule.Rule
...
+				notificationHistory := NotificationHistory{
+					ObjectID:          object.ID(ev.Tags),
+					IncidentHistoryID: hr.ID,
+					EventID:           ev.ID,
+					ContactID:         contact.ID,
+					ContactgroupID:    types.MakeInt(origin.ContactGroupID, types.TransformZeroIntToNull),
+					ScheduleID:        types.MakeInt(origin.ScheduleID, types.TransformZeroIntToNull),
+					ChannelID:         origin.ChannelID,
+					IncidentID:        types.MakeInt(i.Id),
+					Message:           types.MakeString(ev.Message, types.TransformEmptyStringToNull),
```
yhabteab: The `message` column of that table is not nullable and transforming empty string to null here will trigger that constraint.

internal/incident/sync.go
```diff
@@ -131,45 +135,92 @@ func (i *Incident) AddRuleMatched(ctx context.Context, tx *sqlx.Tx, r *rule.Rule
...
+					i.logger.Errorw("Failed to insert incident notification history",
+						zap.String("contact", contact.FullName),
+						zap.Bool("incident_muted", i.IsMuted()),
+						zap.Error(err))
+					return nil, err
+				}
+
+				if suppress {
+					// channels[chID] gets not overwritten with an actual notification history ID and keeps -1 as value,
+					// it means the notification was suppressed and no history entry was created.
```
yhabteab: Outdated.

internal/listener/listener.go
```diff
@@ -427,10 +429,18 @@ func (l *Listener) ProcessEvent(w http.ResponseWriter, r *http.Request) {
...
 	ev.CompleteURL(daemon.Config().IcingaWeb2UrlParsed)
 
+	if ev.ID == (types.UUID{}) {
+		//if err := ev.CreateID(); err != nil {
+		//	l.abort(w, http.StatusInternalServerError, src, "cannot create event ID: %v", err)
+		//	return
+		//}
+		l.logger.Warnw("Received event without an ID; creating a new ID for it", zap.String("source", src.Name), zap.Object("event", &ev))
+	}
+
```
yhabteab, line 432: Why? Please revert all these changes.

internal/rule/rule.go
```diff
@@ -86,32 +86,53 @@ func (r *Rule) Eval(filterable filter.Filterable) (bool, error) {
...
 		for _, c := range er.Recipient.GetContactsAt(t) {
-			if ch[c] == nil {
-				ch[c] = make(map[int64]bool)
-			}
-			if er.ChannelID.Valid {
-				ch[c][er.ChannelID.Int64] = true
-			} else {
-				ch[c][c.DefaultChannelID] = true
+			if !er.ChannelID.Valid {
+				origin.ChannelID = c.DefaultChannelID
```
yhabteab: Either keep the previous if ... else statements as-is and set `origin.ChannelID` to `er.ChannelID.Int64` if it's valie and this one in the else case. Otherwise, if `er.Recipient.GetContactsAt(t)` returns multiple contacts and the first contacts sets `origin.ChannelID` to its default one, and when all the other contacts have valid channel ID (`er.ChannelID.Valid == true`), then origin.ChannelID is left unchanged, i.e., set to the previous contact default channel ID. Or move the `origin` variable decration to within the loop.

internal/testutils/testutils.go
```diff
@@ -17,15 +17,24 @@ import (
...
-// It requires the environment variable "ICINGA_NOTIFICATIONS_DATABASE_TYPE" to be set, otherwise it will
-// skip the test. The function validates the provided configuration struct and populates it with values
-// from the environment variables prefixed with "ICINGA_NOTIFICATIONS_".
-func LoadTestConfig[T config.Validator](t *testing.T, configFile T) {
+// Call this before daemon.InjectTestConfig, not from inside its callback: InjectTestConfig only ever invokes
+// its callback once per test binary (it guards a package-wide singleton), so if the first test to reach it is
+// the one that skips, every later test calling InjectTestConfig silently gets an empty, never-loaded config
+// instead of being skipped itself. Checking the environment independently in each test, before touching the
+// singleton, ensures every test in the package skips consistently when the database isn't configured.
+func RequireTestDBConfig(t *testing.T) {
```
yhabteab: Thanks for fixing this. I actually thought, I've fixed this problem with `LoadTestConfig` and the sync.Once but that's obviously wrong. However, I would name the function something like `SkipTestIfDBConfigIsMissing`.

schema/pgsql/upgrades/notification-history.sql
```diff
@@ -0,0 +1,37 @@
+CREATE TYPE notification_history_state_type AS ENUM ( 'sent', 'failed' );
+
+ALTER TABLE incident_history ADD COLUMN event_id uuid AFTER incident_id; -- used for external references
```
yhabteab: There is no `AFTER` keyword in PostgreSQL.

yhabteab (changes requested):

schema/pgsql/upgrades/notification-history.sql
```diff
@@ -0,0 +1,37 @@
+CREATE TYPE notification_history_state_type AS ENUM ( 'sent', 'failed' );
+
+ALTER TABLE incident_history ADD COLUMN event_id uuid; -- used for external references
+ALTER TABLE event_queue ALTER COLUMN id SET DATA TYPE uuid USING (id::uuid);
```
yhabteab: This fails with the following error:

```plpgsql
CREATE TYPE
ALTER TABLE
ERROR:  cannot cast type bytea to uuid
LINE 1: ...t_queue ALTER COLUMN id SET DATA TYPE uuid USING (id::uuid);
```

yhabteab (changes requested):
Apart from the below comment, I think I'm done reviewing the non-test code now.

yhabteab (commented):
Ah, sorry. I also forget to mention that you've to adjust the retention pruner for the `object` table to consider the `notification_history` reference.

doc/03-Configuration.md
```diff
@@ -179,6 +181,12 @@ Currently, the following components are available:
     closed, not from the time they were opened. If you want to clean up some incidents that are still open too, you need
     to close them first, and then wait for the retention period to pass.
 
+!!! info
+
+    The `notification_history` retention period is computed from the time each notification was triggered, and is
+    independent of its related incident's retention. This means notification history entries can be cleaned up before
+    or after their incident, depending on how the two retention periods are configured.
+
```
yhabteab: I think, this info is redundant since `notification_history` is a component on its own, and thus independent by defintion, it should be already clear that it doesn't depend on anything. Thus, I would just remove this entirely.

internal/incident/incidents_test.go
```diff
@@ -58,6 +63,53 @@ func TestIncidents(t *testing.T) {
...
+
+	contact := recipient.Contact{
+		ExternalUUID:     types.MakeUUID(uuid.New()),
+		FullName:         "testuser",
+		DefaultChannelID: ch.ID,
+	}
+	contact.ChangedAt = types.UnixMilli(time.Now())
+	contact.Deleted = types.MakeBool(false)
+
+	notificationHistoryRule := &rule.Rule{Name: "escalation test rule", SourceType: source.Type}
```
yhabteab: Since this event rule is also used for something else, the `notificationHistoryRule` variable name isn't correct anymore.

internal/incident/incidents_test.go
```diff
@@ -323,10 +356,58 @@ func TestIncidents(t *testing.T) {
...
+
+		t.Run("Filters By Since", func(t *testing.T) {
+			entryCh, errCh := YieldNotificationHistory(t.Context(), db, "9999999999999")
+			var entries []NotificationHistoryPair
+			for entry := range entryCh {
+				entries = append(entries, entry)
+			}
+			require.NoError(t, <-errCh)
+			assert.Empty(t, entries, "an implausibly high since value must exclude all entries")
+		})
```
yhabteab: There's no such as _filter by source_ anymore, so you can remove this entirely.

internal/incident/incidents_test.go
```diff
@@ -323,10 +356,58 @@ func TestIncidents(t *testing.T) {
...
+	t.Run("Notification History", func(t *testing.T) {
+		t.Parallel()
+
+		i := makeIncident(db, logs, runtimeConfig, t, makeEvent(t, source.ID,
+			withIncident(), withSeverity(baseEv.SeverityDebug), withTags(map[string]string{"notification_history_test": "true"})))
+		assert.NotZero(t, i.ID())
+		assert.Zero(t, i.RecoveredAt)
+		assert.Equal(t, baseEv.SeverityDebug, i.Severity)
+
+		t.Run("Is Not Scoped To A Single Source", func(t *testing.T) {
```
yhabteab: When my previous comment is resolved, then you don't have to `t.Run` here. Instead perform the assertion directly in the parent `t.Run("Notification History" ....` test case.

internal/incident/incidents_test.go
```diff
@@ -323,10 +356,58 @@ func TestIncidents(t *testing.T) {
...
+			entryCh, errCh := YieldNotificationHistory(t.Context(), db, "1")
+
+			var entries []NotificationHistoryPair
+			for entry := range entryCh {
+				entries = append(entries, entry)
+			}
```
yhabteab: Why not perform the assertions directly within this loop instead :)?

internal/channel/plugin_test.go
```diff
@@ -53,6 +59,11 @@ func TestPlugin(t *testing.T) {
 
 		ch := makeTestChannel(t, db, logs.GetChildLogger("channel").Desugar(), "sleepy1", "sleep", `{"duration": "2s"}`)
 
+		filterMapMu.Lock()
+		filterMap["channel"] = append(filterMap["channel"], fmt.Sprintf("id = %d", ch.ID))
+		filterMap["channel_state"] = append(filterMap["channel_state"], fmt.Sprintf("channel_id = %d", ch.ID))
+		filterMapMu.Unlock()
```
yhabteab: Please outsource this to a helper closur/function variable.

internal/retention/prune.go
```diff
@@ -258,36 +264,36 @@ func (orp *OrphanRowPruner) Exec(ctx context.Context, db *database.DB, l *loggin
 
 // assembleSelect constructs a select statement to retrieve primary keys of the OrphanRowPruner.
 func (orp *OrphanRowPruner) assembleSelect(limit uint64) string {
+	var joinList, condList []string
+	for i, rel := range orp.ReferencedBy {
+		alias := fmt.Sprintf("ref%d", i)
```
yhabteab:
```suggestion
	for _, rel := range orp.ReferencedBy {
		alias := fmt.Sprintf("%s_ref", rel.Table)
```

Is better than the index based alias name.

internal/retention/prune.go
```diff
@@ -258,36 +264,36 @@ func (orp *OrphanRowPruner) Exec(ctx context.Context, db *database.DB, l *loggin
...
+		joinList = append(joinList, fmt.Sprintf(
+			"LEFT JOIN %[1]s %[2]s ON %[2]s.%[3]s = main.%[4]s", rel.Table, alias, rel.FK, orp.PKorFK,
+		))
+		condList = append(condList, fmt.Sprintf("%s.%s IS NULL", alias, rel.FK))
+	}
+
+	joins := strings.Join(joinList, " ")
+	condition := strings.Join(condList, " AND ")
```
yhabteab: Why use these extra variables? Can't you just do `strings.Join(...)` directly where these variables are currently used?

schema/mysql/upgrades/notification-history.sql
```diff
@@ -0,0 +1,41 @@
+ALTER TABLE incident_history ADD COLUMN event_id binary(16) AFTER incident_id; -- used for external references
+
+-- event_queue is a short-lived processing queue, so any in-flight rows are safe to discard here;
+-- sources resubmit on failure. This also sidesteps the fact that PostgreSQL has no direct cast from
+-- bytea to uuid, and that the old id (a full SHA256 hash) doesn't fit into the new 16-byte uuid anyway.
+TRUNCATE TABLE event_queue;
+ALTER TABLE event_queue MODIFY COLUMN id binary(16) NOT NULL;
```
yhabteab: `event_queue` doesn't exist anymore.

schema/mysql/upgrades/notification-history.sql
```diff
@@ -0,0 +1,41 @@
...
+    message text NOT NULL,
+    state enum('sent', 'failed'),
+    triggered_at bigint NOT NULL,
+
+    CONSTRAINT pk_notification_history PRIMARY KEY (id),
+    CONSTRAINT ck_notification_history_state_notnull CHECK (state IS NOT NULL),
+    CONSTRAINT fk_notification_history_object_id FOREIGN KEY (object_id) REFERENCES object(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+CREATE INDEX idx_notification_history_event_id ON notification_history(event_id);
```
yhabteab: I don't think this index is needed at all. AFAIK, it is not used by anything. Instead, you should create an index on the `triggered_at` column since that's used by the yield history function in the Go code. Don't forget to adjust it everywhere.

schema/mysql/upgrades/notification-history.sql
```diff
@@ -0,0 +1,41 @@
+ALTER TABLE incident_history ADD COLUMN event_id binary(16) AFTER incident_id; -- used for external references
```
yhabteab, line 1: You should also create a compound index on `event_id, incident_id`. That'll will be used by the query used in the `RetriggerEscallations` method.

schema/pgsql/upgrades/notification-history.sql
```diff
@@ -0,0 +1,42 @@
...
+    contactgroup_id bigint,
+    schedule_id bigint,
+    channel_id bigint NOT NULL,
+    incident_id bigint NOT NULL,
+    message text NOT NULL,
+    state notification_history_state_type NOT NULL,
+    triggered_at bigint NOT NULL,
+
+    CONSTRAINT pk_notification_history PRIMARY KEY (id),
+    CONSTRAINT fk_notification_history_object_id FOREIGN KEY (object_id) REFERENCES object(id)
```
yhabteab, line 34: Additionally to what I've commented in the MySQL scripts, you should also create an index on the `object_id` column here to speed the `object` retentions since it will lookup entries matching `object_id` in this table. However, as opposed to MySQL, PostgreSQL doesn't automatically index foreign keys, so you must do it yourself.

yhabteab (changes requested):

schema/pgsql/upgrades/notification-history.sql
```diff
@@ -0,0 +1,53 @@
...
+);
+
+CREATE INDEX idx_notification_history_triggered_at ON notification_history(triggered_at);
+CREATE INDEX idx_notification_history_event_id_incident_id ON notification_history(event_id, incident_id);
```
yhabteab: Please remove this idx. Same goes for the schema.sql file. You've already removed this in the MySQL scripts.

yhabteab:
I'm fine with this PR now. If you want to take a look at the schema changes and do some tests @nilmerg, pls do so now, otherwise, I'm going to approve and merge this today 🎊.

schema/pgsql/schema.sql
```diff
@@ -513,9 +515,49 @@ CREATE TABLE incident_history (
 CREATE INDEX idx_incident_history_incident_id ON incident_history(incident_id);
 CREATE INDEX idx_incident_history_time_type ON incident_history(time, type);
 COMMENT ON INDEX idx_incident_history_time_type IS 'Incident History ordered by time/type';
+-- This will be used by the query in the [Incident.RetriggerEscalations] method in icinga-notifications.
+CREATE INDEX idx_incident_history_event_id_incident_id ON incident_history(event_id, incident_id);
+
+CREATE TABLE notification_history (
+    id bigserial,
+    object_id bytea NOT NULL,
+    incident_history_id bigint NOT NULL,
```
nilmerg: since we have no references inside `incident_history` yet, I don't see a reason for this ref either.

doc/20-HTTP-API.md
```diff
@@ -238,6 +229,42 @@ while the incident for the `database` host and `load` service failed to be modif
...
+
+| Attribute         | Description                                                                                |
+|-------------------|--------------------------------------------------------------------------------------------|
+| event_id          | Hex-encoded ID of the event that caused the notification to be triggered.                  |
+| triggered_at      | Unix timestamp in milliseconds at which the notification attempt was made.                 |
+| contact_name      | Full name of the contact the notification was sent to.                                     |
+| contactgroup_name | Name of the contact group the contact was resolved from, if any.                           |
+| schedule_name     | Name of the on-call schedule the contact was resolved from, if any.                        |
+| channel_name      | Name of the channel used to deliver the notification.                                      |
+| incident_id       | ID of the incident the notification belongs to, if any.                                    |
```
nilmerg: please drop this from the response, incident ID is nothing the source in question should worry about

schema/pgsql/schema.sql
```diff
@@ -513,9 +515,49 @@ CREATE TABLE incident_history (
...
+    event_id uuid NOT NULL,
+    contact_id bigint NOT NULL,
+    contactgroup_id bigint,
+    schedule_id bigint,
+    channel_id bigint NOT NULL,
+    incident_id bigint NOT NULL,
```
nilmerg: If we'd leave this nullable we can avoid a migration once #409 is in the works

doc/20-HTTP-API.md
```diff
@@ -238,6 +229,42 @@ while the incident for the `database` host and `load` service failed to be modif
...
+| message           | The message that was sent to the contact, if any.                                          |
+| state             | The state of the notification attempt, either `sent` or `failed`.                          |
+| error             | An optional attribute that may be present if an error occurred while retrieving the entry. |
+
+The following example shows how to retrieve all notification history entries recorded since
+2026-01-01T00:00:00Z (`1767225600000`):
+
+```
+$ curl -v -u 'example:insecureinsecure' 'http://localhost:5680/notification-history?since=1767225600000'
```
nilmerg: This won't work without a filter

doc/20-HTTP-API.md
```diff
@@ -238,6 +229,42 @@ while the incident for the `database` host and `load` service failed to be modif
...
+| Attribute         | Description                                                                                |
+|-------------------|--------------------------------------------------------------------------------------------|
+| event_id          | Hex-encoded ID of the event that caused the notification to be triggered.                  |
+| triggered_at      | Unix timestamp in milliseconds at which the notification attempt was made.                 |
+| contact_name      | Full name of the contact the notification was sent to.                                     |
+| contactgroup_name | Name of the contact group the contact was resolved from, if any.                           |
+| schedule_name     | Name of the on-call schedule the contact was resolved from, if any.                        |
+| channel_name      | Name of the channel used to deliver the notification.                                      |
+| incident_id       | ID of the incident the notification belongs to, if any.                                    |
+| message           | The message that was sent to the contact, if any.                                          |
```
nilmerg: please rename to `event_message`, otherwise I feel this leads to the expectation that one assumes that's what was sent as part of the notification.

yhabteab:
Looks fine to me now. Please have a quick look at it @oxzi, otherwise, I'll approve and merge this next.

doc/20-HTTP-API.md
```diff
@@ -238,6 +229,41 @@ while the incident for the `database` host and `load` service failed to be modif
 {"object_tags":{"environment":"08434a503ec43bb67cd380c5d0b6217a1ebf924b","host":"database","service":"load"},"error":"failed to modify incident, see server logs for details"}
 ```
 
+## Notification History
+
+See [Authentication](#authentication) for the transport-specific rules that apply here.
+
+In order to retrieve notification history entries, send a `GET` request to the `/notification-history` endpoint
+with a `since` query parameter set to a Unix timestamp in milliseconds. Only entries whose `triggered_at` is
+greater than or equal to this value, and that belong to the requesting source, are returned. This endpoint will
```
oxzi: > and that belong to the requesting source

This is not checked, potential privacy or confidentiality leak. At the moment, `YieldNotificationHistory` is directly called, not filtered by any source.

oxzi: Duplicate of https://github.com/Icinga/icinga-notifications/pull/472#discussion_r3934410512 - haven't seen those before. Furthermore, as discussed with @nilmerg, this is the expected behavior and the docs are incorrect. Thus, this just needs a doc change.

oxzi: https://github.com/Icinga/icinga-notifications/compare/b1b0f2bea12985abb5f902aeae5374dc886075e1..5fc130ff56bdd8414b0e2da1cb17817b6db8c79c

yhabteab (approved):
🤯

PR #473, opened 2026-07-27
channel: use real `jsonrpc` & cleanup package

### LINKED ISSUE #178: Channel: Logging output
Currently the channels are not able to log anything, which makes debugging unnecessarily difficult[^0].

Due to the strict request-response API logic, a new "Log" method should be introduced, which just got consumed from the RPC, 
wrapping the message and logging it to icinga-notification's logger. On the channel's side, a specially configured zap logger is needed.

[^0]: I just stumbled about Icinga/icinga-notifications-web#154 again, even as I filled this issue back then..

yhabteab:
This PR is a follow-up to the previous PR of mine, which introduced the `jsonrpc` package in Icinga Go Library (see the requires section below). This PR now makes use of the `jsonrpc` package and cleans up the `channel` package a bit. It was a bit messy before because we didn't change anything in that package since its initial implementation. There is no behavior change in this PR apart from the following two changes:

1. Since the new `jsonrpc` package allows for a bidirectional communication, Icinga Notifications acts now as a JSON-RPC client and server at the same time. Meaning, for the former case, it will continue to send JSON-RPC requests to the channel plugins, and for the latter case, it will now also be able to receive JSON-RPC requests from the channel plugins. Currently, it only implements a `Log` method, which allows channel plugins to send log messages with fine-grained log level and  ontexts to Icinga Notifications. This is useful for debugging purposes, as it allows channel plugins to log messages directly to the Icinga Notifications log as opposed to the currently used `stderr` mechanism to report errors.
2. The second behavior change is that Icinga Notifications will not restart the channel plugins with each config change anymore. Instead, it will just call the `SetConfig` method on the channel plugins, which allows them to replace their configs without restarting. The only time a channel plugin will be restarted is when it crashes or when the `channel.type` changes in the config.

## Requires

- https://github.com/Icinga/icinga-go-library/pull/226

resolves #178


yhabteab:
Bumped the go.mod file to require IGL main branch.

internal/channel/channel.go
```diff
@@ -81,65 +87,72 @@ func (c *Channel) initPlugin(cType string, config string) *Plugin {
...
+			c.Logger.Warnw("Channel plugin stopped unexpectedly, restarting it", zap.Int("pid", current.cmd.Process.Pid))
+			stopReset()
+
+		case newConf := <-c.restartCh:
+			// The RPC connection is safe for concurrent use, so we can try to reload the config without going
+			// through the whole plugin restart process. Though, if the plugin type has changed entirely, then
+			// we definitely need to stop the current plugin and start a new one.
+			if current != nil && currentConfig.config == newConf.ctype {
+				c.Logger.Infow("Reloading channel plugin config on the fly", zap.Int("pid", current.cmd.Process.Pid))
+				if err := current.SetConfig(c.pluginCtx, currentConfig.config); err != nil {
```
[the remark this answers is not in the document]
yhabteab: Good catch.

cmd/channels/email/main.go
```diff
@@ -122,20 +126,32 @@ func (ch *Email) GetInfo() *plugin.Info {
...
+	err := plugin.PopulateDefaults(&tmpEm)
 	if err != nil {
 		return err
 	}
 
-	err = json.Unmarshal(jsonStr, ch)
+	err = json.Unmarshal(jsonStr, &tmpEm)
 	if err != nil {
 		return fmt.Errorf("failed to load config: %s %w", jsonStr, err)
 	}
```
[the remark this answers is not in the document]
yhabteab: I didn't even touch that code :).

yhabteab:
Sorry, force pushed again! While writing unittests for the channel plugin starting,reloading,communcating logic, I noticed that the stderr handler of the started plugin was allowed to read unlimited message in 512KiB chunks until the ticker fires. Withing that 10s interval it might read a huge amount of data from stderr causing `bytes.Buffer.ReadFrom` ~to panic~ (`int(^uint(0) >> 1)` is the max limit 🙈) and thus crash Icinga Notifications. I've changed that handler to never exceed that `512KiB` limit, and will always flush everything it has before it reads something again.

doc/10-Channels.md
```diff
@@ -15,100 +15,156 @@ which are then started, configured, and finally used to send notification events
...
-|----------|------------------|----------------------------------------------|
-| `method` | String           | **Required.** Request method to call.        |
-| `params` | JSON object      | **Optional.** Params for the request method. |
-| `id`     | Unsigned integer | **Required.** Unique identifier.             |
+| Field   | Type             | Description                                  |
+|---------|------------------|----------------------------------------------|
+| jsonrpc | String           | **Required.** Must be `"2.0"`.               |
+| method  | String           | **Required.** Request method to call.        |
+| params  | JSON object      | **Optional.** Params for the request method. |
+| id      | Unsigned integer | **Required.** Unique identifier.             |
```
oxzi: Marking the `id` field as required contradicts the statement above (and the spec):

> For each request that includes an `id`, the channel must return a response, even if the request was invalid or could not be processed. Requests without an `id` are JSON-RPC notifications and must not be responded to.

doc/10-Channels.md
```diff
@@ -15,100 +15,156 @@ which are then started, configured, and finally used to send notification events
...
+The `channel::Log` method allows channel plugins to log messages to Icinga Notifications' logging system.
+The request's `params` field must be a JSON object with the following fields:
+
+| Field   | Type   | Description                                                                  |
+|---------|--------|------------------------------------------------------------------------------|
+| level   | String | **Required.** Log level, one of `"debug"`, `"info"`, `"warn"`, or `"error"`. |
+| message | String | **Required.** Log message to be logged.                                      |
+| fields  | Array  | **Optional.** Additional fields to be logged as key-value pairs.             |
+
+The `fields` array must contain an even number of elements, where each pair of elements represents a key-value pair.
```
oxzi: The table states that `fields` is optional, now here it _MUST contain_ something.


```suggestion
If `fields` are defined, they must be an array of an even number of elements, with each pair representing a key-value pair.
```

doc/10-Channels.md
oxzi, line 1: This might be just me, but when reading such an RFC-adjacent document, I am expecting RFC 2119 words to be spelled in caps.

cmd/channels/rocketchat/main.go
```diff
@@ -5,21 +5,37 @@ import (
...
+		mu     sync.Mutex // Protects access to the above fields.
+	}
+
+	// roundTripper is a custom http.RoundTripper that adds the Rocket.Chat authentication headers to each request.
+	roundTripper struct {
+		http.RoundTripper
+
+		token  string
+		userID string
+	}
```
oxzi: I cannot really follow why you have added this `roundTripper` instead of setting the request headers in `SendNotification` as before. Prior, `http.DefaultTransport` was already used implicitly, already allowing to reuse connections. So, just adding `drainAndClose` should be enough.

doc/10-Channels.md
```diff
@@ -182,13 +240,14 @@ They are also used for channel configuration in Icinga Notifications Web.
 
 The `SetConfig` method configures the channel.
 
-The Icinga Notifications daemon will call this method at least once on each channel
-before sending the first notifications to initialize the channel plugin.
+The Icinga Notifications will call this method at least once on each channel before sending the first notifications to
```
```suggestion
Icinga Notifications will call this method at least once on each channel before sending the first notification to
```

doc/10-Channels.md
```diff
@@ -182,13 +240,14 @@ They are also used for channel configuration in Icinga Notifications Web.
...
+initialize the channel plugin. However, it may also be called multiple times during the lifetime of a channel plugin,
+for example, if the user changes the configuration in Icinga Notifications Web. Therefore, the channel plugin must be
+able to handle multiple calls to `SetConfig` and update its configuration accordingly at any given time.
 
 The passed JSON object in the request's `param` field reflects the objects from `GetInfo`'s `config_attrs`.
```
```suggestion
The passed JSON object in the request's `params` field reflects the objects from `GetInfo`'s `config_attrs`.
```

doc/10-Channels.md
```diff
@@ -285,7 +351,7 @@ or if the channel is missing required configuration values.
 
 A channel offers its configuration options through its response to the [`GetInfo` method call](#getinfo).
 Each configuration option is described by a
-[`ConfigOption` entry](https://pkg.go.dev/github.com/icinga/icinga-notifications/pkg/plugin#ConfigOption)
+[`ConfigOption` entry](https://pkg.go.dev/github.com/icinga/icinga-go-library/notifications/plugin#ConfigOption)
 in the `config_attrs` array.
 Those information will be queried once by Icinga Notifications upon startup and stored in the database.
```
```suggestion
This information will be queried once by Icinga Notifications upon startup and stored in the database.
```

doc/10-Channels.md
```diff
@@ -325,23 +396,29 @@ Therefore, we encourage you to develop your own channels that cover your specifi
...
-libraries already used internally can be reused.
-In particular, the [`Plugin`](https://pkg.go.dev/github.com/icinga/icinga-notifications/pkg/plugin#Plugin)
-interface must be implemented, requesting methods for all the RPC methods described above.
+libraries already used internally can be reused. In particular, the
+[Icinga Notifications Plugin](https://pkg.go.dev/github.com/icinga/icinga-go-library/notifications/plugin) and
+[Icinga Notifications JSON-RPC](https://pkg.go.dev/github.com/icinga/icinga-go-library/notifications/jsonrpc) packages
+provide a framework for writing channel plugins in Go. The `jsonrpc` package implements the JSON-RPC 2.0 protocol,
+while the `plugin` package provides types and functions for implementing channel plugins. Also, the `plugin` package
+provides a [`Run`](https://pkg.go.dev/github.com/icinga/icinga-go-library/notifications/plugin#Run) function that takes
+care of setting up a JSON-RPC server and dispatching of requests to the appropriate methods of the channel plugin.
```
```suggestion
care of setting up a JSON-RPC server and dispatching requests to the appropriate methods of the channel plugin.
```

doc/10-Channels.md
```diff
@@ -15,100 +15,156 @@ which are then started, configured, and finally used to send notification events
...
-    As this is still an early preview version, things might change.
-    There may still be severe bugs and incompatible changes may happen without any notice.
-
 Channel plugins are independent processes that run continuously, started and supervised by Icinga Notifications.
-They receive JSON-formatted requests on `stdin` and reply with JSON-formatted responses on `stdout`.
-For logging or debugging purposes, channels can write to `stderr`,
-which is being forwarded to the Icinga Notifications log.
+They receive [JSON-RPC 2.0](https://www.jsonrpc.org/specification) requests on `stdin` and reply with JSON-RPC 2.0
+responses on `stdout`. For each request that includes an `id`, the channel must return a response, even if the request
+was invalid or could not be processed. Requests without an `id` are JSON-RPC notifications and must not be responded to.
```
```suggestion
was invalid or could not be processed. Requests without an `id` are JSON-RPC Notifications and must not be responded to.
```

Capitalized as below in L140.

internal/channel/plugin.go
```diff
@@ -1,168 +1,214 @@
...
+					flush()
+
+				default:
+					// Plugin might literally be spamming stderr, so flush before we read more data to avoid excessive memory usage.
+					flush()
+					if err := r.SetReadDeadline(time.Now().Add(10 * time.Second)); err != nil {
+						logger.Warnw("Failed to set read deadline for channel plugin stderr pipe", zap.Error(err))
+						return
+					}
+					// Read up to 512KB from the plugin's stderr pipe. This is a safeguard to prevent excessive memory
```
oxzi: Most pedantic nit pick: 512KiB, not 512KB. Maybe also drop the comment in L80, as this is something one should understand, and it is now stated twice.

doc/10-Channels.md
oxzi, line 391: Damn, this hint is straight wrong.

https://github.com/Icinga/icinga-notifications/blob/28379e15a66fbc0bac3bc43570ac00076cab8cfc/internal/channel/plugin.go#L267-L268

Since you are already touching this code and the current regex is too tight IMO, would you mind adding the underscore there as well?

oxzi: There is no `event.type` anymore.

internal/channel/plugin.go
```diff
@@ -1,168 +1,214 @@
...
+		}
+		if params.Level == zapcore.InvalidLevel {
+			h.logger.Warnw("Received invalid log level from plugin", zap.String("level", params.Level.String()))
+			return
+		}
+		if params.Message == "" {
+			h.logger.Warnw("Received empty log message from plugin")
+			return
+		}
+		h.logger.Logw(params.Level, params.Message, params.Fields...)
```
oxzi, line 225: The `params.Level` filter above only exclude `zapcore.InvalidLevel`, but not `DPanicLevel`, `PanicLevel`, or `FatalLevel`. So, a channel plugin might crash Icinga Notifications.

The safe option would be to check `params.Level` against an allow list, containing those levels you have already documented:
https://github.com/Icinga/icinga-notifications/blob/28379e15a66fbc0bac3bc43570ac00076cab8cfc/doc/10-Channels.md?plain=1#L134

internal/channel/plugin.go
```diff
@@ -1,168 +1,214 @@
...
+				if buf.Len() > 0 {
+					logger.Errorw("Channel plugin stderr", zap.Int("pid", cmd.Process.Pid), zap.String("stderr", buf.String()))
+					buf.Reset()
+				}
+			}
+
+			for {
+				select {
+				case <-ctx.Done():
+					return
```
oxzi, line 95: This context is `c.pluginCtx` and survives plugin restarts. So, each plugin restart will leak a goroutine and the `os.Pipe()` fds.

internal/channel/channel.go
```diff
@@ -81,65 +87,72 @@ func (c *Channel) initPlugin(cType string, config string) *Plugin {
...
+			if current != nil && currentConfig.ctype == newConf.ctype {
+				c.Logger.Infow("Reloading channel plugin config on the fly", zap.Int("pid", current.cmd.Process.Pid))
+				if err := current.SetConfig(c.pluginCtx, newConf.config); err != nil {
+					c.Logger.Warnw("Failed to reload plugin config, restarting the plugin", zap.Error(err))
+					// If we got a JSON-RPC error, then it's because the plugin rejected the new config for some
+					// reason, so we can just keep the plugin running with the old config. Otherwise, the plugin is
+					// probably already gone, so call stopReset() to clean up and prepare for a new plugin instance.
+					if _, ok := errors.AsType[*jsonrpc.Error](err); !ok {
+						stopReset()
+					}
```
oxzi, line 141: Keeping a plugin running with a diverting configuration from the database configuration is unexpected, IMO. Especially as the prior log message says that the plugin is being restarted, which did not happen. Why not restart it regardless?

internal/channel/plugin.go
```diff
@@ -1,168 +1,214 @@
...
+		}
+		if params.Message == "" {
+			h.logger.Warnw("Received empty log message from plugin")
+			return
+		}
+		h.logger.Logw(params.Level, params.Message, params.Fields...)
+
+	default:
+		if err := jsonrpc.ReplyMethodNotFound(ctx, conn, req.ID); err != nil {
+			h.logger.Error("Failed to send method not found reply", zap.Error(err))
```
oxzi:
```suggestion
			h.logger.Errorw("Failed to send method not found reply", zap.Error(err))
```

internal/channel/plugin.go
```diff
@@ -1,168 +1,214 @@
...
-func forwardLogs(errPipe io.Reader, logger *zap.SugaredLogger) {
-	scanner := bufio.NewScanner(errPipe)
-	for scanner.Scan() {
-		logger.Info(scanner.Text())
-	}
+// Handle handles the JSON-RPC requests made by any channel plugins to Icinga Notifications.
+func (h rpcHandler) Handle(ctx context.Context, conn *jsonrpc.Conn, req *jsonrpc.Request) {
+	switch req.Method {
+	case notifications.MethodLog:
+		if req.Params == nil {
```
oxzi, line 207: If a channel sends a `channel::Log` JSON-RPC request with an `id` (not a JSON-RPC Notification), this code would accept it correctly, but would not reply. This would break protocol. Consider adding a check if we are dealing with a JSON-RPC Notification and, if not, reply with an error response.

internal/channel/plugin.go
```diff
@@ -1,168 +1,214 @@
...
+				select {
+				case <-ctx.Done():
+					return
+
+				case <-ticker.C:
+					flush()
+
+				default:
+					// Plugin might literally be spamming stderr, so flush before we read more data to avoid excessive memory usage.
+					flush()
```
oxzi: Using both a ticker and an always matching `default` to call `flush()` is redundant. The `ticker` can be dropped.

doc/10-Channels.md
oxzi: Nevermind tho :D

cmd/channels/rocketchat/main.go
yhabteab: Idk. Will change it back.

internal/channel/plugin.go
yhabteab: Yep. That would be fatal.

internal/channel/plugin.go
yhabteab: I have wrongly assumed that when closing the write end of the pipe would cause the reader to return os.EOF but it doesn't seem, so I'll change it.

internal/channel/channel.go
yhabteab: > Especially as the prior log message says that the plugin is being restarted, which did not happen.

Yeah, I noticed that as well and fixed it in the other PR with https://github.com/Icinga/icinga-notifications/pull/483/changes/9387a1b04eab108d253e5dcd66df8cd58f0764bf.

> Why not restart it regardless?

Because restarting the plugin will bring nothing if the new config is invalid and we would be just wasting resource and time restarting and killing the plugin endlessly. This on the other hand, let the plugin keep its existing config regardless whether it's valid or invalid, and I don't have to mindlessly spawn a new process with each loop tick.

yhabteab:
Rebased and addressed all the requested changes except https://github.com/Icinga/icinga-notifications/pull/473#discussion_r3774062620, so please have a look at it again.

doc/10-Channels.md
oxzi: The second part of the change is still open: singular notification instead of notifications.

yhabteab: Right. Will push that together with https://github.com/Icinga/icinga-notifications/pull/483/changes/9387a1b04eab108d253e5dcd66df8cd58f0764bf after you finish reviewing it.

internal/channel/channel.go
oxzi: The argument makes sense. I could have thought of this by myself. At this point, I haven't started reviewing the other PR. So, this is fine then.

internal/channel/plugin.go
```diff
@@ -1,168 +1,230 @@
...
+		}
+		if _, allowed := allowedLogLevels[params.Level]; !allowed {
+			h.logger.Warnw("Received invalid log level from plugin", zap.String("level", params.Level.String()))
+			return
+		}
+		if params.Message == "" {
+			h.logger.Warnw("Received empty log message from plugin")
+			return
+		}
+		h.logger.Logw(params.Level, params.Message, params.Fields...)
```
oxzi, line 225: It should be checked if `len(params.Field) % 2 == 0`. Otherwise, the last field would be ignored (just verified this).

internal/channel/plugin.go
```diff
@@ -1,168 +1,230 @@
...
-	}
-	cmd.Stderr = logWrite
-	childIOPipes = append(childIOPipes, logWrite)
-	parentIOPipes = append(parentIOPipes, logRead)
+	ctx, cancel := context.WithCancel(ctx)
+	// Set up a pipe for the plugin's stderr to capture and log any crashes or error messages.
+	// This is important because if the plugin crashes, we want to know why, and stderr is where
+	// such messages are typically sent.
+	cmd.Stderr = func() *os.File {
+		r, w, err := os.Pipe()
```
oxzi, line 71: It seems that `w` - or `cmd.Stderr` - is never closed as well. So, this would also be a leaking fd.

internal/channel/plugin.go
```diff
@@ -221,10 +280,10 @@ func UpsertPlugins(ctx context.Context, channelPluginDir string, logger *logging
...
-// pluginTypeValidateRegex defines Regexp with only allowed characters of the channel plugin type
-var pluginTypeValidateRegex = regexp.MustCompile("^[a-zA-Z0-9]+$")
+// pluginTypeValidateRegex defines Regexp with only allowed characters of the channel plugin type.
+var pluginTypeValidateRegex = regexp.MustCompile("^[a-zA-Z0-9_-]+$")
 
-// ValidateType returns an error if non-allowed chars are detected, nil otherwise
+// ValidateType returns an error if non-allowed chars are detected, nil otherwise.
 func ValidateType(t string) error {
 	if !pluginTypeValidateRegex.MatchString(t) {
 		return fmt.Errorf("type contains invalid chars, may only contain a-zA-Z0-9, %q given", t)
```
oxzi: Thanks for fixing the regex. However, now this error is outdated, sorry.

internal/channel/plugin.go
yhabteab: What should I do if that's the case then? I'm not going to validate that field whether it contains a matching key-value pairs because that's documented and if a plugin doesn't do that, well then just let zap handle it.

internal/channel/plugin.go
yhabteab: Errhnmmmm. Yeah.

yhabteab:
Addressed all the newly requested changes except the first one and also the previously requested change from https://github.com/Icinga/icinga-notifications/pull/473#discussion_r3775216967 together with the changes from the other PR.

internal/channel/plugin.go
oxzi: Well, you could send back an error, as the client would not satisfy the contract defined in the docs. But we can also keep it that way. I'm fine with it.

yhabteab: > Well, you could send back an error, as the client would not satisfy the contract defined in the docs.

No, I can't. This request is a JSON-RPC Notification and I'm not allowed to reply to it.

oxzi (approved):
LGTM, thanks!

PR #475, opened 2026-07-29
Fix resource leak and logging madness

yhabteab:
This PR fixes a resource leak caused by an endless defer statement within an endless loop in the `event.ProcessQueue` function and removes a bunch of log statements for the same event processing logic at multiple places without containing any different contex.

### DIFF
--- a/internal/event/queue.go
+++ b/internal/event/queue.go
@@ -201,7 +201,6 @@ func ProcessQueue(
 			zap.Stringer("object_id", q.ObjectId))
 
 		callbackCtx, cancel := context.WithTimeout(ctx, time.Minute)
-		defer cancel()
 
 		if ev, err := q.toEvent(); err != nil {
 			logger.Errorw("Invalid event queue event cannot get decoded", zap.Stringer("id", q.ID), zap.Error(err))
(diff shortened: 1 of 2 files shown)


internal/event/queue.go
```diff
@@ -196,23 +196,29 @@ func ProcessQueue(
...
-		} else if err = callback(callbackCtx, ev); err != nil {
-			logger.Errorw("Event queue event cannot be processed", zap.Stringer("id", q.ID), zap.Error(err))
-			q.State = QueueStateError
-		} else {
-			logger.Debugw("Processed event from event queue", zap.Stringer("id", q.ID))
-			q.State = QueueStateDone
-		}
+		func() {
+			callbackCtx, cancel := context.WithTimeout(ctx, time.Minute)
+			defer cancel()
```
oxzi: I don't think there is a need to call an anonymous function with a deferred cancel here. Just use the old code, but manually call the cancel after the if/else if/else block.

yhabteab: Done.

oxzi (approved):
LGTM, thanks!

PR #482, opened 2026-08-06
schema: Only require an `external_uuid` for non-deleted rows

nilmerg:
This enables Icinga Notifications Web to release external UUIDs upon deletion of a contact, contact group and channel.

Previously, re-creating such and re-using a UUID was not possible. Since it was never documented to behave this way, I consider this a bug and the fix is to change the schema and introduce checks to avoid having non-deleted rows without an external UUID.

### DIFF
--- a/schema/mysql/upgrades/nullable_external_uuid.sql
+++ b/schema/mysql/upgrades/nullable_external_uuid.sql
@@ -1,0 +1,14 @@
+ALTER TABLE channel
+  MODIFY COLUMN external_uuid char(36),
+  ADD CONSTRAINT ck_channel_non_deleted_needs_external_uuid CHECK (deleted = 'y' OR external_uuid IS NOT NULL);
+UPDATE channel SET external_uuid = NULL WHERE deleted = 'y';
+
+ALTER TABLE contact
+  MODIFY COLUMN external_uuid char(36),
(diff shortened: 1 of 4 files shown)


schema/mysql/upgrades/nullable_external_uuid.sql
```diff
@@ -0,0 +1,14 @@
+ALTER TABLE channel
+  ALTER COLUMN external_uuid DROP NOT NULL,
+  ADD CONSTRAINT ck_channel_non_deleted_needs_external_uuid CHECK (deleted = 'y' OR external_uuid IS NOT NULL);
+UPDATE channel SET external_uuid = NULL where deleted = 'y';
```
yhabteab: There's no such as `DROP NOT NULL` in MySQL/MariaDB. That's postgres only thing. Also, please use consistent SQL keyword casing, i.e., s/where/WHERE/.

yhabteab (changes requested):

schema/mysql/upgrades/nullable_external_uuid.sql
nilmerg: Done.

yhabteab (approved):

PR #483, opened 2026-08-07
Stateful channel plugins

### LINKED ISSUE #102: Implement extra channel types for testing/debugging purposes
For testing the daemon implementation, it would be helpful to have a channel that
1. always returns an error from its send function.
2. sleeps for a while in its send function.

### LINKED ISSUE #128: Stateful Channel Plugins
Probably not the best title, so let me try to explain what I mean by that.

It would be nice if notifications for the same incident could be grouped together, so that for example, with e-mail, follow-up notifications show up as a response to the previous e-mail and for a chat, messages for the same incident are within one thread.

In order to achieve this, it's probably necessary to provide some mechanism to notification plugins to keep some state, like for example the Message-ID of the last and/or first e-mail for an incident or some thread reference for other messengers.

The goal of this issue is to evaluate what exactly would be needed to achieve this and suggest how this could be implemented.

yhabteab:
Adds a stateful plugin mechanism so channel plugins can persist small pieces of key/value state (e.g. last message ID, dedupkey) across restarts, keyed per channel. For this, I've added a new `channel_state` table plus `UpsertState` and `DeleteState` RPC methods, letting plugins read/write their own state indirectly. The plugins use Icinga Notifications as a proxy to perform the required DB operations. Every time Icinga Notifications sends a `SendNotification` RPC to a channel plugin, it also sends the current state for that channel, so the plugin doesn't have to make a separate RPC call to read it.

Apart from that, I've also added new internal/channel/plugin_test.go covering config reload, crash recovery, state persistence, and stderr handling using a real plugin binary without using a test double.

I've also added the corresponding docs for this new behaviour, so the PR cann't get any bigger now.

## Requires

- #473
- https://github.com/Icinga/icinga-go-library/pull/231

resolves #128
resolves #102


yhabteab:
Yuhuu, tests are working fine now 😌.

cmd/channels/email/main.go
```diff
@@ -156,29 +184,58 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
 }
 
 func (ch *Email) SendNotification(req *plugin.NotificationRequest) error {
-	var to []mail.Address
+	var to *mail.Address
 	for _, address := range req.Contact.Addresses {
 		if address.Type == "email" {
-			to = append(to, mail.Address{Name: req.Contact.FullName, Address: address.Address})
+			to = &mail.Address{Name: req.Contact.FullName, Address: address.Address}
+			break
```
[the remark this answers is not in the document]
yhabteab: That's an intended change as we don't currently allow a contact to have more than one address per channel plugin type.

cmd/channels/sleep/main.go
```diff
@@ -0,0 +1,217 @@
...
+	s.duration = duration
+	s.PersistState = tmp.PersistState
+	s.SpamStderr = tmp.SpamStderr
+	s.mu.Unlock()
+
+	return nil
+}
+
+func (s *Sleep) SendNotification(nr *plugin.NotificationRequest) error {
+	slog.Info("Sleep plugin received notification request", "contact", nr.Contact.FullName, "incident_id", nr.Incident.Id, "recovered", nr.Incident.IsRecovered)
```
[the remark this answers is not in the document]
yhabteab: You're right. Will fix it.

cmd/channels/email/main.go
```diff
@@ -156,29 +184,58 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
 }
 
 func (ch *Email) SendNotification(req *plugin.NotificationRequest) error {
```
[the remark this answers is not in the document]
yhabteab: How am I supposed to mock an SMT server for such kind of tests? Nonsense.

internal/channel/channel.go
```diff
@@ -130,16 +134,25 @@ func (c *Channel) pluginControlLoop(currentConfig newConfig) {
...
 					}
 				}
 			} else {
 				stopReset()
+				if currentConfig.ctype != newConf.ctype {
+					// If the plugin type has changed, we need to clean up the plugin state in the database,
+					// as it's no longer relevant for the new plugin type.
+					if err := deleteByChannelID(c.pluginCtx, c.db, c.ID); err != nil {
+						c.Logger.Warnw("Failed to clean up channel plugin state after plugin type change", zap.Error(err))
+					}
```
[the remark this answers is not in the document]
yhabteab: Wrapped it in a `retry.WithBackoff` with a 5m timeout.

internal/channel/plugin_test.go
```diff
@@ -0,0 +1,276 @@
...
+		require.Eventually(t, func() bool { plugin1 = getPluginS(ch); return plugin1 != nil }, 5*time.Second, 100*time.Millisecond)
+		require.NotNil(t, plugin1)
+
+		// checkSleepy is a helper function to test the plugin's sleep duration by sending a notification request and measuring the time taken.
+		checkSleepy := func(timeout time.Duration, expectedDuration time.Duration) {
+			ctx, cancel := context.WithTimeout(t.Context(), timeout)
+			defer cancel()
+			require.ErrorContains(t,
+				plugin1.SendNotification(ctx, makeTestRequest("sleep", false, false)),
+				fmt.Sprintf("plugin slept for %s", expectedDuration))
```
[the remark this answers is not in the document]
yhabteab: Fixed.

internal/channel/state.go
```diff
@@ -0,0 +1,84 @@
...
+	defer func() { _ = stmt.Close() }()
+
+	if _, err := stmt.ExecContext(ctx, channelID, key); err != nil {
+		return errors.Wrap(err, "failed to delete state for channel state")
+	}
+	return nil
+}
+
+// deleteByChannelID deletes all states associated with a specific channel ID from the database.
+func deleteByChannelID(ctx context.Context, db *database.DB, channelID int64) error {
```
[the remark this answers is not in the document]
yhabteab: Done.

yhabteab:
The Go linter is starting to hallucinate as well :(
```
  Error: internal/channel/channel.go:78:2: G118: Goroutine uses context.Background/TODO while request-scoped context is available (gosec)
  	go c.pluginControlLoop(newConfig{c.Type, c.Config})
  	^
```

yhabteab:
> The Go linter is starting to hallucinate as well :(
> 
> ```
>   Error: internal/channel/channel.go:78:2: G118: Goroutine uses context.Background/TODO while request-scoped context is available (gosec)
>   	go c.pluginControlLoop(newConfig{c.Type, c.Config})
>   	^
> ```

Hrrmm, I guess, it's complaining about this code block because the location it's pointing to doesn't make any sense.
https://github.com/Icinga/icinga-notifications/blob/febf595346cb78cf0f167e994add42bb0b74b940/internal/channel/channel.go#L130-L138

This however is within a `<-ctx.Done()` case, so using the just canceled context for other operation wouldn't also make any sense since the context is already canceled. It's so annoying!

internal/channel/state.go
```diff
@@ -0,0 +1,131 @@
...
+	)
+}
+
+// deleteStates deletes the provides states from the database, matching the channel ID and state key for each state.
+func deleteStates(ctx context.Context, db *database.DB, states []*State) error {
+	return retry.WithBackoff(
+		ctx,
+		func(ctx context.Context) error {
+			return db.ExecTx(ctx, nil, func(ctx context.Context, tx *sqlx.Tx) error {
+				stmt, err := db.PrepareNamedContext(ctx, `DELETE FROM channel_state WHERE channel_id = :channel_id AND state_key = :state_key`)
```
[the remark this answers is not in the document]
yhabteab: Yep. Such a stupidity is exactly I wanted to you catch :).

cmd/channels/email/main.go
```diff
@@ -156,29 +192,63 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
...
 	}
 
-	if len(to) == 0 {
+	if to == nil {
 		return fmt.Errorf("contact user %s does not have an e-mail address", req.Contact.FullName)
 	}
 
+	if err := ch.loadState(req.States); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
yhabteab: > Decode the supplied states into a request-local map and use that for this send.

Nice catch and also a good suggestion. Fixed.

doc/10-Channels.md
```diff
@@ -377,6 +479,38 @@ if the channel plugin config changes at runtime, Icinga Notifications will call
...
 
+### Channel State
+
+Channel plugins can store their state in Icinga Notifications' database using the [`channel::UpsertState`](#upsertstate)
+method. This allows channel plugins to persist their state across restarts, which can be useful for scenarios where the
+channel plugin needs to maintain information about previously sent notifications or other relevant data. The state is
+stored as key-value pairs in combination with the channel plugin's database unique identifier known only to Icinga
+Notifications. Each of the key-value pairs are represented by the
+[`State` type](https://pkg.go.dev/github.com/icinga/icinga-go-library/notifications/plugin#State) and contains
+the following fields:
```
```suggestion
Notifications. Each key-value pairs is represented by the
[`State` type](https://pkg.go.dev/github.com/icinga/icinga-go-library/notifications/plugin#State) and contains
the following fields:
```

doc/10-Channels.md
```diff
@@ -165,6 +170,103 @@ This request will result in a log message in Icinga Notifications' logging syste
...
+
+And here is an example of an error response, if the channel plugin tries to upsert a state with a key that is too long:
+
+```json
+{
+  "id": 1022,
+  "jsonrpc":"2.0",
+  "error": {
+    "code": -32602,
+    "message": "jsonrpc2: code -32600 message: invalid state key, must be non-empty and at most 255 chars, \"xxxxxxx<...>\" given"
```
oxzi: 1. `id` 102 for consistency with 101 above and 103 below.
2. The error `code` does not match the code in the `message`.

doc/10-Channels.md
```diff
@@ -165,6 +170,103 @@ This request will result in a log message in Icinga Notifications' logging syste
...
+
+```json
+{
+  "id": 101,
+  "jsonrpc": "2.0",
+  "method": "channel::UpsertState",
+  "params": [
+    {
+      "state_key": "john@doe.com.#12345",
+      "value": "{\"last_mail_message_id\": \"<bb7364a5-8f24-4b6a-a2d8-4658f42cd5cd-john@doe.com>\"}"
```
oxzi: It is called `last_message_id`.

https://github.com/Icinga/icinga-notifications/blob/990a99fc9647f04b7fab8b883742288b6ba05dda/cmd/channels/email/main.go#L58

Also requires an update in the paragraph below.

doc/10-Channels.md
```diff
@@ -165,6 +170,103 @@ This request will result in a log message in Icinga Notifications' logging syste
...
+    },
+    {
+      "state_key": "john.doe@example.com.#12346",
+      "value": "{\"last_mail_message_id\": \"<2e9dd0b2-9555-4627-871e-c8696e742486-john.doe@example.com>\"}"
```
oxzi: Unsure if this is too sophisticated for this example doc, but nevertheless: The `state_key` implies we are dealing with the Incident ID 12345 or 12346 for the recipient `john.doe@example.com`. However, the Message-ID contains the sender's email as the second part, not the recipient's.

And yes, I am aware that there is a difference between this example and the `email` channel (readable `state_key` instead of a hash; multiple upserts at once), but after I saw this, I failed to unsee it.

cmd/channels/email/main.go
```diff
@@ -156,29 +188,61 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
...
+			return err
+		}
+		b = b.Header("In-Reply-To", s.LastMessageID).Header("References", s.LastMessageID)
+	}
+
+	if err := b.Text(msg.Bytes()).Send(ch); err != nil {
+		return err
+	}
+
+	if req.Incident != nil && req.Incident.IsRecovered {
```
oxzi, line 232: Unless I am missing something, `IsRecovered` is nowhere set outside the tests. Thus, the following is currently unreachable code; similar in the `sleep` channel.

cmd/channels/email/main.go
```diff
@@ -222,3 +286,18 @@ func (ch *Email) Send(reversePath string, recipients []string, msg []byte) error
...
+// makeStateKey composes a unique key for the channel state based on the recipient's email address and the incident ID.
+func makeStateKey(to *mail.Address, i *plugin.Incident) string {
+	if i == nil {
+		return ""
+	}
+	h := sha256.New()
+	if err := binary.Write(h, binary.BigEndian, i.Id); err != nil {
+		return ""
+	}
+	if err := binary.Write(h, binary.BigEndian, []byte(to.String())); err != nil {
```
oxzi: `to.String` also contains the name. So, when updating the  contact name, the hash would change.

https://github.com/Icinga/icinga-notifications/blob/990a99fc9647f04b7fab8b883742288b6ba05dda/cmd/channels/email/main.go#L194

internal/channel/state.go
```diff
@@ -0,0 +1,131 @@
...
+
+// upsertStates upserts (inserts or updates) the provided states into the database.
+func upsertStates(ctx context.Context, db *database.DB, states []*State) error {
+	query, _ := db.BuildUpsertStmt(new(State))
+	return retry.WithBackoff(
+		ctx,
+		func(ctx context.Context) error {
+			return db.ExecTx(ctx, nil, func(ctx context.Context, tx *sqlx.Tx) error {
+				// Since we don't have a control over what the channel plugin will send, we need to use prepared stmt for this.
+				stmt, err := tx.PrepareNamedContext(ctx, query)
```
oxzi, line 70: I am a bit unsure how this works out in an HA scenario. Is the default isolation level enough or do we need serializable here as for the Event Queue?

oxzi (commented):
In general, LGTM. Thanks!

Unfortunately, this review is a bit thin due to time constraints.

doc/10-Channels.md
yhabteab: > The error `code` does not match the code in the `message`.

That's because Icinga Notifications sends an error response with invalid params and the sleep test plugin returns that JSON-RPC error to the `SendNotification` call, which wraps it in an internal error. Changed it  now.

doc/10-Channels.md
yhabteab: Changed.

cmd/channels/email/main.go
yhabteab: That's right. But since we're going to rework that logic with #409 in the coming weeks, I decided to not change anything here (instead of extending the `contracts.Incident` yet again, it will probably go away with that issue, so I don't want to wast time on it now).

internal/channel/state.go
yhabteab: We don't need a serialization here since we don't how to serialize it. Normally, each plugin process should only upsert and delete a state entry that it just used and since two Icinga Notifications nodes should never send notification for the same incident in parallell, there should be no conflicts.

yhabteab:
> Unfortunately, this review is a bit thin due to time constraints.

Yeah, but without a proper review, I don't want to merge it either, so it will be waiting till you're back. Thanks.

internal/channel/plugin.go
```diff
@@ -200,35 +213,157 @@ func (h rpcHandler) Handle(ctx context.Context, conn *jsonrpc.Conn, req *jsonrpc
...
+			return // deserializePluginState already sent an error reply to the plugin.
+		}
+
+		dbCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
+		defer cancel()
+
+		if err := upsertStates(dbCtx, h.ps.db, stateRecords); err != nil {
+			h.ps.logger.Warnw("Failed to send upsert state error reply", zap.Error(err))
+			if err := jsonrpc.ReplyError(ctx, conn, req.ID, jsonrpc.CodeInternalError, "failed to upsert state"); err != nil {
+				h.ps.logger.Error("Failed to send upsert state error reply", zap.Error(err))
```
oxzi:
```suggestion
				h.ps.logger.Errorw("Failed to send upsert state error reply", zap.Error(err))
```

Same a few more times below. (We should look into a golangci-lint check for this.)

internal/channel/plugin.go
```diff
@@ -182,6 +185,16 @@ func (p *pluginSupervisor) SetConfig(ctx context.Context, config string) error {
...
+	dbCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+
+	if states, err := getStateByChannelID(dbCtx, p.db, p.ChannelID); err != nil {
+		return fmt.Errorf("cannot retrieve channel state: %w", err)
+	} else {
+		for _, s := range states {
+			req.States = append(req.States, s.State)
+		}
+	}
```
oxzi, line 202: Honestly, I am a bit afraid how this will play out in the long run. This attaches the complete channel state (`SELECT * FROM channel_state WHERE channel_id = ?`) for each notification.

Considering the `email` channel, there is one row for each incident ID and address combination, carrying a JSON object of `{"last_message_id": "__LAST_MESSAGE_ID__"}`. (Note: there is no cleanup at the moment, see https://github.com/Icinga/icinga-notifications/pull/483#discussion_r3776625910.) Using an example payload similar to your docs, I would end up with the following bytes being transmitted every time - including DB fetching, JSON serialization, and JSON deserialization.

```
$ cat payload
{"state_key": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA", "value": "{\"last_message_id\": \"<2e9dd0b2-9555-4627-871e-c8696e742486-sender@example.com>\"}"}
$ wc -c payload
179 payload
```

For 1k unresolved incidents - or one hundred incidents dispatched to ten email addresses -, this would be a payload of 179 KB for each notification. While 1k is a relatively small number, consider hundreds of thousands unresolved incidents. And all this is sent back and forth just for the email channel plugin to sequentially scan all rows and only act upon the one row with the matching key.

---

This architecture allows outsourcing the logic to the channel plugin, which makes totally sense and allows the required flexibility. However, for lots of unresolved incidents, this might become noticeably expensive. Personally, I do not want to complicate the API any further by introducing an endpoint to explicitly fetch certain `state_keys`.

So, how about adding a warning log message if the amount of rows exceed some high arbitrary number when fetched in `getStateByChannelID`. This indicates that something might be wrong and things might get slow. Following my simple math above, let's say 10k lines, which are round about 1.8 MB of JSON.

internal/channel/channel.go
```diff
@@ -117,6 +129,14 @@ func (c *Channel) pluginControlLoop(currentConfig newConfig) {
 		select {
 		case c.pluginCh <- current:
 		case <-c.pluginCtx.Done():
+			if errors.Is(context.Cause(c.pluginCtx), ErrChannelDeleted) {
+				c.Logger.Info("Channel has been deleted, purging its plugin state from the database")
+				ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
oxzi, line 135: Why those short five seconds? There is our usual retry logic in `deleteByChannelID`, but limiting it to five seconds might be a hard miss if the database is unhealthy during this time (update and restart, for example).

If this call fails, there is no other retry and we would end up with orphaned entries, still having a valid foreign key to the `deleted = 'y'` channel.

yhabteab: Because I don't want to block that long here since I'm not bound to the context from our main function anymore. I can raise it to 10s, but I wouldn't use more seconds. Btw. this only covers the case when deleting a channel while Icinga Notifications is running, but it might still miss some deletion events when it was stopped at that time. So, this is not a guarantee that we won't end up with orphaned channel states. But at east it should reduce the risk of that happening.

We will need some kind of cleanup mechanism for soft deleted entries including channels anyway, so the orphaned states will be cleaned up with that eventually.

internal/channel/plugin.go
yhabteab: > So, how about adding a warning log message if the amount of rows exceed some high arbitrary number when fetched in `getStateByChannelID`. This indicates that something might be wrong and things might get slow. Following my simple math above, let's say 10k lines, which are round about 1.8 MB of JSON.

An alternative solution would be to require the state to be incident specific and `SendNotification` would only fetch states matching the incident ID from notification request. That would also drastically reduce the trafic but also constrain the plugins to cache something that's only incident related.

oxzi: Honestly, I am fine with most solutions; including the current state. I only wanted to discuss this point, while trying not to prolong this PR any further. So, if you think this is enough, I would be fine with it.

internal/channel/channel.go
oxzi: Then I would argue to keep it as it is and add another retention in the future, cleaning up all entries where the joined channel is marked as deleted.

internal/channel/plugin.go
yhabteab: Ok, I think we should constrain the states to be incident specific for. We could then change that constraint if we actually have a use case for it or someone requests otherwise.

yhabteab: Or maybe not. This seems to involve a lot more changes than I though. Ok, I give up. I'll just add the log warning as requested.

yhabteab: Done.

internal/channel/plugin.go
yhabteab: Fixed.

oxzi (approved):
Seems good, thanks!

PR #486, opened 2026-08-12
listener: disallow unknown fields in `ProcessEvent`

yhabteab:
This PR prevents sources from providing unknown event fields instead of just accepting and discarding it silently. I've also added a basic test cases for this.


Jan-Schuppik (approved):
looks good to me

Issue #450, opened 2026-06-24
Notify new escalation recipients only once in case of a designated manager


nilmerg:
A recipient is related to a given incident and represents one or more contacts who are notified about updates to it. We have three roles of recipients at the moment:

* recipient
* subscriber
* manager

A *manager* is a *contact* who used the respective action in the UI. A *recipient* or *subscriber* is a *contact*, *contact group* or *schedule* and here's where differences get somewhat loose and what the original reason for this issue is. Because at first, without a designated manager, every *recipient* is just that. Only once a *contact* used the *Subscribe* action in the UI, it's role is changed to *subscriber*. But with a designated manager, no further *recipient* will be associated with an incident, only new *subscribers*. (i.e. New escalation recipients are automatically *subscriber*)

The reason for this is to notify contacts, identified by later escalations, continuously even in case of a designated manager. They're deemed more important it seems. And this is what I want to have changed. The goal is to reserve the role *subscriber* to *contacts* using the UI action.

## Expected Behavior

If a new escalation is triggered, the new set of _recipients_ is not handled differently than usual. But in case of a designated manager, they're notified **immediately** about the state of the incident **but also that there's a _manager_ associated and who it is.** Afterwards, they wont be notified later on unless they manually subscribe.

In addition to this, **designating a manager must also trigger a notification for every current *recipient*.**

## Technical Considerations

The UI is responsible to designate a *manager* and *subscriber* right now and this has always been identified as possible issue, which it now is. Icinga Notifications should not have to poll `incident_contact` in order to detect a new *manager*. We are currently about to introduce a job queue to the database in #394 and that's what I propose to use as communication channel and let Icinga Notifications **Web** insert jobs like *this contact is manager now* or *this contact subscribed to this incident*. Eventually, it will not manipulate the database anymore other than configuration related tables. (i.e. no more inserts into `incident_history`, yay!)

yhabteab:
> In addition to this, **designating a manager must also trigger a notification for every current _recipient_.**

That's not possible. Icinga Notifications Web is the one that changes the contact role in the database and not Icinga Notifications. NVM, it's states in the last section how this will be done :).

yhabteab:
> and let Icinga Notifications **Web** insert jobs like _this contact is manager now_ or _this contact subscribed to this incident_.

[...]

PR #487, opened 2026-08-14
Rework HA job queue entry claiming & add concurrent processing

yhabteab:
This PR reworks the existing event queue processing logic to allow for more generic and concurrent processing of job queue entries. The existing logic was tightly coupled to the `Event` type and its processing, which made it difficult to extend the queue processing to other types of jobs in the future, such as [Icinga Notifications quick actions](#450). For this reason, the existing logic has been refactored to allow for more generic processing of job queue entries as discussed offline with @nilmerg and @BastianLedererIcinga.

First and foremost, the database queueing table is no longer called `event_queue`, but rather `job_queue`, and its `Json` column has been replaced with a versioned, typed `envelope` column, so the table can eventually carry different types of jobs in the future. As a consequence, the `object_id` and `user_agent` columns have been completely removed. Object identity is now computed at processing time based on the job type and its payload, and the user agent is replaced by the aforementioned typed envelope, which can carry any metadata that is relevant to the job type.

Now, since we no longer have `object_id` in the job queue table, we can no longer use it to claim and lock a job queue entry for processing. Instead, a new `job_processing_lock` table has been introduced, which is used exclusively by Icinga Notifications to claim and lock a job queue entry and its associated object for processing. The table's primary key `object_id` column, and a unique and not nullable `job_queue_id` referencing the `job_queue` table, provides the same guarantees as the previously used LEFT JOIN and `object_id` based locking mechanism. So, this inevitably leads to refactoring the existing job queue entry claiming and locking logic, and has now two main phases:

1. First, it plain selects a batch of job queue entries that are in pending state without any locking or claiming. Next, it iterates over that batch and tries to convert the job queue entry's envelope into a typed job (currently only `Event`). If the conversion fails, the job queue entry is marked as failed and skipped. If the conversion succeeds, it computes the object identity of the job and tries insert a row into the `job_processing_lock` table. If the insert succeeds, it then compares and updates the job queue entry's state to `processing` and streams the claimed job to a channel for concurrent processing. If the **INSERT IGNORE** affects zero rows, it means that the object is already being locked by another worker, so the tx is rolled back and the job queue entry is skipped. Though, the compare and update of the job queue entry's state to `processing` can fail too, if we race and lose to another worker that has just claimed the job queue entry. In that case, the tx is rolled back and the job queue entry is skipped as well. All of this, except the plain select of the batch of job queue entries, is done in a single tx without using SERIALIZABLE isolation level.
2. The second phase is the actual processing of the claimed job queue entries, which is done concurrently by a pool of workers (currently limited to 4). Each worker receives a claimed job queue entry from the channel and do the actual processing of the job. If the processing succeeds, the job queue entry is marked as completed, and the associated `job_processing_lock` row is deleted. If the processing fails, the job queue entry is marked as failed, and the associated `job_processing_lock` row is deleted as well. The only time that we might leave a locked job and its associated `job_processing_lock` row is when the worker panics or Icinga Notifications shuts down unexpectedly, in which case the existing `retention.ResetPruner` will eventually reset the job queue entry's state to pending and delete the associated `job_processing_lock` row, so the job can be retried later.

Example of the new envelope structure for an `Event` job queue entry:

```json
{"version":1,"format":"event","payload":{"source_id":1,"id":"icinga-master2.devlab.com: ssh!1786712186500","name":"icinga-master2.devlab.com: ssh","url":"https://watchtower.devlab.com/icingaweb/icingadb/service?name=ssh\u0026host.name=icinga-master2.devlab.com","tags":{"environment":"0162e462daea7c02249e1995a7fff030c776380f","host":"icinga-master2.devlab.com","service":"ssh"},"severity":"warning","message":"connect to address icinga-master1.devlab.com and port 80: Connection refused HTTP CRITICAL - Unable to open TCP socket","muted":false,"muted_reason":"Checkable is not muted (no active downtime, no acknowledgement, and not flapping)","incident":true,"complete_relations":["object.type","host.name","host.display_name","services[*].name","services[*].display_name"],"relations":{"host":{"display_name":"icinga-master2.devlab.com","name":"icinga-master2.devlab.com"},"object":{"type":"service"},"services":[{"display_name":"ssh","name":"ssh"}]}}}
{"version":1,"format":"event","payload":{"source_id":1,"id":"icinga-master2.devlab.com: swap!1786712202311","name":"icinga-master2.devlab.com: swap","url":"https://watchtower.devlab.com/icingaweb/icingadb/service?name=swap\u0026host.name=icinga-master2.devlab.com","tags":{"environment":"0162e462daea7c02249e1995a7fff030c776380f","host":"icinga-master2.devlab.com","service":"swap"},"severity":"crit","message":"SWAP CRITICAL - 0% free (0MB out of 0MB) - Swap is either disabled, not present, or of zero size. ","muted":false,"muted_reason":"Checkable is not muted (no active downtime, no acknowledgement, and not flapping)","incident":true,"complete_relations":["object.type","host.name","host.display_name","services[*].name","services[*].display_name"],"relations":{"host":{"display_name":"icinga-master2.devlab.com","name":"icinga-master2.devlab.com"},"object":{"type":"service"},"services":[{"display_name":"swap","name":"swap"}]}}}
```

## Why removing the `object_id` column from the job queue table?

In the previous implementation, the `object_id` was populated by the listener when the job queue entry was enqueued, and it was of course used to claim and lock the job queue entry for processing. However, having the `object_id` in the job queue table limits each and every job queue entry to be associated with a single object, and of course, it requires every client to pre-compute the object identity before enqueuing. Till now, this was not a problem because the only client that was enqueuing jobs, was the Icinga Notifications listener, which is aware of the object identity. However, with #450, Icinga Notifications Web will also be able to enqueue jobs for quick actions that are very different from the existing and more complex events. As opposed to events, quick actions can be batched for multiple objects/incidents, and having the `object_id` in the job queue table would not make sense anymore, because a single job queue entry can target multiple objects (not yet supported, but is planned for the future). So, Icinga Notifications Web will have to scan the `object_id_tags` table to determine the complete list of tags, compute the object ID for each object, and then enqueue a separate job queue entry for each of them. According to @nilmerg, this slows down the UI interactions dramatically, and wants to avoid in any way possible. Consequently, it was decided to remove the `object_id` column from the job queue table, and instead let the job enqueuer (whoever that is) to just provide the id tags in its payload, and let Icinga Notifications do the heavy lifting.


yhabteab:
Added some basic enqueueing and dequeueing test cases for the new job queue processing logic now. The tests require #489 though.

yhabteab:
This is the baseline for https://github.com/Icinga/icinga-notifications/pull/493. I've tested both of them extensively and my dev setup is also using this branch and works fine.

PS: most of the code changes are new tests for job queue enqueueing and processing logic.

doc/03-Configuration.md
oxzi, line 1: `doc/05-Distributed-Setups.md` also references `event_queue` and needs an update.

internal/event/queue.go
```diff
@@ -48,51 +62,107 @@ type Queue struct {
...
+	// The Envelope contains the version and format of the payload it embodies. The format denotes the
+	// type of the processable entity, while the version is used to determine if the payload needs a
+	// migration step before it can be deserialized into a newer version of the processable entity of
+	// the same format.
+	Envelope Envelope `db:"envelope"`
+}
 
-	// ObjectId is the Object ID of the event, used to exclude related queue entries on the database level.
-	ObjectId types.Binary `db:"object_id"`
+type (
```
oxzi: Bit nitpicky, but please move the `Queue` struct type into this multi-`type` block.

internal/event/queue.go
```diff
@@ -132,55 +204,157 @@ func Enqueue(ctx context.Context, db *database.DB, ev *Event, objectID types.Bin
...
+
+			go func(j job) {
+				defer sem.Release(1)
+
+				if err := processAndFinalize(ctx, db, logger, j, cbs, updStmt, unlockStmt); err != nil {
+					logger.Errorw("Failed to process and finalize job queue entry",
+						zap.Stringer("id", j.Q.ID),
+						zap.Error(err))
+				}
+				logger.Debugw("Successfully processed and finalized job queue entry", zap.Stringer("id", j.Q.ID))
```
oxzi:
```suggestion
				} else {
					logger.Debugw("Successfully processed and finalized job queue entry", zap.Stringer("id", j.Q.ID))
				}
```

internal/event/queue.go
```diff
@@ -132,55 +204,157 @@ func Enqueue(ctx context.Context, db *database.DB, ev *Event, objectID types.Bin
...
+					var claimedSome bool // Whether at least one candidate was successfully claimed in this batch.
+					for _, q := range candidates {
+						var (
+							err error
+							ev  *Event
+						)
+
+						switch q.Envelope.Format {
+						case EnvelopeFmtEvent:
+							ev, err = q.toEvent()
```
oxzi, line 310: `toEvent` does not only returns an error for a broken serialization, but also if the version is too new. In case of an HA setup where another node was already updated, this will result in valid job queue entries being marked as failed.

internal/event/queue.go
```diff
@@ -40,6 +45,15 @@ const (
 	QueueStateError int16 = 64
 )
 
+var (
+	// errClaimLost is returned internally by tryClaim when a candidate could not be claimed because its
+	// target object is already locked, or because a concurrent node claimed the same entry first.
+	errClaimLost = errors.New("claim lost")
+
+	// errEnvelopeVersionToNew is returned when a queue entry's envelope version is newer than the version supported by this node.
+	errEnvelopeVersionToNew = errors.New("envelope version is too new")
```
oxzi:
```suggestion
	errEnvelopeVersionTooNew = errors.New("envelope version is too new")
```

internal/event/queue.go
yhabteab: I wanted to move it too, but left it as-is to not make the PR even more bigger, but if you explicitly request me to change it, then I'll happily change it.

internal/event/queue.go
yhabteab: Well, that's the price one has to pay in that case because I've no other way to make that work. You might be thinking to simply skip such entries, but the not yet updated node can enqueue events for that very same object too, so I would also need to prevent it to process events of already skipped objects due to a version mismatch, which I can't do it in a reasonable time and code complexity.

internal/event/queue.go
yhabteab: Obviously 🙄.

internal/event/queue.go
yhabteab: Or should I give it a try nonetheless?

yhabteab: This also applies to the default case. A newly updated node might enqueue an envelope with a completely new format not known to the current node, such events shouldn't be marked as failed either.

yhabteab: Committed a reasonable and working workaround separately here https://github.com/Icinga/icinga-notifications/pull/487/changes/d464971068f0179908b8518f4c4594cc6dcd5a5e. If you've any better suggestion, pls let me know.

doc/03-Configuration.md
yhabteab: Fixed.

internal/event/queue.go
yhabteab: Changed. Ty.

oxzi (approved):
LGTM and tested it successfully.

PR #489, opened 2026-08-17
tests: load DB config from real env variables

yhabteab:
The previously used env vars were introduced at the time where we didn't have a proper environment support for our config. Now, that we fully support configuring everything via env vars, the tests should also use the same env vars instead of requiring custom ones that needs to be parsed separately. The daemon config initialization is now wrapped in a sync.Once to prevent data races when running the tests in parallel both calling `daemon.SetConfigForTesting`.

The commit is cherry-picked from #483, so that it gets merged quickly and use it in other PRs as well.

Additionally, I've also moved the complex test logging construction from the incidents tests into its own `testutils.GetTestLogging` function to make it easier to use.

### DIFF
--- a/.github/workflows/tests_with_database.yml
+++ b/.github/workflows/tests_with_database.yml
@@ -6,6 +6,12 @@ on:
       - main
   pull_request: {}
 
+env:
+  ICINGA_NOTIFICATIONS_ICINGAWEB2_URL: https://localhost/icingaweb2
+  ICINGA_NOTIFICATIONS_DATABASE_DATABASE: notifications
+  ICINGA_NOTIFICATIONS_DATABASE_PASSWORD: notifications
+  ICINGA_NOTIFICATIONS_DATABASE_HOST: 127.0.0.1
+
 jobs:
   mysql:
     runs-on: ubuntu-latest
@@ -27,19 +33,16 @@ jobs:
           - mariadb:latest
 
     env:
-      NOTIFICATIONS_TESTS_DB_TYPE: mysql
-      NOTIFICATIONS_TESTS_DB: notifications
-      NOTIFICATIONS_TESTS_DB_USER: root
-      NOTIFICATIONS_TESTS_DB_PASSWORD: notifications
-      NOTIFICATIONS_TESTS_DB_HOST: 127.0.0.1
-      NOTIFICATIONS_TESTS_DB_PORT: 3306
+      ICINGA_NOTIFICATIONS_DATABASE_TYPE: mysql
+      ICINGA_NOTIFICATIONS_DATABASE_USER: root
+      ICINGA_NOTIFICATIONS_DATABASE_PORT: 3306
 
     services:
       mysql:
         image: ${{ matrix.image }}
         env:
-          MYSQL_ROOT_PASSWORD: ${{ env.NOTIFICATIONS_TESTS_DB_PASSWORD }}
-          MYSQL_DATABASE: ${{ env.NOTIFICATIONS_TESTS_DB }}
+          MYSQL_ROOT_PASSWORD: ${{ env.ICINGA_NOTIFICATIONS_DATABASE_PASSWORD }}
+          MYSQL_DATABASE: ${{ env.ICINGA_NOTIFICATIONS_DATABASE_DATABASE }}
         # Wait until MySQL becomes ready
         options: >-
           --health-cmd "${{ (startsWith(matrix.image, 'mysql:') || startsWith(matrix.image, 'mariadb:10')) && 'mysqladmin ping' || 'healthcheck.sh --connect --innodb_initialized' }}"
@@ -60,8 +63,8 @@ jobs:
 
       - name: Importing Schema
         run: >
-          mysql -h127.0.0.1 -u$NOTIFICATIONS_TESTS_DB_USER -p$NOTIFICATIONS_TESTS_DB_PASSWORD
-          $NOTIFICATIONS_TESTS_DB < ${{ github.workspace }}/schema/mysql/schema.sql
+          mysql -h127.0.0.1 -u$ICINGA_NOTIFICATIONS_DATABASE_USER -p$ICINGA_NOTIFICATIONS_DATABASE_PASSWORD
+          $ICINGA_NOTIFICATIONS_DATABASE_DATABASE < ${{ github.workspace }}/schema/mysql/schema.sql
 
       - name: Download dependencies
         run: go get -v -t -d ./...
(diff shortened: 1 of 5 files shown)


internal/incident/incidents_test.go
```diff
@@ -7,30 +7,35 @@ import (
...
-	logs := logging.NewLoggingWithFactory("testing", zapcore.DebugLevel, time.Second, func(level zap.AtomicLevel) zapcore.Core {
-		return zaptest.NewLogger(t, zaptest.Level(level.Level())).Core()
-	})
+	// The same check in testutils.GetTestDB() is too late, as we need to skip the entire
+	// test suite before we even try to load the configuration.
+	testutils.SkipIfNoDBConfig(t)
+
+	conf := new(daemon.ConfigFile)
+	require.NoError(t, baseConfing.FromEnv(conf, baseConfing.EnvOptions{Prefix: "ICINGA_NOTIFICATIONS_"}))
+	daemon.SetConfigForTesting(conf)
```
julianbrost: Is there a particular need why you need to do `daemon.SetConfigForTesting(conf)` here? At least from the other changes in this PR, I don't see an obvious reason for that. Couldn't you just populate the database config from the environment here, something like this?

```suggestion
	dbConfig := new(database.Config)
	require.NoError(t, baseConfing.FromEnv(dbConfig, baseConfing.EnvOptions{Prefix: "ICINGA_NOTIFICATIONS_DATABASE_"}))
```

By doing this inside `testutils.GetTestDB()`, it should also be possible to keep its signature unchanged and keep the callers of it simpler (although it just seems to be a single one at this time). Also, you shouldn't need `SetConfigForTesting()` then anymore as well.

internal/testutils/testutils.go
```diff
@@ -4,60 +4,52 @@ import (
...
 }
 
+// GetTestLogging returns a logging.Logging instance for testing purposes.
+//
+// It sets the logging level to Debug and uses a zaptest logger to capture logs during tests.
+func GetTestLogging(t *testing.T) *logging.Logging {
+	return logging.NewLoggingWithFactory("testing", zapcore.DebugLevel, time.Second, func(level zap.AtomicLevel) zapcore.Core {
+		return zaptest.NewLogger(t, zaptest.Level(level.Level())).Core()
+	})
+}
```
julianbrost, line 51: That change is neither mentioned in the PR description nor the commit message, so I'm not sure if this is in here intentionally (if so, please mention it as well), or made its way in here by accident during cherry-picking.

internal/incident/incidents_test.go
yhabteab: I need to initialize the daemon config singleton, otherwise the test cases from #483 crashes here:
https://github.com/Icinga/icinga-notifications/blob/027e2701958f697f8c0bc0dd5ac4a0acfedf1f19/internal/channel/plugin.go#L45-L46

internal/testutils/testutils.go
yhabteab: Ok, I'll mention it in the PR desc. This actually does nothing but extracting that complex logging construction into a helper function.

yhabteab: Done.

yhabteab: Btw. I've also now replaced that very same construct in listener tests with this new helper function.

internal/incident/incidents_test.go
julianbrost: No chance to make that test work without having to rely on global state? That might make things cleaner overall as you don't have to worry about multiple tests accessing the global state in the first place.

yhabteab: I could make it work, but I doubt that it's better option than this as I have to make the channels_dir a field of some struct (probably the `Channel` type and populate it from the env var in the tests and the production code uses `daemon.Config()` conditionally or the like.

yhabteab: NVM. It actually is quite easy to make that work, so I'll change this.

yhabteab: Changed it now as suggested.

.github/workflows/tests_with_database.yml
```diff
@@ -6,6 +6,12 @@ on:
       - main
   pull_request: {}
 
+env:
+  ICINGA_NOTIFICATIONS_ICINGAWEB2_URL: https://localhost/icingaweb2
```
julianbrost: I think that variable is now unused (I guess it was added because it might be required for the whole daemon config to validate).

yhabteab:
The incidents test now crash constantly when running in parallel with the newly introduced plugin tests in https://github.com/Icinga/icinga-notifications/pull/483 and I don't know, how I should fix that. The incidents test use `runtimeConfig.UpdateFromDatabase(t.Context())` to load some runtime config required for some test cases and since the plugin tests will insert some valid plugins into the database that runtime config method will pick them up and try to start them, which end up crashing while accessing `daemon.Config()` as there's no other way to provide the channelsDir otherwise in that case.

yhabteab:
Do you've an idea how to fix that or do I need to revert the previous version of this PR? Of course, without making `go test -p 1` a requirement for that.

julianbrost:
I only looked at the other PR very briefly, so take what I say with a grain of salt. Given that you go through the full dynamic runtime config update mechanism, I don't immediately see a way of getting this to work with minimal changes and without global state. You'd probably have to do some decoupling in the sense that in the runtime config, instead of directly starting the channel directly, this would be delegated to some implementation that could be swapped for some mock implementation for the tests. Though not sure how big of a change that would be, but as a bonus, that might allow for mocking even bigger parts of the channel so that it might be possible to perform more tests without having to set up some directory with channels, but instead use some in-process mock implementations.

yhabteab:
> Though not sure how big of a change that would be, but as a bonus, that might allow for mocking even bigger parts of the channel so that it might be possible to perform more tests without having to set up some directory with channels, but instead use some in-process mock implementations.

The unittests introduced in #483 are supposed to test the actual plugin spawning and managing logic using real plugin processes and not using some test doubles. Now, having to mock that instead wouldn't make any sense as that will probably be nothing comparable to the production code then. The only other alternative I see to using the global daemon.Config is not using `runtimeConfig.UpdateFromDatabase(t.Context())` in the tests at all. Instead, one could populate the maps needed for the incidents tests manually and be fine with it.

yhabteab:
I don't see any other trivial way to fix this, so I've reverted this with a slightly different version.

.github/workflows/tests_with_database.yml
```diff
@@ -60,8 +63,8 @@ jobs:
 
       - name: Importing Schema
         run: >
-          mysql -h127.0.0.1 -u$NOTIFICATIONS_TESTS_DB_USER -p$NOTIFICATIONS_TESTS_DB_PASSWORD
-          $NOTIFICATIONS_TESTS_DB < ${{ github.workspace }}/schema/mysql/schema.sql
+          mysql -h127.0.0.1 -u$ICINGA_NOTIFICATIONS_DATABASE_USER -p$ICINGA_NOTIFICATIONS_DATABASE_PASSWORD
```
oxzi: To match the other env vars, the hard-coded localhost could be replaced by `ICINGA_NOTIFICATIONS_DATABASE_HOST`. Same below for PostgreSQL.

internal/testutils/testutils.go
```diff
@@ -4,60 +4,50 @@ import (
...
-		c.Port = port
-	}
+	conf := new(T)
+	validator, ok := any(conf).(config.Validator)
+	require.True(t, ok, "type must implement config.Validator")
 
-	require.NoError(t, c.Validate(), "database config validation should not fail")
-
-	db, err := database.NewDbFromConfig(c, logging.NewLogger(zaptest.NewLogger(t).Sugar(), time.Hour), database.RetryConnectorCallbacks{})
+	require.NoError(t, config.FromEnv(validator, config.EnvOptions{Prefix: "ICINGA_NOTIFICATIONS_"}))
```
oxzi: Switching from the testing-only environment variables to the main config parsing results in `env:"…,unset"` to be applied. So, after one `GetTestDB` call, `ICINGA_NOTIFICATIONS_DATABASE_PASSWORD` would be unset and a second call would proceed with an empty password (which should validate).

Since there is only one `GetTestDB` call at the moment, this is currently fine, but might be a future footgun. Btw, a second call is introduced in #483.

oxzi (commented):
Next to the nitpick and the potential environment config unset issue, looks good to me. Thanks.

.github/workflows/tests_with_database.yml
yhabteab: Changed.

internal/testutils/testutils.go
yhabteab: Fixed.

yhabteab: Just for the record. This was not an issue since the tests are located in two different packages, and thus run in separate processes, unsetting the environment variable in one process does not affect the other. But fixed it anyway to be future proof.

oxzi (approved):

Issue #451, opened 2026-06-25
Loosen source linkage with other relations


nilmerg:
Sources are right now hard-linked with every `object` and `rule`. It gets more and more clear to me that this linkage isn't suitable anymore and must be softened.

## Do not incorporate source id into how object ids are generated

From the perspective of a source, the `object_tags` are the sole information to identify an object and differentiate it from others. Right now, the source id, an autoincrement integer, is being taken into account plus said object tags, in case Icinga Notifications constructs an object's id. This is wrong and prone to accidental object duplication, very similar to what happens in case Icinga DB accidentally gets a different environment id with the otherwise same configuration. If object tags are not enough to prevent duplicates across a source, a source has to provide more unique object tags.

If objects of different sources conflict, because they happen to use the same tags, that's fine. Separate rules, as filters depend on the source, are to be expected as well. But it's already expected that different rules apply to the same incident anyway.

The `object` table therefore doesn't need to have a `source_id` foreign key anymore, but for the visual experience in Icinga Notifications Web I still like to see a link between sources and objects, but in a `nm` relation.

## Allow more than one source be a provider for a particular escalation rule

This changes the `1n` relation of `source <- rule` to a `nm` relation `source <-> rule` instead. Icinga Notifications Web already does not require a particular source for an escalation rule, it only needs a type identifier. The same must therefore be the case for Icinga Notifications as it must then be able to identify rules the same way.

This opens up the possibility of a contradiction as it must be invalid for differently *typed* sources being linked with the same escalation rule. Icinga Notifications Web is responsible to ensure this but the database should also forbid this by using appropriate constraints.

julianbrost:
Not like we've already changed this once the other way around 🙈

> If objects of different sources conflict, because they happen to use the same tags, that's fine.

What does fine mean? Should be handled in some way or you don't care what happens in this case? I mean that could easily result in the object flapping between the states provided by the different source all the time (unless this somehow implies tracking state per object and source and an object potentially having multiple different states at once).

nilmerg:
> Not like we've already changed this once the other way around 🙈

Can't remember a similar change.

> don't care

Is essentially what my view in this regard is and what this issue is about. It's the source's responsibility to avoid such conflicts, not ours. And even if, definitely not by baking in a random integer that's not configurable.

nilmerg:
This just got a whole lot more interesting. Updated the OP.

nilmerg:
[...]

PR #493, opened 2026-08-18
Process incident quick actions like regular events

### LINKED ISSUE #450: Notify new escalation recipients only once in case of a designated manager

yhabteab:
This PR introduces a new job queue envelope format `QuickAction` that will be used by Icinga Notifications Web to enqueue incident `manage`, `subscribe` quick actions into the now more general `job_queue` database table. These events have nothing special other than having a different format than the existing `event.Event`, and will therefore be processed just like the other regular events. The format for this new envelope payload is described in https://github.com/Icinga/icinga-notifications/issues/450#issuecomment-5267692701.

Additionally, it makes sure that new escalation recipients will be notified only once in case the incident was managed by someone else at the time the escalation was triggered as requested in #450.

## Requires

- #487
- #451

resolves #450


internal/event/queue.go
```diff
@@ -161,6 +167,27 @@ func (q *Queue) toEvent() (*Event, error) {
 	return &ev, nil
 }
 
+// toQuickAction converts the Queue entry into an event.QuickAction.
+func (q *Queue) toQuickAction() (*QuickAction, error) {
```
oxzi: This method is very similar to `Queue.toEvent` but differs slightly, for example differing error strings or JSON decoding to a pointer variable or to a dereferenced variable. Please either keep them identical or consider refactoring them to a common method.

internal/incident/incident.go
```diff
@@ -350,9 +359,103 @@ func (i *Incident) Close(ctx context.Context, tx *sqlx.Tx) error {
...
+				zap.Stringer("action", qa.Kind),
+				zap.Error(err))
+			return fmt.Errorf("cannot restore incident state: %w", err)
+		}
+
+		switch qa.Kind {
+		case event.ActionManage:
+			if i.HasManager() {
+				i.logger.Warnw("Cannot add recipient as manager, incident already has a manager", zap.Stringer("recipient", r))
+				return errors.New("incident already has a manager")
```
oxzi: While the warning log level seems reasonable, the error bubbles up to `Process`, to the `processAndFinalize` callbacks, where it is logged again at error level. Same for the following cases.

internal/incident/incident.go
```diff
@@ -350,9 +359,103 @@ func (i *Incident) Close(ctx context.Context, tx *sqlx.Tx) error {
 	return nil
 }
 
-// ErrSeverityChangeWithoutIncidentFlag is returned when an event tries to change the severity of an incident
-// but does not set the 'incident' flag.
-var ErrSeverityChangeWithoutIncidentFlag = errors.New("cannot change severity of an incident with an event that doesn't set the 'incident' flag")
+// DoQuickAction executes the given quick action for the current incident in an own transaction.
+func (i *Incident) DoQuickAction(ctx context.Context, qa *event.QuickAction) error {
+	recipientKey := recipient.Key{ContactID: types.MakeInt(qa.ContactID)}
+	r := i.runtimeConfig.GetRecipient(recipientKey)
```
oxzi: Other calls to `GetRecipient` have an `RLock` on their path. Same should apply to transitive calls via `Incident`, such as `i.HasManager()` below.

yhabteab: Nice catch.

internal/event/queue.go
yhabteab: Fixed the inconsistencies but didn't refactore it out.

internal/incident/incident.go
yhabteab: Removed the extra logs in the incident file.

oxzi (dismissed):
LGTM, thanks. Let's hope for the actions 🙃

yhabteab:
Unfortunately, I had to rebase this since it conflicted with just merged PR, please have a look again. Thanks.

yhabteab:
The tests are failing due to cross-package parallel testing and table trancations. I've an idea how to elegantly fix this without making `go test -p 1` a requirement again.

yhabteab:
I've fixed all the cross-package parallel testing problems now.

oxzi (approved):
Works for me, thanks!

PR #494, opened 2026-08-18
Loosen source linkage with other relations

### LINKED ISSUE #421: object.ID: sortedTags contains empty elements
The `ID` function in `internal/object/object.go` creates the `sortedTags` slice with `len(tags)` empty fields, before adding `len(tags)` populated fields.

https://github.com/Icinga/icinga-notifications/blob/710c8184d8ed036fef3bede2838d4c4be5015a53/internal/object/object.go#L205-L216

This is most likely a bug and the code should be `make([]KV, 0, len(tags))` to create the slice with a predefined capacity. However, due to the function warning this is maybe not something we can just fix.

https://github.com/Icinga/icinga-notifications/blob/710c8184d8ed036fef3bede2838d4c4be5015a53/internal/object/object.go#L187-L189

A simplified demo is available on the Go Playground: https://go.dev/play/p/TtBg3FsIG0e

Thanks to @BastianLedererIcinga for finding this issue while working on a reimplementation on the web side.

### LINKED ISSUE #451: Loosen source linkage with other relations

yhabteab:
This PR loosens the linkage between sources and other relations, such as objects and event rules. Previously, objects and event rules were linked to an exactly one source, which became a problem when using two different Icinga DB instances running in HA mode, each with its own source. Objects are now no longer linked directly to a source but through the new `object_source` table instead, and their IDs is generated purely from their ID tags. Likewise, the event rules are linked to a source as a many-to-many relationship, allowing for the same event rule to be linked to multiple similar sources. This is done by replacing the existing `rule.source_id` column with a `rule.source_type`, which links event rules to sources of that very same type but without any foreign keys.

resolves #451
fixes #421


schema/pgsql/schema.sql
```diff
@@ -314,6 +312,15 @@ CREATE TABLE object_id_tag (
     CONSTRAINT fk_object_id_tag_object FOREIGN KEY (object_id) REFERENCES object(id)
 );
 
+CREATE TABLE object_source_association (
```
nilmerg: `object_source` please. It'd be in-line with e.g. `incident_rule` and e.g. `icingadb.notification_user`.

schema/pgsql/schema.sql
```diff
@@ -383,6 +388,15 @@ CREATE TABLE rule_escalation_recipient (
 
 CREATE INDEX idx_rule_escalation_recipient_changed_at ON rule_escalation_recipient(changed_at);
 
+CREATE TABLE rule_source_association (
```
nilmerg: `rule_source` please. It's also better for ipl-orm since it would assume either `rule_source_id` or `source_association_id` instead of just `rule_id` and `source_id`.

schema/pgsql/upgrades/loosen-source-linkage.sql
```diff
@@ -0,0 +1,26 @@
+CREATE TABLE object_source_association (
+    object_id bytea NOT NULL,
+    source_id bigint NOT NULL, -- The specific source that the object referenced by object_id is associated with.
+
+    CONSTRAINT pk_object_source_association PRIMARY KEY (object_id, source_id),
+    CONSTRAINT fk_object_source_association_object FOREIGN KEY (object_id) REFERENCES object(id),
+    CONSTRAINT fk_object_source_association_source FOREIGN KEY (source_id) REFERENCES source(id)
+);
+
+ALTER TABLE object
```
nilmerg, line 14: Why not migrate links prior like it's done below for rules?

schema/pgsql/schema.sql
```diff
@@ -383,6 +388,15 @@ CREATE TABLE rule_escalation_recipient (
...
+    rule_id bigint NOT NULL,
+    source_id bigint NOT NULL, -- The specific source that the rule referenced by rule_id is associated with.
```
nilmerg: From the issue:

> This opens up the possibility of a contradiction as it must be invalid for differently typed sources being linked with the same escalation rule. Icinga Notifications Web is responsible to ensure this but the database should also forbid this by using appropriate constraints.

Why isn't this the case? Isn't it simply possible to add `source_type` as column here as well and introduce a check that verifies no rule is linked with different types?

nilmerg (changes requested):
sql only for now

schema/pgsql/upgrades/loosen-source-linkage.sql
yhabteab: Because their ID is going to be invalidated anyway, and thus will no longer be associated with that source.

schema/pgsql/schema.sql
yhabteab: > Why isn't this the case? Isn't it simply possible to add `source_type` as column here as well and introduce a check that verifies no rule is linked with different types?

Really? You want me to add a column just so I can use it for a check constraint but for nothing else? I didn't implement this because the requirement doesn't make any sense to me. A link table should just, well, link different tables and doesn't need to get that complicated.

schema/pgsql/upgrades/loosen-source-linkage.sql
nilmerg: Right. :man_facepalming:

schema/pgsql/schema.sql
nilmerg: No. We concluded in person that it suffices to introduce `rule.source_type` and drop `rule_source_association` completely. Web is satisfied with this information and the daemon should also be able to fetch rules based on the conditon `source.type == rule.source_type`.

schema/pgsql/schema.sql
yhabteab: Changed.

schema/pgsql/schema.sql
yhabteab: We don't have a link table for this any more.

schema/pgsql/schema.sql
yhabteab: Changed.

internal/config/source.go
```diff
@@ -30,7 +31,7 @@ type Source struct {
 	// Each of these IDs corresponds to a rule in the [ConfigSet.Rules] map and is used to quickly access
 	// the rules for a specific source without iterating over all rules. It is not stored in the database,
 	// but is updated when applying pending rules in [RuntimeConfig.applyPendingRules].
-	ruleIDs []int64
+	ruleIDs map[int64]struct{}
```
nilmerg, line 34: just a question, because to me it seems this is only a change to simplify key lookups. This is really required in Go and the compiler does not normalize things like this on its own?

internal/object/object.go
```diff
@@ -90,53 +91,20 @@ func (o *Object) DisplayName() string {
...
+//
+// The identifier is generated by creating a SHA256 hash of the concatenated key-value pairs of the tags,
+// sorted by key to ensure consistent ordering. The resulting hash is returned as a [types.Binary] value.
 //
-// TODO: the return value of this function must be stable like forever
-func ID(source int64, tags map[string]string) types.Binary {
+// The implementation of this function must remain stable to not break existing object references in the database.
+// Any changes to the algorithm must be carefully considered and documented, as they may affect the integrity of
+// existing data.
+func ID(tags map[string]string) types.Binary {
```
nilmerg, line 102: Is there a test somewhere else, since there's no `object_test.go` in this directory? If not, the referenced issue demands a regression test IMHO.

nilmerg (commented):
*Personally* (read: without AI) looked at the Go changes and they mostly seem legit to me. But with one caveat, as I don't have the required knowledge about the file hierarchy here and callsites, I didn't verify anything further than what the diff on Github indicates. (But I didn't saw any interface/signature changes, except the ID function, anyway…)

The change runs also fine, the migrated rules match and notifications are sent.

internal/config/source.go
yhabteab: It's not just for efficiency but to also eliminate possible duplicates. Previously it was just a simple slice and could contain the same rule ID multiple times. Instead of having to search for duplicates in the slice by myself, I changed it to map (always unique keys) to automatically do that for me and it's value `struct{}` is not a real value and allocates exactly 0 bytes of memory (meaning not at all).

internal/object/object.go
yhabteab: Added a new test.

nilmerg (dismissed):
Looks fine, my quick PHP implementation also yields the same result as the test:

```php
<?php

function ID(array $tags): string {
    $buffer = '';
    ksort($tags);
    foreach ($tags as $k => $v) {
        $buffer .= $k;
        $buffer .= chr(0);
        $buffer .= $v;
        $buffer .= chr(0);
    }
    
    return hash('sha256', $buffer);
}

$tags1 = [
	"foo" => "bar",
	"baz" => "qux",
];
$tags2 = [ // same as tags1 but with different order.
	"baz" => "qux",
	"foo" => "bar",
];
$tags3 = [
	"some" =>      "thing",
	"different" => "value",
];
	
echo ID($tags1) . PHP_EOL;
echo ID($tags2) . PHP_EOL;
echo ID($tags3) . PHP_EOL;
```

> 4bfe8b2596005172c9db4d2b4b400a12b478b87a793ed9577e9d2d165fd07e7a
> 4bfe8b2596005172c9db4d2b4b400a12b478b87a793ed9577e9d2d165fd07e7a
> d42b5e17d0868536dcf004e2aac9b8478a632e03cb666eb0dc96c3a3f608dcfa

yhabteab:
Now, I'm done. I've extended the test cases a bit and didn't change anything else.

yhabteab:
Just noticed that I forgot to truncate the `object` and `rule` tables in PostgreSQL case in the incident tests. Since the source table is no longer linked to them directly, just truncating the `source` table isn't enough anymore, so I've added these tables to the list of tables to be truncated.

internal/config/source.go
```diff
@@ -81,16 +82,21 @@ func (source *Source) PasswordCompare(password []byte) error {
...
+	}
+	source.ruleIDs[ruleID] = struct{}{}
 }
 
 // deleteRuleID removes a rule ID from the list of rule IDs belonging to this source.
 func (source *Source) deleteRuleID(ruleID int64) {
-	source.ruleIDs = slices.DeleteFunc(source.ruleIDs, func(id int64) bool { return id == ruleID })
+	if source.ruleIDs != nil {
+		delete(source.ruleIDs, ruleID)
+	}
```
oxzi: A `nil` guard should not be necessary for `delete`.

>  If the map m is nil or the element m[k] does not exist, delete is a no-op.

https://go.dev/ref/spec#Deletion_of_map_elements

schema/mysql/upgrades/loosen-source-linkage.sql
```diff
@@ -0,0 +1,28 @@
...
+
+    CONSTRAINT pk_object_source PRIMARY KEY (object_id, source_id),
+    CONSTRAINT fk_object_source_object FOREIGN KEY (object_id) REFERENCES object(id),
+    CONSTRAINT fk_object_source_source FOREIGN KEY (source_id) REFERENCES source(id)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
+
+DELETE FROM incident_history;
+DELETE FROM incident_rule_escalation_state;
+DELETE FROM incident_rule;
+DELETE FROM incident;
```
oxzi, line 14:
```
--------------
DELETE FROM incident
--------------

ERROR 1451 (23000) at line 13: Cannot delete or update a parent row: a foreign key constraint fails (`notifications2`.`incident_contact`, CONSTRAINT `fk_incident_contact_incident` FOREIGN KEY (`incident_id`) REFERENCES `incident` (`id`))
```

https://github.com/Icinga/icinga-notifications/blob/508a9df254b3554135c11465c873db31b61d1afc/schema/mysql/schema.sql#L398

---

For the PostgreSQL schema upgrade, there should be no such problem:

https://github.com/Icinga/icinga-notifications/blob/508a9df254b3554135c11465c873db31b61d1afc/schema/pgsql/upgrades/loosen-source-linkage.sql#L9-L10

internal/config/source.go
```diff
@@ -81,16 +82,21 @@ func (source *Source) PasswordCompare(password []byte) error {
 }
 
 // RuleIDs returns the list of rule IDs belonging to this source.
-func (source *Source) RuleIDs() []int64 { return source.ruleIDs }
+func (source *Source) RuleIDs() []int64 { return slices.Collect(maps.Keys(source.ruleIDs)) }
```
oxzi: How about dropping the `slices.Collect` and directly return the `iter.Seq[int64]`? At the moment, both calls of this method are going to iterate over it anyway.

schema/pgsql/schema.sql
```diff
@@ -314,6 +312,15 @@ CREATE TABLE object_id_tag (
     CONSTRAINT fk_object_id_tag_object FOREIGN KEY (object_id) REFERENCES object(id)
 );
 
+CREATE TABLE object_source (
+    object_id bytea NOT NULL,
+    source_id bigint NOT NULL, -- The specific source that the object referenced by object_id is associated with.
+
+    CONSTRAINT pk_object_source PRIMARY KEY (object_id, source_id),
+    CONSTRAINT fk_object_source_object FOREIGN KEY (object_id) REFERENCES object(id),
```
oxzi, line 334: Since you implemented retention based on this key, an index should be necessary for PostgreSQL; see #479.

schema/mysql/upgrades/loosen-source-linkage.sql
yhabteab: Hrrrm, obviously. I've just looked on the list of tables cleared here but there incident_contact is not there because the tests didn't insert into that table.
https://github.com/Icinga/icinga-notifications/blob/c33169bef4b7db401787f276582a25049ff92ad0/internal/incident/incidents_test.go#L382-L391

internal/config/source.go
yhabteab: Nice idea.

schema/pgsql/schema.sql
yhabteab: But as opposed to the referenced PR this is part of the primary key. Shouldn't that be enough?

yhabteab: Never mind. object_id_tag has a similar composed PK und an extra index on the object_id column. I'll add one.

oxzi: > But as opposed to the referenced PR this is part of the primary key. Shouldn't that be enough?

In my understanding, this happens automatically in InnoDB, but not for PostgreSQL - especially not for combined PKs.

_Edit: Just saw your second answer now.. GitHub's partial reloads are sometimes quite an annoyance._

yhabteab:
Addressed the requested changes and also rebased against the latest main branch.

schema/pgsql/schema.sql
yhabteab: > In my understanding, this happens automatically in InnoDB, but not for PostgreSQL

Yep. InnoDB automatically create an index for columns with foreign key constraint while PostgreSQL does nothing and has to be done manually when needed. But since the object_id is part of the composed PK, I though that would be enough index to speed up the clean up queries, but no, it requires an extra one.

oxzi (approved):
Looks good, thanks!

PR #496, opened 2026-08-19
Add missing `external_uuid` fields & change type to `binary(16)` in MySQL

Jan-Schuppik:
Change the external_uuid column on channel, contact, and contactgroup from char(36) to binary(16) in the MySQL schema, with an upgrade migration that converts existing values in place. Update the affected Go structs and test fixtures to use types.UUID instead of raw strings, and add MakeRandomUUID to testutils for generating fixture values.


nilmerg:
> Change the external_uuid column on channel, contact, and contactgroup from char(36) to binary(16) in the MySQL schema

As discussed, I'd rather want to see the same approach for new UUID related columns as we have right now. PostgreSQL is probably our target platform anyway and I don't care about the limitations of MySQL. It's right now not justifiable enough for me to introduce this refactoring since it causes additional refactoring in Notifications Web for no real advantage.

yhabteab:
> Update the affected Go structs and test fixtures to use types.UUID instead of raw strings, and add MakeRandomUUID to testutils for generating fixture values.

Nevertheless, please submit the changes mentioned in the PR description that are still missing. I don't know why they were left out initially when these columns were first introduced but it's become so annoying lately not able to insert something into these tables from within Icinga Notifications.

julianbrost:
> I don't know why they were left out initially when these columns were first introduced but it's become so annoying lately not able to insert something into these tables from within Icinga Notifications.

When those columns were added, the question whether to store the UUID as raw binary or string representation already came up. The latter was picked because at that time, the assumption was that these columns are solely for Icinga Notifications Web and the Go part doesn't even need to care about them (that's why #216 only added the columns to the schema and nothing on the Go side), so Team Web picked the column type that worked best for them.

nilmerg:
> that these columns are solely for Icinga Notifications Web and the Go part doesn't even need to care about them

And this doesn't change. The reason this came up again is that @yhabteab wrote Go tests that need to create fixtures for those types (`contact`, etc) and the column `external_uuid` is not nullable, which forces him to care about it.

The quick solution to this is to make the `external_uuid` nullable, but only because I'm not familiar with how Go abstracts database writes, so maybe there's an alternative to this? Like a write-only or test-specific approach.

yhabteab:
> the Go part doesn't even need to care about them (that's why #216 only added the columns to the schema and nothing on the Go side), so Team Web picked the column type that worked best for them.

Even if the Go part doesn't need to care about them, it should still have been added the corresponding fields to the Go structs, instead of just pretending they don't exist. This now forces us to either introduce a surrogate struct that has the fields we need, so we can use them to dynamically build the query and insert something into these tables, which I don't want to do, or do the work and add the fields to the Go structs with a proper type, which we should have done a long time ago.

> The reason this came up again is that @yhabteab wrote Go tests that need to create fixtures for those types (`contact`, etc) and the column `external_uuid` is not nullable, which forces him to care about it.

We've already discussed this, and this isn't just about the Go tests. In a follow-up PR, we are going to change the event_queue ID column to be a UUID that will of course be populated by the Go code, and was planned to use  `binary(16)` for that, which has become a problem since the other columns are now `char(36)`. So, we first decided to change the existing columns to `binary(16)` as well, until you objected to that with https://github.com/Icinga/icinga-notifications/pull/496#issuecomment-5342560481 again.

julianbrost:
> Even if the Go part doesn't need to care about them, it should still have been added the corresponding fields to the Go structs, instead of just pretending they don't exist. This now forces us to either introduce a surrogate struct that has the fields we need, so we can use them to dynamically build the query and insert something into these tables, which I don't want to do, or do the work and add the fields to the Go structs with a proper type, which we should have done a long time ago.

Removing the `not null` constraint would also be an option. That's the thing that breaks the "Go doesn't need to care" assumption.

> In a follow-up PR, we are going to change the event_queue ID column to be a UUID that will of course be populated by the Go code, and was planned to use `binary(16)` for that, which has become a problem since the other columns are now `char(36)`. So, we first decided to change the existing columns to `binary(16)` as well, until you objected to that with [#496 (comment)](https://github.com/Icinga/icinga-notifications/pull/496#issuecomment-5342560481) again.

Please don't use two different ways of storing UUIDs within the same schema. That sounds like the worse option compared to sticking with `char` for MySQL. So I don't know the discussions you had before, are there strong reasons against using `char`? If so, that could be a reason to change the existing columns as well. In the end, probably someone has to bite the bullet.

yhabteab:
> Please don't use two different ways of storing UUIDs within the same schema. That sounds like the worse option compared to sticking with `char` for MySQL.

Yes, I second that. That's why this PR was opened in the first place :).

> So I don't know the discussions you had before, are there strong reasons against using `char`?

Not really, we just need to add a type that supports both text and binary encodings when persisting UUIDs. The existing types.UUID supports only binary encoding, which causes the MySQL tests to crash fatally while PostgreSQL tests run just fine. So, Jan is working on [extending the types.UUID type](https://github.com/Icinga/icinga-go-library/pull/235) to support both encodings, which we will then make use of that for these and all other UUID columns introduced in the future in Icinga Notifications while Icinga for Kubernetes uses that type with binary encoding for all its MySQL UUID columns.

julianbrost:
> which we will then make use of that for these and all other UUID columns introduced in the future in Icinga Notifications while Icinga for Kubernetes uses that type with binary encoding for all its MySQL UUID columns.

Well, so we already have an inconsistency there in how UUIDs are stored? Wouldn't it also be more convenient for the different Web modules if this was consistent? So I'd actually suggest to reconsider what's still in the PR title:

> Store external_uuid as binary(16) in MySQL schema

With the reasons for being:

1. Seems to be what's currently already implemented in both icinga-go-library and icinga-kubernetes-web. Either those would need to be changed, or there's an inconsistency.
2. There is only a single valid representation of every UUID. With `char`, you can store inconsistent representations, for example lower or upper-case hex digits, which doesn't interact that well with a `unique` constraint. That could probably be addressed with database constraints, but that's more complexity than just switching to `binary`.
3. It's the more compact and thus slightly more efficient option.

The only reasons against this seem to be the:

1. icinga-notifications-web needs some refactoring.
2. `binary` is a bit annoying as you need to remember that you probably want to invoke the CLI client with `--binary-as-hex`, but that's nothing new that would come from these columns as the object IDs here and also in Icinga DB already use `binary`.

yhabteab:
> The only reasons against this seem to be the:
> 
>     1. icinga-notifications-web needs some refactoring.

That's the only reason against using binary encoding. I don't know yet where these columns are to be used as a filter but Johannes' objection is that Icinga Notifications Web would've to introduce a new behavior in IPL orm that transforms the binary data into UUID string before it can be used in a filter. AFAIK, these columns are not used in any filter context yet, but Johannes can correct me if I'm wrong.

nilmerg:
I'm still baffled why this topic got so controversial. To me it boils down to this:

* Notifications went PostgreSQL first and uses it's UUID type as a result:
  > A UUID is written as a sequence of lower-case hexadecimal digits, in several groups separated by hyphens, specifically a group of 8 digits followed by three groups of 4 digits followed by a group of 12 digits, for a total of 32 digits representing the 128 bits (https://www.postgresql.org/docs/current/datatype-uuid.html)

  That's 36 chars. When MySQL support was introduced, we naturally chose `char(36)`.
* Kubernetes went MySQL first and has [no PostgreSQL support to this day](https://github.com/Icinga/icinga-kubernetes/issues/191). They chose `binary(16)` as a result and probably will also for PostgreSQL (without using the native type).
* The *icinga-go-library* provides a type that is used by Kubernetes exclusively up until this point and is now extended with support for a string capability in response to my argument in my first comment here.

I don't want to be forced to change anything in Notifications Web due to this. It's just not worth any effort in my eyes. The decision we made for Notifications at the time was right and is still. As a consequence, since we all seem to agree that different approaches in the same schema are even worse, `event_queue.id` should be `UUID` for PostgreSQL and `char(36)` for MySQL.

yhabteab:
> That's 36 chars. When MySQL support was introduced, we naturally chose `char(36)`.

As we already discussed last week, I don't really care about those types anymore but it was not _naturally_ to choose `char(36)` for MySQL just because [PostgreSQL's native type input and output format](https://github.com/postgres/postgres/blob/0fd1609afe255b275c2560db6db584081a16e8a0/src/backend/utils/adt/uuid.c#L94-L139) was documented as such. See for example [this MySQL blog](https://dev.mysql.com/blog-archive/mysql-8-0-uuid-support/). Also, MySQL support wasn't added after these columns were introduced but long before that https://github.com/Icinga/icinga-notifications/pull/216.

lippserd:
Don't use `char(36)`. Use `binary(16)` and a proper behavior that calls `UUID_TO_BIN()` and `BIN_TO_UUID()` for MySQL for example. With `utf8mb4`, a `char(36)` column has a maximum of up to 144 bytes of character storage allocation in principle and as @julianbrost already mentioned, hex letters can be upper- or lowercase. Both can be addressed by specifying the column's character set and collation, but it would still be 20 bytes larger. We already use `binary(16)` in Icinga for Kubernetes. It does not make sense to introduce yet another variant. For PostgreSQL, Icinga for Kubernetes would use PostgreSQL's builtin UUID type.

nilmerg:
Oh, I see why this is this controversial as I perceive it. My goal is to keep the representation normalized, i.e. as string in any case. Your goal is to use the most efficient storage for the specific vendor. Both are incompatible to each other and there's no compromise I see.

I'm just not convinced that it is justified to favor something because of a few bytes saved, instead of the other thing that allows us to *assume* a standardized default and avoid architectural conditions on a high level. We did the latter already in the past with `\ipl\Orm\Behavior\Binary`, but for a reason outside our control: PDO/pgsql uses resources for binary blobs.

Now we would do this because with MySQL we'd have to apply mentioned functions. I thought we were fed up with [this](https://github.com/search?q=repo%3AIcinga%2Ficingaweb2-module-monitoring+UNIX_TIMESTAMP&type=code).

But who am I to decide…

lippserd:
> Oh, I see why this is this controversial as I perceive it. My goal is to keep the representation normalized, i.e. as string in any case. Your goal is to use the most efficient storage for the specific vendor. Both are incompatible to each other and there's no compromise I see.

Keeping a normalized, ecosystem-wide UUID representation is a valid goal. However, that representation does not need to be the physical storage format. The compromise is to use canonical UUIDs in our code while using the appropriate physical storage format for each database.

> I'm just not convinced that it is justified to favor something because of a few bytes saved...

The difference is more significant than it first appears:

For every UUID column, the larger representation uses more disk space and increases backup and replication/binlog volume. If the UUID column is indexed, larger index entries also mean fewer entries per page, more index pages, and potentially more memory use and I/O.

If the UUID is the primary key, the effect is larger with InnoDB because every secondary index record includes the table's primary key columns.

All of these effects grow with row count.

> ...instead of the other thing that allows us to _assume_ a standardized default and avoid architectural conditions on a high level.

Again, I agree that we should have one UUID representation across our ecosystem. However, requiring `char(36)` in MySQL solely for uniformity would create recurring storage, bandwidth, and indexing costs without improving the code.

nilmerg:
> For every UUID column, the larger representation uses more disk space and increases backup and replication/binlog volume. If the UUID column is indexed, larger index entries also mean fewer entries per page, more index pages, and potentially more memory use and I/O.

For every additional piece of code, the more CPU resources are needed and increases energy consumption and dissipated heat. The implementation responsible for this must be maintained and increases burden in the foreseeable future.

Yeah, *storage is cheap* didn't age well, but what led to climate change was energy consumption. I know I exaggerate *heavily* here, but you came up with this.

--

Didn't I say there's no compromise to make? It's either this, or that. Put simply.

And if I'm not convinced, so be it. I don't want to further fuel the controversy as I'm obviously alone with my standpoint. But I clearly don't support the decision.

yhabteab (dismissed):

yhabteab (approved):

PR #498, opened 2026-08-26
Add a channel for Telegram

### LINKED ISSUE #503: Add Telegram channel
Add a new channel plugin for Telegram integration.

DsgnrFH:
Added a channel for Telegram
Right now it sends notifications in this format:

IMG_4098

, which of course can be changed to overcome some restrictions (Telegram dodes not allow embedded links to be localhost, etc.)

### DIFF
--- a/cmd/channels/telegram/main.go
+++ b/cmd/channels/telegram/main.go
@@ -1,0 +1,251 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"crypto/sha256"
+	"encoding/binary"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"log/slog"
+	"net/http"
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/icinga/icinga-go-library/notifications"
+	"github.com/icinga/icinga-go-library/notifications/jsonrpc"
+	"github.com/icinga/icinga-go-library/notifications/plugin"
+	"github.com/icinga/icinga-go-library/utils"
+	"github.com/icinga/icinga-notifications/internal"
+)
+
+func main() {
+	plugin.Run(&Telegram{})
+}
+
+type Telegram struct {
+	BotToken string `json:"bot_token"`
+
+	mu sync.Mutex
+
+	rpcCtx context.Context
+	rpcEp  *jsonrpc.Endpoint
+}
+
+// state represents the state of the Telegram channel, holding the last message sent to a chat for this incident.
+type state struct {
+	MessageID int64 `json:"message_id"`
+}
+
+func (s state) String() string {
+	b, err := json.Marshal(s)
+	if err != nil {
+		panic(err)
+	}
+	return string(b)
+}
+
+// replyParameters describes the message a sendMessage request replies to.
+// API Doc: https://core.telegram.org/bots/api#replyparameters
+type replyParameters struct {
+	MessageID int64 `json:"message_id"`
+
+	// AllowSendingWithoutReply lets Telegram send the message even if the replied to message is gone
+	// (if original message was deleted).
+	AllowSendingWithoutReply bool `json:"allow_sending_without_reply"`
+}
+
+// ReceiveEndpoint implements the [plugin.RPCEndpointReceiver] interface.
+func (ch *Telegram) ReceiveEndpoint(ctx context.Context, ep *jsonrpc.Endpoint) {
+	ch.rpcCtx = ctx
+	ch.rpcEp = ep
+}
+
+func (ch *Telegram) GetInfo() *plugin.Info {
+	configAttrs := plugin.ConfigOptions{
+		{
+			Name: "bot_token",
+			Type: "secret",
+			Label: map[string]string{
+				"en_US": "Bot token",
+				"de_DE": "Bot Token",
+			},
+			Help: map[string]string{
+				"en_US": "Telegram bot API token from BotFather",
+				"de_DE": "Telegram Bot API Token von BotFather",
+			},
+			Required: true,
+		},
+	}
+
+	return &plugin.Info{
+		Name:             "Telegram",
+		Version:          internal.Version.Version,
+		Author:           "Icinga GmbH",
+		ConfigAttributes: configAttrs,
+	}
+}
+
+func (ch *Telegram) SetConfig(jsonStr json.RawMessage) error {
+	var tmp Telegram
+	if err := json.Unmarshal(jsonStr, &tmp); err != nil {
+		return fmt.Errorf("could not unmarshal configuration: %w", err)
+	}
+
+	ch.mu.Lock()
+	defer ch.mu.Unlock()
+	ch.BotToken = tmp.BotToken
+
+	return nil
+}
+
+func (ch *Telegram) SendNotification(req *plugin.NotificationRequest) error {
+	var chatID string
+	for _, address := range req.Contact.Addresses {
+		if address.Type == "telegram" {
+			chatID = address.Address
+			break
+		}
+	}
+	if chatID == "" {
+		return fmt.Errorf("contact %q has no Telegram address", req.Contact.FullName)
+	}
+
+	var output bytes.Buffer
+	_, _ = fmt.Fprint(&output, plugin.FormatSubject(req)+"\n\n")
+	plugin.FormatMessage(&output, req)
+
+	message := struct {
+		ChatID                string           `json:"chat_id"`
+		Text                  string           `json:"text"`
+		DisableWebPagePreview bool             `json:"disable_web_page_preview"`
+		ReplyParameters       *replyParameters `json:"reply_parameters,omitempty"`
+	}{
+		ChatID: chatID,
+		// Telegram limits sendMessage "text" field to 4096 characters.
+		// https://core.telegram.org/bots/api#sendmessage
+		Text:                  utils.Ellipsize(output.String(), 4096),
+		DisableWebPagePreview: true,
+	}
+
+	// When this chat has already been notified about this incident, reply to the last message sent for it.
+	stateKey := makeStateKey(chatID, req.Incident)
+	var knownThread bool
+
+	for _, s := range req.States {
+		if stateKey == "" || s.Key != stateKey {
+			continue
+		}
+
+		var st state
+		if err := json.Unmarshal([]byte(s.Value), &st); err != nil {
+			return fmt.Errorf("cannot unmarshal channel state %q: %w", s.Key, err)
+		}
+		knownThread = true
+		message.ReplyParameters = &replyParameters{MessageID: st.MessageID, AllowSendingWithoutReply: true}
+		break
+	}
+
+	messageID, err := ch.sendMessage(message)
+	if err != nil {
+		return err
+	}
+
+	if req.Incident == nil || stateKey == "" {
+		return nil
+	}
+
+	if req.Incident.IsRecovered && knownThread {
+		// The thread is done, no further notifications will reference it.
+		states := []plugin.State{{Key: stateKey}}
+		if err := ch.rpcEp.Call(ch.rpcCtx, notifications.MethodDeleteState, states, nil); err != nil {
+			slog.ErrorContext(ch.rpcCtx, "Failed to delete channel state", "error", err)
+		}
+	} else if !req.Incident.IsRecovered {
+		// Remember this message so that the next notification for this incident replies to it.
+		states := []plugin.State{{Key: stateKey, Value: state{MessageID: messageID}.String()}}
+		if err := ch.rpcEp.Call(ch.rpcCtx, notifications.MethodUpsertState, states, nil); err != nil {
+			slog.ErrorContext(ch.rpcCtx, "Failed to upsert channel state", "error", err)
+		}
+	}
+
+	return nil
+}
+
+// sendMessage posts the given message to the Telegram bot API and returns the message id.
+func (ch *Telegram) sendMessage(message any) (int64, error) {
+	body, err := json.Marshal(message)
+	if err != nil {
+		return 0, err
+	}
+
+	ch.mu.Lock()
+	token := ch.BotToken
+	ch.mu.Unlock()
+
+	// The Telegram bot API carries a bot token in the request path,
+	// so every request is of the form
+	// https://api.telegram.org/bot<token>/method.
+	// API Doc: https://core.telegram.org/bots/api#making-requests
+	url := fmt.Sprintf("https://api.telegram.org/bot%s/sendMessage", token)
+	client := &http.Client{Timeout: 10 * time.Second}
+	//nolint:bodyclose // False positive, drainAndClose is called in the defer statement below.
+	resp, err := client.Post(url, "application/json", bytes.NewReader(body))
+	if err != nil {
+		return 0, fmt.Errorf("error while sending http request to Telegram: %w", redactToken(err, token))
+	}
+	defer drainAndClose(resp.Body)
+
+	var response struct {
+		Ok          bool   `json:"ok"`
+		Description string `json:"description,omitempty"`
+		Result      struct {
+			MessageID int64 `json:"message_id"`
+		} `json:"result"`
+	}
+	if err := json.NewDecoder(resp.Body).Decode(&response); err != nil {
+		return 0, fmt.Errorf("cannot decode Telegram response for HTTP status %q: %w", resp.Status, err)
+	}
+	if resp.StatusCode != http.StatusOK || !response.Ok {
+		return 0, fmt.Errorf("Telegram rejected the message with HTTP status %q and description %q",
+			resp.Status, response.Description)
+	}
+
+	return response.Result.MessageID, nil
+}
+
+// redactToken removes the bot token from the request URL an url.Error carries, as this error would leak
+// the token into the daemons log:
+// https://github.com/golang/go/issues/44819
+func redactToken(err error, token string) error {
+	var urlErr *url.Error
+	if token != "" && errors.As(err, &urlErr) {
+		urlErr.URL = strings.ReplaceAll(urlErr.URL, token, "<redacted>")
+	}
+	return err
+}
+
+// makeStateKey composes a unique key for the channel state based on the recipients chat ID and the incident ID.
+func makeStateKey(chatID string, i *plugin.Incident) string {
+	if i == nil {
+		return ""
+	}
+	h := sha256.New()
+	if err := binary.Write(h, binary.BigEndian, i.Id); err != nil {
+		return ""
+	}
+	if err := binary.Write(h, binary.BigEndian, []byte(chatID)); err != nil {
+		return ""
+	}
+	return hex.EncodeToString(h.Sum(nil))
+}
+
+func drainAndClose(r io.ReadCloser) {
+	_, _ = io.Copy(io.Discard, r)
+	_ = r.Close()
+}
--- a/doc/01-About.md
+++ b/doc/01-About.md
@@ -35,6 +35,7 @@ Icinga Notifications comes with multiple channels out of the box:
 
 * _email_: Email submission via SMTP
 * _rocketchat_: Rocket.Chat
+* _telegram_: [Telegram](10-Channels.md#telegram)
 * _webhook_: Configurable HTTP/HTTPS queries for your backend
 
 Additional custom channels can be developed independently of Icinga Notifications,
--- a/doc/10-Channels.md
+++ b/doc/10-Channels.md
@@ -13,6 +13,46 @@ and stores those options in the database.
 Using this information, Icinga Notifications Web allows channels to be configured,
 which are then started, configured, and finally used to send notification events from Icinga Notifications.
 
+## Available Channels
+
+Icinga Notifications comes with multiple channels out of the box:
+
+* _email_: Email submission via SMTP
+* _rocketchat_: Rocket.Chat
+* _telegram_: Telegram
+* _webhook_: Configurable HTTP/HTTPS queries for your backend
+
+Most of them are configured entirely in Icinga Notifications Web.
+The sections below cover those channels that need some preparation on the other side first.
+
+### Telegram
+
+Telegram messages are sent by a bot, which is created and owned by a regular Telegram user account.
+Each contact receives their notifications in their personal chat from this bot.
+
+1. Create a bot by talking to [@BotFather](https://t.me/botfather) and using its `/newbot` command, as described in
+the [Telegram bot documentation](https://core.telegram.org/bots/features#creating-a-new-bot).
+BotFather answers with an authentication token, which is the credential for the bot's whole API access.
+Treat it like a password.
+
+2. Create a new channel of the type _Telegram_ in Icinga Notifications Web and set the _Bot token_ to that token.
+
+3. Let every contact who should be notified open a chat with the bot and send it a message, for example `/start`.
+A bot is not allowed to write messages to someone who has never written to it, so this step can not be skipped.
+
+4. Look up the chat ID of each contact and configure it as their _telegram_ address in Icinga Notifications Web.
+
+The messages from the previous step are queued as updates for the bot and can be read with the
+[`getUpdates`](https://core.telegram.org/bots/api#getupdates) method:
+
+```
+curl -s 'https://api.telegram.org/bot<token>/getUpdates'
+```
+
+In each update, `message.chat.id` is the chat ID to configure.
+
+A group can be notified the same way: add the bot to the group, post a message there, and use the group's chat ID.
+
 ## Technical Channel Description
 
 Channel plugins are independent processes that run continuously, started and supervised by Icinga Notifications.
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,6 +4,7 @@ Johannes Meyer <johannes.meyer@icinga.com>
 Johannes Rauh <johannes.rauh@icinga.com>
 Julian Brost <julian.brost@icinga.com>
 Noé Costa <noe.costa@icinga.com>
+Oleksandr Barbashyn <oleksandr.barbashyn@icinga.com>
 Ravi Kumar Kempapura Srinivasa <ravi.srinivasa@icinga.com>
 Sukhwinder Dhillon <sukhwinder.dhillon@icinga.com>
 Yonas Habteab <yonas.habteab@icinga.com>


julianbrost:
Please ask someone who's currently actually working on this project for a review instead (ideally after taking a look at the CI failure).

DsgnrFH:
False positive, drainAndClose is called in the defer statement below (line 157)

DsgnrFH:
> False positive, drainAndClose is called in the defer statement below (line 157)

Still checked it and logically it would be smarter and better to add the defer right after response body was opened

yhabteab:
> > False positive, drainAndClose is called in the defer statement below (line 157)
> 
> Still checked it and logically it would be smarter and better to add the defer right after response body was opened

Posting such comments here won't help. You've to do it exactly like how it's done for the Rocket.Chat implementation to fix the action.

DsgnrFH:
> > > False positive, drainAndClose is called in the defer statement below (line 157)
> > 
> > 
> > Still checked it and logically it would be smarter and better to add the defer right after response body was opened
> 
> Posting such comments here won't help. You've to do it exactly like how it's done for the Rocket.Chat implementation to fix the action.

Okay, I added the comment and squished all the commits from before

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,172 @@
...
+		return errors.New(resp.Status)
+	}
+
+	return nil
+}
+
+
+func main() {
+	plugin.Run(&Telegram{})
+}
```
yhabteab: Please move this up to the top of the file but still after the imports. See the other plugins for example.

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,172 @@
...
+	body, err := json.Marshal(message)
+	if err != nil {
+		return err
+	}
+
+	t.mu.Lock()
+	token := t.BotToken
+	t.mu.Unlock()
+
+	url := fmt.Sprintf("https://api.telegram.org/bot%s/sendMessage", token)
```
yhabteab, line 195: Is this base URL always same for all users out there? If not please make this also user configurable. Also, please add some comments and reference on why you're putting the token into the request URL instead of using a basic Auth header. If that's required by the Telegram API, add a reference to it in the comment.

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,172 @@
...
+	if req.Incident != nil {
+		sev := req.Incident.Severity.String()
+		circle := severityCircle[sev]
+		fmt.Fprintf(&b, "%s <b>%s</b>  |  #%d\n", circle, html.EscapeString(strings.ToUpper(sev)), req.Incident.Id)
+	}
+
+	fmt.Fprintf(&b, "<b>%s</b>\n", html.EscapeString(req.Object.Name))
+
+	if msg := strings.TrimSpace(req.Event.Message); msg != "" {
+		fmt.Fprintf(&b, "\n%s\n", html.EscapeString(truncate(msg, messageLimit-600)))
```
yhabteab: You can replace `truncate` with [`utils.Ellipsize()`](https://github.com/Icinga/icinga-go-library/blob/main/utils/utils.go#L111-L121) and remove that function.  Also, why are you subtracting `600` from the limit instead of just using it as-is?

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,172 @@
...
+func formatText(req *plugin.NotificationRequest) string {
+	var b strings.Builder
+
+	if req.Incident != nil {
+		sev := req.Incident.Severity.String()
+		circle := severityCircle[sev]
+		fmt.Fprintf(&b, "%s <b>%s</b>  |  #%d\n", circle, html.EscapeString(strings.ToUpper(sev)), req.Incident.Id)
+	}
+
+	fmt.Fprintf(&b, "<b>%s</b>\n", html.EscapeString(req.Object.Name))
```
yhabteab: Please just use `plugin.FormatSubject` for now. Using custom formatting for this plugin might be desirable but the messages look way too fishy now. So, you don't need to waste time on this now without deciding whether we want to use a custom format for this at all, which probably can only be answered by someone else.

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,172 @@
...
+	"emerg": "⚫️",
+}
+
+type Telegram struct {
+	BotToken string `json:"bot_token"`
+
+	mu sync.Mutex
+}
+
+func formatText(req *plugin.NotificationRequest) string {
```
yhabteab: Move this and all the other helper function to the end of the file please.

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,172 @@
...
+	"net/http"
+	"strings"
+	"sync"
+	"time"
+	"io"
+
+	"github.com/icinga/icinga-go-library/notifications/plugin"
+)
+
+const messageLimit = 4000
```
yhabteab: Please document why this is needed. Also, why are you setting this to 4000 instead of 4096. After, some googling on my onw (because there's no reference to it here), I've found this [API doc](https://core.telegram.org/bots/api#sendmessage) and states the limit for the `text` field is 4096.

yhabteab (changes requested):

cmd/channels/telegram/main.go
DsgnrFH: IT is constant and is the same for all users. Telegram API requires exactly this, I also don't find it the best approach on Telegram's side.
I will add a comment about it

cmd/channels/telegram/main.go
DsgnrFH: Good suggestion, will do

cmd/channels/telegram/main.go
DsgnrFH: 600 was a safe average for leaving some headspace above the message length limit. I thought it would make it better, but it seems like Telegram doesn't count formatting as characters, so I will replace it

cmd/channels/telegram/main.go
DsgnrFH: Fair enough, I will replicate the Rocketchat implementation. Should we at least leave the circles? the green/yellow/red ones at the very start, as seen on the screenshot.

yhabteab: No, you don't have to replicate the Rocket.Chat implementation. If Telegram allows to send prettified text, then you can do it but the used circles are, for one, incomplete and does not include all the available severity levels, and for two, those are some random colors and does not align with our state balls in Icinga Notifications Web. If in doubt, then the circles should use the same coloring as in Icinga Notifications Web (in the incidents view).

cmd/channels/telegram/main.go
DsgnrFH: Will adjust this too

cmd/channels/telegram/main.go
DsgnrFH: Exactly this function is no longer needed as we use FormatSubject now.

DsgnrFH: But I did move every helper function to the end of the files

yhabteab: > Exactly this function is no longer needed as we use FormatSubject now.

No, `FormatSubject` formats only the subject not the text or the body.

cmd/channels/telegram/main.go
yhabteab: Still no reference to the API doc here.

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,128 @@
...
+	t.mu.Unlock()
+
+	url := fmt.Sprintf("https://api.telegram.org/bot%s/sendMessage", token)
+	client := &http.Client{Timeout: 10 * time.Second}
+	//nolint:bodyclose // False positive, drainAndClose is called in the defer statement below.
+	resp, err := client.Post(url, "application/json", bytes.NewReader(body))
+	defer drainAndClose(resp.Body)
+
+	if err != nil {
+		return fmt.Errorf("error while sending http request to telegram: %v", err)
```
yhabteab: Errors are wrapped using the `%w` verb not `%v`. You can check the difference by yourself :).

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,128 @@
...
+		Author: "Icinga GmbH",
+		ConfigAttributes: configAttrs,
+	}
+}
+
+func (t *Telegram) SetConfig(jsonStr json.RawMessage) error {
+	var tmp Telegram
+	if err := plugin.PopulateDefaults(&tmp); err != nil {
+		return err
+	}
```
yhabteab: I know, it states exactly like this in our channel specification but given that this Plugin only require a single configurable field without any defaults, I would remove this if statement entirely.

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,128 @@
...
+				"en_US": "Telegram bot API token from BotFather",
+				"de_DE": "Telegram Bot API Token von BotFather",
+			},
+			Required: true,
+		},
+	}
+
+	return &plugin.Info{
+		Name: "Telegram",
+		Version: "0.0.1",
```
yhabteab: Please use the same version as in the other channel plugins.

yhabteab (changes requested):
Also, apart from the below inline comments, have you considered adding threading support for the plugin, so that every notification related to the same incident gets grouped together? Looking at the [API docs](https://core.telegram.org/bots/api#message), the `message_thread_id` looks exactly what we the plugin could make use of.

cmd/channels/telegram/main.go
DsgnrFH: You're right, i checked the docs. Updating the verbs

cmd/channels/telegram/main.go
DsgnrFH: Makes sence after moving from testing

cmd/channels/telegram/main.go
DsgnrFH: I updated it. Also I noticed tabs and adjusted them too

cmd/channels/telegram/main.go
DsgnrFH: I added a short comment about it. Didn't find a proper API doc for the url except for this part:
https://core.telegram.org/bots/api#authorizing-your-bot

yhabteab: Yes. That's what I was missing. Please add this URl https://core.telegram.org/bots/api#making-requests to your newly added comment.

cmd/channels/telegram/main.go
yhabteab: > Also I noticed tabs and adjusted them too

Ehmm. Wait, why can't your IDE do this automatically? Normally, when writing Go code you shouldn't care about things like this as `go fmt` or ideally your IDE should just do this atomatically. Though, you don't seem to use any of those as the imported standard libs still aren't sorted correctly.

yhabteab:
> have you considered adding threading support for the plugin, so that every notification related to the same incident gets grouped together? Looking at the [API docs](https://core.telegram.org/bots/api#message), the `message_thread_id` looks exactly what we the plugin could make use of.

No reaction to this?

cmd/channels/telegram/main.go
DsgnrFH: I am using neovim because I am used to it in my personal projects. I will try GoLand though

DsgnrFH:
> > have you considered adding threading support for the plugin, so that every notification related to the same incident gets grouped together? Looking at the [API docs](https://core.telegram.org/bots/api#message), the `message_thread_id` looks exactly what we the plugin could make use of.
> 
> No reaction to this?

I couldn't reply to it, I have considered it, and I agree that message_thread_id would be a good way to group notifications for the same incident. The main thing that needs to be solved first is persisting a mapping between an Icinga incident and the corresponding Telegram thread ID. The current plugin is stateless, so simply adding message_thread_id to sendMessage wouldn't be sufficient.
I also didn't see a use for threading in Telegram, so I would have to research and test it a bit too

yhabteab:
> The current plugin is stateless, so simply adding message_thread_id to sendMessage wouldn't be sufficient.

Well, if we want to do this, then we have to make it stateful (see the mail channel implementation for inspiration).

> I also didn't see a use for threading in Telegram, so I would have to research and test it a bit too

Please do this. It would be nice to have this support. Because if the same recipient is notified about different incidents, it would be helpful if those notifications of a specific incident could be grouped together. This way, the recipient can easily see all related notifications in one place, rather than searching for multiple separate messages.

DsgnrFH:
Update:
After discussing this topic with @yhabteab, we have decided to start with saving chat IDs and Incidents as keys to their original messages in the notifications db similar to how [email channel](https://github.com/Icinga/icinga-notifications/blob/main/cmd/channels/email/main.go) does it.
So instead of just sending a message we now:
-  Chekc if this chatID already has this IncidentID, and if so, then take its original first message ID and reply to it.
-  If it does not, then we send the message as is, and get its ID from telegram, then save it with current ChatID and IcncidentID so we find it later.
- When the incident is resolved, we reply to that original message (first option) and then delete that pair (ChatID:IncidentID + MessageID)

So we save the ID of the first original message that was the first to tell the user about that particular Incident.
@yhabteab please correct me if I'm wrong

yhabteab:
Yes, that's correct. However, as also discussed, there should be a proof concept first, so that we can compare whether sending 10+ notifications as replies is better than the current state. As to the message threading, it turns out, that it's kinda complicated and doesn't suit to be used here by the plugin, so that idea is dropped entirely now.

DsgnrFH:
Well, another idea would be to update one same message, but also writing history in itself, but who knows how long one's incident might get, it could be too long to read and/or exceed the 4096 characters limit. I would say we start with replies

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,250 @@
...
+// API Doc: https://core.telegram.org/bots/api#replyparameters
+type replyParameters struct {
+	MessageID int64 `json:"message_id"`
+
+	// AllowSendingWithoutReply lets Telegram send the message even if the replied to message is gone (if original message was deleted)
+	AllowSendingWithoutReply bool `json:"allow_sending_without_reply"`
+}
+
+// ReceiveEndpoint implements the [plugin.RPCEndpointReceiver] interface.
+func (t *Telegram) ReceiveEndpoint(ctx context.Context, ep *jsonrpc.Endpoint) {
```
oxzi: Please name the struct receiver variable `ch` (for channel) as it is named in every other channel plugin implementation. For example:

https://github.com/Icinga/icinga-notifications/blob/63a6b434eab169d11b7f543b3790da6aef4c4a4b/cmd/channels/webhook/main.go#L36

https://github.com/Icinga/icinga-notifications/blob/63a6b434eab169d11b7f543b3790da6aef4c4a4b/cmd/channels/email/main.go#L75

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+	return string(b)
+}
+
+// replyParameters describes the message a sendMessage request replies to.
+// API Doc: https://core.telegram.org/bots/api#replyparameters
+type replyParameters struct {
+	MessageID int64 `json:"message_id"`
+
+	// AllowSendingWithoutReply lets Telegram send the message even if the replied to message is gone (if original message was deleted)
```
oxzi: Please keep the line length below 120 chars.

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,250 @@
...
+func (t *Telegram) SendNotification(req *plugin.NotificationRequest) error {
+	var chatID string
+	for _, address := range req.Contact.Addresses {
+		if address.Type == "telegram" {
+			chatID = address.Address
+			break
+		}
+	}
+	if chatID == "" {
+		return fmt.Errorf("contact %s has no telegram address", req.Contact.FullName)
```
oxzi:
```suggestion
		return fmt.Errorf("contact %q has no telegram address", req.Contact.FullName)
```

Please always prefer the escaped variant if you are dealing with user input.

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,250 @@
...
+		Text                  string           `json:"text"`
+		DisableWebPagePreview bool             `json:"disable_web_page_preview"`
+		ReplyParameters       *replyParameters `json:"reply_parameters,omitempty"`
+	}{
+		ChatID:                chatID,
+		Text:                  utils.Ellipsize(output.String(), messageLimit),
+		DisableWebPagePreview: true,
+	}
+
+	// When this chat has already been notified about this incident, reply to the message that thread started withj
```
oxzi: > […] started withj

I guess you have forgot something here :)

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,250 @@
...
+	h := sha256.New()
+	if err := binary.Write(h, binary.BigEndian, i.Id); err != nil {
+		return ""
+	}
+	if err := binary.Write(h, binary.BigEndian, []byte(chatID)); err != nil {
+		return ""
+	}
+
+	return hex.EncodeToString(h.Sum(nil))
+}
```
oxzi, line 246: The `makeStateKey` implementation is almost identical to those in the `email` plugin, but has a few empty lines lacking over there. Please remove those to keep the implementation looking consistent.

https://github.com/Icinga/icinga-notifications/blob/63a6b434eab169d11b7f543b3790da6aef4c4a4b/cmd/channels/email/main.go#L291-L303

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,250 @@
...
+			slog.ErrorContext(t.rpcCtx, "Failed to delete channel state", "error", err)
+		}
+
+	case !knownThread:
+		// This message started the thread for this incident, remember it for all upcoming notifications
+		states := []plugin.State{{Key: stateKey, Value: state{MessageID: messageID}.String()}}
+		if err := t.rpcEp.Call(t.rpcCtx, notifications.MethodUpsertState, states, nil); err != nil {
+			slog.ErrorContext(t.rpcCtx, "Failed to upsert channel state", "error", err)
+		}
+	}
```
oxzi: The conditions for this `switch` are a bit confusing, especially as the first `case` follows with an additional `if` check which aborts. How about:

```
if req.Incident.IsRecovered && knownThread {
  // first part: deletion
} else if !knownThread {
  // second part: upsert
}
```

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,250 @@
...
+	client := &http.Client{Timeout: 10 * time.Second}
+	//nolint:bodyclose // False positive, drainAndClose is called in the defer statement below.
+	resp, err := client.Post(url, "application/json", bytes.NewReader(body))
+	if err != nil {
+		return 0, fmt.Errorf("error while sending http request to telegram: %w", err)
+	}
+	defer drainAndClose(resp.Body)
+
+	if resp.StatusCode != http.StatusOK {
+		return 0, errors.New(resp.Status)
```
oxzi: This would result in an HTTP status method being returned as the lone error back to the user. Please add some information by creating an error explaining what's wrong, and including the status.

```
fmt.Errorf("unexpected HTTP status code %q from Telegram API", resp.Status)
```

Furthermore, when just aborting after an unexpected status code, you will skip the error description I described below: https://github.com/Icinga/icinga-notifications/pull/498/changes#r3879716048. So, please move this check down and merge with `if !response.Ok`.

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	// API Doc: https://core.telegram.org/bots/api#making-requests
```
oxzi: This comment adds no value as you linked the same URL already ten lines above.

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,250 @@
...
+		Ok     bool `json:"ok"`
+		Result struct {
+			MessageID int64 `json:"message_id"`
+		} `json:"result"`
+	}
+	if err := json.NewDecoder(resp.Body).Decode(&response); err != nil {
+		return 0, fmt.Errorf("cannot decode telegram response: %w", err)
+	}
+	if !response.Ok {
+		return 0, errors.New("telegram rejected the message")
```
oxzi: > The response contains a JSON object, which always has a Boolean field 'ok' and may have an optional String field 'description' with a human-readable description of the result. If 'ok' equals True, the request was successful and the result of the query can be found in the 'result' field. In case of an unsuccessful request, 'ok' equals False and the error is explained in the 'description'.

https://core.telegram.org/bots/api#making-requests

---

Please add the `description` field to your `response` struct type and include its text in this error. Based on the docs, I would assume its absence in a successful case, so you can mark it as `omitempty`.

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,250 @@
...
+	"github.com/icinga/icinga-go-library/notifications"
+	"github.com/icinga/icinga-go-library/notifications/jsonrpc"
+	"github.com/icinga/icinga-go-library/notifications/plugin"
+	"github.com/icinga/icinga-go-library/utils"
+	"github.com/icinga/icinga-notifications/internal"
+)
+
+// Telegram limits sendMessage "text" field to 4096 characters
+// https://core.telegram.org/bots/api#sendmessage
+const messageLimit = 4096
```
oxzi: Having `messageLimit` as the sole const at the start of the file feels a bit out of place. Why not inline it in `SendNotification` and add the same explaining comment there?

cmd/channels/telegram/main.go
oxzi: Please write _Telegram_ with a capitalized _T_ in the errors as well, being a product name and all. This applies to this error and all following.

cmd/channels/telegram/main.go
oxzi, line 1: Please add your channel to the list of available channels in the docs.

https://github.com/Icinga/icinga-notifications/blob/63a6b434eab169d11b7f543b3790da6aef4c4a4b/doc/01-About.md?plain=1#L32-L38

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,250 @@
...
+	// So every request is of the form
+	// https://api.telegram.org/bot<token>/method
+	// API Doc: https://core.telegram.org/bots/api#making-requests
+	url := fmt.Sprintf("https://api.telegram.org/bot%s/sendMessage", token)
+	client := &http.Client{Timeout: 10 * time.Second}
+	//nolint:bodyclose // False positive, drainAndClose is called in the defer statement below.
+	resp, err := client.Post(url, "application/json", bytes.NewReader(body))
+	if err != nil {
+		return 0, fmt.Errorf("error while sending http request to telegram: %w", err)
+	}
```
oxzi, line 201: Having the sensitive token in the URL might result in it being mirrored in this error, and so logged by the Notifications daemon. See [CWE-532: Insertion of Sensitive Information into Log File](https://cwe.mitre.org/data/definitions/532.html) and <https://github.com/golang/go/issues/44819>.

Using a slightly altered version from @tmthrgd's [comment](https://github.com/golang/go/issues/44819#issuecomment-793231744), one might end up with something like: https://go.dev/play/p/km2Mwt2dY-T

cmd/channels/telegram/main.go
oxzi, line 1: This channel misses a bit of information how to start using it. I know, at the moment there is no such documentation for the other channels, but how about starting it right now. Please add a new subsection to `doc/10-Channels.md` called "Available Channels" - mirroring the subsection in `01-About.md` - and add a sub-subsection for this Telegram Channel, briefly explaining how to get it working. Therefore, you can link to the official Telegram docs and do not have to duplicate instructions already written there. Just a few simple steps to help users getting started.

cmd/channels/telegram/main.go
oxzi: Feel free to continue whatever you like. For Vim, NeoVim, and others, there are multiple plugins you can choose from: https://go.dev/gopls/editor/vim

oxzi (commented):
Thanks for this channel implementation. I went over it twice and came up with mostly smaller issuer. So please don't feel discouraged by the number of comments.

cmd/channels/telegram/main.go
DsgnrFH: I adjusted this

cmd/channels/telegram/main.go
DsgnrFH: I split the comment on two lines. I wonder if we should actually drop the comment in braces because it is already clear why allow_sending_without_reply is needed

cmd/channels/telegram/main.go
DsgnrFH: Switched to using %q verb

cmd/channels/telegram/main.go
DsgnrFH: A typo, fixed it

cmd/channels/telegram/main.go
DsgnrFH: Okay, removed em

cmd/channels/telegram/main.go
DsgnrFH: Addressed this

cmd/channels/telegram/main.go
DsgnrFH: You're right, I removed this one

cmd/channels/telegram/main.go
DsgnrFH: Done

cmd/channels/telegram/main.go
DsgnrFH: Yup, did this change

cmd/channels/telegram/main.go
DsgnrFH: Changed it up

cmd/channels/telegram/main.go
DsgnrFH: Added it

cmd/channels/telegram/main.go
DsgnrFH: Addressed it

cmd/channels/telegram/main.go
DsgnrFH: Shoud I remove the links to the Telegram docs in the code then?

DsgnrFH: I added some documentation to it

cmd/channels/telegram/main.go
oxzi: It's still there.

cmd/channels/telegram/main.go
oxzi: > Shoud I remove the links to the Telegram docs in the code then?

No, those are technical information for developers, while the documentation in `./doc/` is for users. Those are two different kinds of documentation.

doc/10-Channels.md
```diff
@@ -13,6 +13,47 @@ and stores those options in the database.
...
+
+1. Create a bot by talking to [@BotFather](https://t.me/botfather) and using its `/newbot` command, as described in
+the [Telegram bot documentation](https://core.telegram.org/bots/features#creating-a-new-bot).
+BotFather answers with an authentication token, which is the credential for the bot's whole API access.
+Treat it like a password.
+
+2. Create a new channel of the type _Telegram_ in Icinga Notifications Web and set the _Bot token_ to that token.
+
+3. Let every contact who should be notified open a chat with the bot and send it a message, for example `/start`.
+A bot is not allowed to write messages to someone who has never written to it, so thos step can not be skipped.
```
```suggestion
A bot is not allowed to write messages to someone who has never written to it, so this step can not be skipped.
```

cmd/channels/telegram/main.go
```diff
@@ -0,0 +1,254 @@
...
+	_, _ = fmt.Fprint(&output, plugin.FormatSubject(req)+"\n\n")
+	plugin.FormatMessage(&output, req)
+
+	message := struct {
+		ChatID                string           `json:"chat_id"`
+		Text                  string           `json:"text"`
+		DisableWebPagePreview bool             `json:"disable_web_page_preview"`
+		ReplyParameters       *replyParameters `json:"reply_parameters,omitempty"`
+	}{
+		ChatID:                chatID,
```
oxzi: Please `gofmt` your code. After saving the file, it looked like this on my machine.

```diff
diff --git a/cmd/channels/telegram/main.go b/cmd/channels/telegram/main.go
index 6a7568f..ef98179 100644
--- a/cmd/channels/telegram/main.go
+++ b/cmd/channels/telegram/main.go
@@ -126,7 +126,7 @@ func (ch *Telegram) SendNotification(req *plugin.NotificationRequest) error {
                DisableWebPagePreview bool             `json:"disable_web_page_preview"`
                ReplyParameters       *replyParameters `json:"reply_parameters,omitempty"`
        }{
-               ChatID:                chatID,
+               ChatID: chatID,
                // Telegram limits sendMessage "text" field to 4096 characters
                // https://core.telegram.org/bots/api#sendmessage
                Text:                  utils.Ellipsize(output.String(), 4096),
```

doc/10-Channels.md
```diff
@@ -13,6 +13,47 @@ and stores those options in the database.
...
+[`getUpdates`](https://core.telegram.org/bots/api#getupdates) method:
+
+```
+curl -s 'https://api.telegram.org/bot<token>/getUpdates'
+```
+
+In each update, `message.chat.id` is the chat ID to configure.
+
+A group can be notified the same way: add the bot to the group, post a message there, and use the group's chat ID.
+The only difference is, for the group chats to work, the bot's owner has to disable the privacy mode in its settings.
```
oxzi: Pardon, I am a bit out of touch with Telegram since I haven't used it in ages. So, I tried to look up the group privacy mode and found the official ["What is the group privacy mode of Telegram bots? And how to turn it off?" article](https://www.teleme.io/articles/group_privacy_mode_of_telegram_bots?hl=en) quite useful. However, reading this article, I don't think this very bot needs to have a disabled privacy mode, as it only needs to send messages and do not need to receive those. Am I right or did I missed something?

cmd/channels/telegram/main.go
oxzi: Was added, I am going to resolve this comment.

oxzi (commented):
Thanks. Please address those last things and squash your commits to one.

cmd/channels/telegram/main.go
DsgnrFH: Okay, I removed the comments block around lines 204

doc/10-Channels.md
DsgnrFH: Thanks, fixed the typo

cmd/channels/telegram/main.go
DsgnrFH: I gofmt'ed the file

doc/10-Channels.md
DsgnrFH: Well in order to get the group ChatID the bot would need to receive a message from the group. With the privacy setting on (default), it only receives messages that are meant to be recieved by the bot, so like this: `/start@icinga-notif-bot`, and in my own experience turning off the privacy mode was proven to save a few seconds by not having to type extra stuff and just having the bot react to any message, be it `gnjurienbe` or `abc123`. Telegram could have changed this in the time though, so It looks like it isn't needed anymore. And yes, we don't need it because we only need to send messages in :D

oxzi (dismissed):
Thanks, looks good to me!

@yhabteab: Do you want to take another look, since you have started the review?

doc/01-About.md
```diff
@@ -35,6 +35,7 @@ Icinga Notifications comes with multiple channels out of the box:
 
 * _email_: Email submission via SMTP
 * _rocketchat_: Rocket.Chat
+* _telegram_: Telegram
```
yhabteab: Please make this a link to your newly introduced section in the `10-Channels.md` file.

yhabteab (changes requested):
While you're addressing the below comment, can you also make sure that the sentences of your comments in the code ends with a period (`.`). Currently, it's mixed.

DsgnrFH:
After discussing this with @yhabteab we decided that it would be better for the messages to reply not to the first message of the incident, but rather to the last message addressing this incident

yhabteab (approved):

PR #505, opened 2026-08-31
Remov every trace of `icingaweb2_url` option

### LINKED ISSUE #501: Drop config option `icingaweb2_url`
The *mandatory* configuration option `icingaweb2_url` is currently used in the following cases:

1. To construct a URL to the incident detail in Notifications Web, passed to a channel when transmitting a notification
2. To fully qualify an event's object URL, in case it's not empty and relative

Given the updated goal of what Notifications Web should represent, an administration and configuration interface, this must change. The contact who's receiving the notification does not necessarily have access to Notifications Web, as the role to visualize the incident has shifted to the source that delivered the event. A notification should accordingly be updated and only include the event's object URL.

The object URL on the other hand must be absolute at all times, and the `/process-event` API endpoint should make sure that is the case and complain otherwise.

DsgnrFH:
resolves #501


doc/04-Upgrading.md
```diff
@@ -78,8 +78,13 @@ Starting with version v1.0, the Icinga Notifications configuration options have
...
 `listener` section. If you are upgrading to version v1.0 from an earlier version, you will need to update your existing
 configuration file to reflect this change. Also, the `debug-password`, `debug-password_file`, and `listen` options have
-been renamed to `debug_password`, `debug_password_file`, and `address` respectively. Last but not least, the
-`icingaweb2-url` and `channels-dir` global options have been renamed to `icingaweb2_url` and `channels_dir` as well.
+been renamed to `debug_password`, `debug_password_file`, and `address` respectively. The `channels-dir` global option
+has been renamed to `channels_dir` as well.
+
+The `icingaweb2-url` global option has been dropped without replacement and you will have to remove it from your config.
+Icinga Notifications no longer builds any URLs on its own, so the events submitted by a source must now have an
+absolute `url` field. The `/process-event` endpoint rejects events with a relative `url`. If you are running a source
```
yhabteab: As you already noticed and documented in the other file, the `url` field is optional but the _events submitted by source must now have an absolute `url` field_ implies as if it's required.

yhabteab (changes requested):
Apart from the below comment, please also remove the following line:
https://github.com/Icinga/icinga-notifications/blob/076907e1738311a33d4e40e111badb07d85d0040/.github/workflows/tests_with_database.yml#L10

doc/04-Upgrading.md
DsgnrFH: I adjusted it

yhabteab:
Looks good now, but we have to merge https://github.com/Icinga/icinga-go-library/pull/239 first and bump the IGL version to main here.

DsgnrFH:
Bumped the icinga-go-library version after the merge

yhabteab (approved):

PR #507, opened 2026-09-02
Make the SMTP authentication mechanism configurable

### LINKED ISSUE #474: Feature Request: Support SMTP AUTH LOGIN in addition to PLAIN
Hi,

currently the SMTP notification channel appears to only support the PLAIN authentication mechanism.

Some SMTP server configurations require the use of LOGIN authentication. Without support for this mechanism, it is currently not possible to use the SMTP notification channel in such setups without modifying the code or changing the mail server configuration.

It would be great if the SMTP client could also support LOGIN authentication.
My suggestion would be to provide a configuration option to explicitly select the authentication mechanism in the notification channel configuration.

Thanks for considering

Max

DsgnrFH:
Resolves #474

### DIFF
--- a/cmd/channels/email/main.go
+++ b/cmd/channels/email/main.go
@@ -12,6 +12,7 @@ import (
 	"log/slog"
 	"net"
 	"net/mail"
+	"strconv"
 	"sync"
 
 	"github.com/emersion/go-sasl"
@@ -33,6 +34,8 @@ const (
 	EncryptionNone     = "none"
 	EncryptionStartTLS = "starttls"
 	EncryptionTLS      = "tls"
+
+	AuthMethodNone = "none"
 )
 
 type Email struct {
@@ -43,6 +46,8 @@ type Email struct {
 	User       string `json:"user"`
 	Password   string `json:"password"` // #nosec G117 -- exported password field
 	Encryption string `json:"encryption"`
+	// AuthMethod is a SASL Authentication Mechanism.
+	AuthMethod string `json:"auth_method"`
 
 	mu sync.Mutex // Protects access to the above fields.
 
@@ -147,6 +152,29 @@ func (ch *Email) GetInfo() *plugin.Info {
 				EncryptionTLS:      "TLS",
 			},
 		},
+		{
+			Name:     "auth_method",
+			Type:     "option",
+			Required: true,
+			Label: map[string]string{
+				"en_US": "SMTP Authentication Method",
+				"de_DE": "SMTP Authentifizierungsmethode",
+			},
+			Help: map[string]string{
+				"en_US": "The method has to be supported by the SMTP server. PLAIN and LOGIN require the SMTP user " +
+					"and password, while OAUTHBEARER expects a bearer token in the SMTP password field. " +
+					"If the NONE method is chosen, the SMTP server will be contacted without authentication.",
+				"de_DE": "Die Methode muss vom SMTP Server unterstützt werden. PLAIN und LOGIN benötigen den SMTP " +
+					"Benutzer und das Passwort, während OAUTHBEARER ein Bearer Token im SMTP Passwortfeld erwartet. " +
+					"Ist die NONE Methode gewählt, wird der SMTP Server ohne Authentifizierung kontaktiert.",
+			},
+			Options: map[string]string{
+				AuthMethodNone:   "NONE",
+				sasl.Plain:       sasl.Plain,
+				sasl.Login:       sasl.Login,
+				sasl.OAuthBearer: sasl.OAuthBearer,
+			},
+		},
 	}
 
 	return &plugin.Info{
@@ -169,8 +197,24 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
 		return fmt.Errorf("failed to load config: %s %w", jsonStr, err)
 	}
 
-	if (tmpEm.User == "") != (tmpEm.Password == "") {
-		return fmt.Errorf("user and password fields must both be set or empty")
+	// Channels configured before the authentication method became selectable carry no auth_method field and were
+	// authenticated with PLAIN whenever an SMTP password was set.
+	if tmpEm.AuthMethod == "" && (tmpEm.User != "" || tmpEm.Password != "") {
+		tmpEm.AuthMethod = sasl.Plain
+	}
+
+	if tmpEm.AuthMethod != "" {
+		switch tmpEm.AuthMethod {
+		case AuthMethodNone:
+			// Nothing to validate for this method.
+		case sasl.Plain, sasl.Login, sasl.OAuthBearer:
+			// All three carry a username and a secret, the latter being a bearer token for OAUTHBEARER.
+			if tmpEm.User == "" || tmpEm.Password == "" {
+				return fmt.Errorf("user and password fields are required for the %s authentication method", tmpEm.AuthMethod)
+			}
+		default:
+			return fmt.Errorf("unsupported SMTP authentication method %q", tmpEm.AuthMethod)
+		}
 	}
 
 	ch.mu.Lock()
@@ -183,6 +227,7 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
 	ch.User = tmpEm.User
 	ch.Password = tmpEm.Password
 	ch.Encryption = tmpEm.Encryption
+	ch.AuthMethod = tmpEm.AuthMethod
 
 	return nil
 }
@@ -253,10 +298,12 @@ func (ch *Email) Send(reversePath string, recipients []string, msg []byte) error
 	)
 
 	ch.mu.Lock()
-	serverAddr := net.JoinHostPort(ch.Host, ch.Port)
+	host, port := ch.Host, ch.Port
+	serverAddr := net.JoinHostPort(host, port)
 	encryption := ch.Encryption
 	password := ch.Password
 	username := ch.User
+	authMethod := ch.AuthMethod
 	ch.mu.Unlock()
 
 	switch encryption {
@@ -274,8 +321,30 @@ func (ch *Email) Send(reversePath string, recipients []string, msg []byte) error
 	}
 	defer func() { _ = client.Close() }()
 
-	if password != "" {
-		if err = client.Auth(sasl.NewPlainClient("", username, password)); err != nil {
+	if authMethod != "" && authMethod != AuthMethodNone {
+		var auth sasl.Client
+		switch authMethod {
+		case sasl.Plain:
+			auth = sasl.NewPlainClient("", username, password)
+		case sasl.Login:
+			auth = sasl.NewLoginClient(username, password)
+		case sasl.OAuthBearer:
+			smtpPort, err := strconv.Atoi(port)
+			if err != nil {
+				return fmt.Errorf("invalid SMTP port %q: %w", port, err)
+			}
+
+			auth = sasl.NewOAuthBearerClient(&sasl.OAuthBearerOptions{
+				Username: username,
+				Token:    password,
+				Host:     host,
+				Port:     smtpPort,
+			})
+		default:
+			return fmt.Errorf("unsupported SMTP authentication method %q", authMethod)
+		}
+
+		if err := client.Auth(auth); err != nil {
 			return err
 		}
 	}
--- a/cmd/channels/email/main_test.go
+++ b/cmd/channels/email/main_test.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"github.com/emersion/go-sasl"
 	"github.com/stretchr/testify/assert"
 	"testing"
 )
@@ -55,6 +56,56 @@ func TestEmail_SetConfig(t *testing.T) {
 			name:    "user-but-missing-pass",
 			jsonMsg: `{"user": "foo"}`,
 			wantErr: true,
+		},
+		{
+			name:    "login-without-credentials",
+			jsonMsg: `{"auth_method":"LOGIN"}`,
+			wantErr: true,
+		},
+		{
+			name:    "oauthbearer-without-credentials",
+			jsonMsg: `{"auth_method":"OAUTHBEARER"}`,
+			wantErr: true,
+		},
+		{
+			name:    "login-with-credentials",
+			jsonMsg: `{"auth_method":"LOGIN","user":"foo","password":"bar"}`,
+			want: &Email{
+				SenderName: "Icinga",
+				User:       "foo",
+				Password:   "bar",
+				AuthMethod: sasl.Login,
+			},
+		},
+		{
+			name:    "external-is-not-supported",
+			jsonMsg: `{"auth_method":"EXTERNAL"}`,
+			wantErr: true,
+		},
+		{
+			name:    "unsupported-auth-method",
+			jsonMsg: `{"auth_method":"foo"}`,
+			wantErr: true,
+		},
+		{
+			name:    "plain-without-credentials",
+			jsonMsg: `{"auth_method":"PLAIN"}`,
+			wantErr: true,
+		},
+		{
+			name:    "empty-auth-method-is-unauthenticated",
+			jsonMsg: `{"auth_method":""}`,
+			want:    &Email{SenderName: "Icinga"},
+		},
+		{
+			name:    "legacy-config-with-credentials",
+			jsonMsg: `{"user":"foo","password":"bar"}`,
+			want: &Email{
+				SenderName: "Icinga",
+				User:       "foo",
+				Password:   "bar",
+				AuthMethod: sasl.Plain,
+			},
 		},
 	}
 	for _, tt := range tests {


cmd/channels/email/main.go
```diff
@@ -35,6 +35,15 @@ const (
...
 
+const (
+	// AuthMechanismAuto picks the SASL mechanism.
+	AuthMechanismAuto = "auto"
+	// AuthMechanismPlain enforces the SASL PLAIN mechanism.
+	AuthMechanismPlain = "plain"
+	// AuthMechanismLogin enforces the still widely deployed SASL LOGIN mechanism
+	AuthMechanismLogin = "login"
+)
+
```
yhabteab: Do you really need these constants? Why not just use the ones from the `sasl` package instead?

cmd/channels/email/main.go
```diff
@@ -43,6 +52,8 @@ type Email struct {
 	User       string `json:"user"`
 	Password   string `json:"password"` // #nosec G117 -- exported password field
 	Encryption string `json:"encryption"`
+	// AuthMechanism is one of: AuthMechanismAuto, AuthMechanismPlain or AuthMechanismLogin.
+	AuthMechanism string `json:"auth_mechanism"`
```
yhabteab: It should be called `auth_method` instead.

cmd/channels/email/main.go
```diff
@@ -275,7 +307,12 @@ func (ch *Email) Send(reversePath string, recipients []string, msg []byte) error
 	defer func() { _ = client.Close() }()
 
 	if password != "" {
-		if err = client.Auth(sasl.NewPlainClient("", username, password)); err != nil {
+		auth, err := saslClient(client, authMechanism, username, password)
```
yhabteab: This is now guarded by the password check, but shouldn't we already discussed offline that atleast the `sasl.OAuthBearer` auth method should be supported as well, which doesn't require a password?

cmd/channels/email/main.go
```diff
@@ -287,6 +324,29 @@ func (ch *Email) Send(reversePath string, recipients []string, msg []byte) error
 	return client.Quit()
 }
 
+// saslClient returns a sasl.Client for the requested authentication mechanism.
+// For AuthMechanismAuto, the mechanism is selected from those advertised by the SMTP server, preferring PLAIN over the
+// LOGIN mechanism. This requires an already greeted client, which is the case after the first contact.
+func saslClient(client *smtp.Client, mechanism, username, password string) (sasl.Client, error) {
+	switch mechanism {
+	case AuthMechanismAuto:
```
yhabteab: Please drop the auto options. The user should know which auth method is supported by the SMTP server he's planning to use. No, need to figure this out on our own.

yhabteab (changes requested):

cmd/channels/email/main.go
DsgnrFH: Yee... I will update it

cmd/channels/email/main.go
DsgnrFH: Sure, will be changed

cmd/channels/email/main.go
DsgnrFH: Well technically it does require a password - a token, but I changed things up a little anyway

cmd/channels/email/main.go
```diff
@@ -147,6 +149,27 @@ func (ch *Email) GetInfo() *plugin.Info {
...
+				"de_DE": "SMTP Authentifizierungsmethode",
+			},
+			Help: map[string]string{
+				"en_US": "Only used when SMTP credentials are set. It has to be supported by the SMTP server. " +
+					"OAUTHBEARER expects the bearer token in the SMTP password field.",
+				"de_DE": "Wird nur verwendet, wenn SMTP Zugangsdaten gesetzt sind. Die Methode muss vom SMTP Server " +
+					"unterstützt werden. OAUTHBEARER erwartet das Bearer Token im SMTP Passwortfeld.",
+			},
+			Options: map[string]string{
+				sasl.Plain:       "PLAIN",
```
yhabteab:
```suggestion
				sasl.Plain:       sasl.Plain,
```

Please adjust the other options as well.

cmd/channels/email/main.go
```diff
@@ -147,6 +149,27 @@ func (ch *Email) GetInfo() *plugin.Info {
...
+				"en_US": "SMTP Authentication Method",
+				"de_DE": "SMTP Authentifizierungsmethode",
+			},
+			Help: map[string]string{
+				"en_US": "Only used when SMTP credentials are set. It has to be supported by the SMTP server. " +
+					"OAUTHBEARER expects the bearer token in the SMTP password field.",
+				"de_DE": "Wird nur verwendet, wenn SMTP Zugangsdaten gesetzt sind. Die Methode muss vom SMTP Server " +
+					"unterstützt werden. OAUTHBEARER erwartet das Bearer Token im SMTP Passwortfeld.",
+			},
+			Options: map[string]string{
```
yhabteab, line 171: This option should also support `sasl.External`.

yhabteab (changes requested):
Please also rebase against the latest main branch.

cmd/channels/email/main.go
DsgnrFH: Adjusted them too

cmd/channels/email/main.go
DsgnrFH: Okay, I added it

DsgnrFH:
Forgot to rerequest a rereview :/

cmd/channels/email/main.go
yhabteab: This new auth method requires a new optional text field called `Client Identity` and not mis-use the username for it.

DsgnrFH: I changed it in the [latest commit](https://github.com/Icinga/icinga-notifications/pull/507/changes/ba1bc77ec4599499fc73ab1d55bd1d23071f3cd4)

cmd/channels/email/main.go
yhabteab: You should also reject the config here when auth_method is set to `LOGIN` or `OAuthBearer` and  username or password is not provided. The comment from your newly introduced `saslClient` should be adjusted as well. It doesn't make any sense to allow the user to set the auth_method to `LOGIN` but then make password and username optional.

cmd/channels/email/main.go
```diff
@@ -147,6 +151,44 @@ func (ch *Email) GetInfo() *plugin.Info {
...
+		{
+			Name: "client_identity",
+			Type: "string",
+			Label: map[string]string{
+				"en_US": "Client Identity",
+				"de_DE": "Client-Identität",
+			},
+			Help: map[string]string{
+				"en_US": "Only used by the EXTERNAL authentication method. If left empty, the SMTP server " +
+					"determines the identity from the TLS client certificate.",
```
yhabteab: Where does the SMTP server gets the TLS client cert from? I guess, this depends on the `encription` option being set to either `EncryptionStartTLS` or `EncryptionTLS` with a properly configured client certificate, which we don't support yet. So, maybe make this option required if `auth_method` is set to `EXTERNAL`.

yhabteab (changes requested):

cmd/channels/email/main.go
DsgnrFH: Okay, I will change it

cmd/channels/email/main.go
```diff
@@ -173,6 +213,21 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
 		return fmt.Errorf("user and password fields must both be set or empty")
 	}
 
+	switch tmpEm.AuthMethod {
+	case sasl.Plain:
+		// PLAIN is the default and covers channels without any credentials
```
oxzi: This comment about PLAIN is confusing, as the PLAIN SASL Mechanism carries both a username and a password. It just adds no additional security mechanism and should thus only be used when the connection is wrapped in a secure transport, such as TLS.

https://datatracker.ietf.org/doc/html/rfc4616#section-2

So, I would expect a username and password check here as well.

cmd/channels/email/main.go
```diff
@@ -173,6 +213,21 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
 		return fmt.Errorf("user and password fields must both be set or empty")
 	}
```
oxzi: With the `switch` below, this check can now be merged in there, right?

cmd/channels/email/main.go
```diff
@@ -147,6 +151,42 @@ func (ch *Email) GetInfo() *plugin.Info {
...
+				"en_US": "The method has to be supported by the SMTP server. PLAIN and LOGIN use the SMTP password, " +
+					"OAUTHBEARER expects a bearer token in the SMTP password field and EXTERNAL authenticates " +
+					"through the TLS connection without using a password at all.",
+				"de_DE": "Die Methode muss vom SMTP Server unterstützt werden. PLAIN und LOGIN verwenden das SMTP " +
+					"Passwort, OAUTHBEARER erwartet ein Bearer Token im SMTP Passwortfeld und EXTERNAL " +
+					"authentifiziert über die TLS Verbindung, ohne ein Passwort zu verwenden.",
+			},
+			Options: map[string]string{
+				sasl.Plain:       sasl.Plain,
+				sasl.Login:       sasl.Login,
```
oxzi, line 174: FTR: LOGIN is quite obsolete and PLAIN should be used instead. The RFC expired in the process back in 2003.

https://datatracker.ietf.org/doc/draft-murchison-sasl-login/

cmd/channels/email/main.go
```diff
@@ -287,6 +350,31 @@ func (ch *Email) Send(reversePath string, recipients []string, msg []byte) error
 	return client.Quit()
 }
 
+// saslClient returns a sasl.Client for the requested authentication method.
+//
+// Both secret carrying methods take the password from the SMTP password field, while
+// sasl.OAuthBearer expects a token in there. Only sasl.External needs no secret, as it relies on the TLS connection
+// and takes the client identity instead. That every method has what it needs is already ensured by SetConfig, apart
+// from sasl.Plain without credentials. This is the unauthenticated channel, for which no client and no error is returned.
+func saslClient(method, username, secret, identity string) (sasl.Client, error) {
```
oxzi: Honest question: Why extract this `switch` into a helper function only called once? I would keep the code above, which eases the readability imo.

cmd/channels/email/main.go
```diff
@@ -287,6 +350,31 @@ func (ch *Email) Send(reversePath string, recipients []string, msg []byte) error
...
+	if method == sasl.Plain && secret == "" {
+		return nil, nil
+	}
```
oxzi: Pardon, but where did that come from? And shouldn't this be unreachable because of the check that either both username and password or none is set?

https://github.com/Icinga/icinga-notifications/blob/be42f51bcd851ad3037d8bd68ae737a46fa832e0/cmd/channels/email/main.go#L212-L214

cmd/channels/email/main.go
```diff
@@ -147,6 +151,42 @@ func (ch *Email) GetInfo() *plugin.Info {
...
+					"through the TLS connection without using a password at all.",
+				"de_DE": "Die Methode muss vom SMTP Server unterstützt werden. PLAIN und LOGIN verwenden das SMTP " +
+					"Passwort, OAUTHBEARER erwartet ein Bearer Token im SMTP Passwortfeld und EXTERNAL " +
+					"authentifiziert über die TLS Verbindung, ohne ein Passwort zu verwenden.",
+			},
+			Options: map[string]string{
+				sasl.Plain:       sasl.Plain,
+				sasl.Login:       sasl.Login,
+				sasl.OAuthBearer: sasl.OAuthBearer,
+				sasl.External:    sasl.External,
```
oxzi: How should the EXTERNAL SASL Mechanism work here? Unless one relies on IPSec or some IP oracle, I would expect TLS client certificates. However, there is no support for them yet. What is the idea here?

https://datatracker.ietf.org/doc/html/rfc4422#appendix-A

cmd/channels/email/main.go
```diff
@@ -173,6 +213,21 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
...
 
+	switch tmpEm.AuthMethod {
+	case sasl.Plain:
+		// PLAIN is the default and covers channels without any credentials
+	case sasl.Login, sasl.OAuthBearer:
+		if tmpEm.User == "" || tmpEm.Password == "" {
+			return fmt.Errorf("user and password fields are required for the %s authentication method", tmpEm.AuthMethod)
+		}
+	case sasl.External:
+		if tmpEm.ClientIdentity == "" {
```
oxzi: Based on RFC 4422 A.1, the authorization identity string can be left empty.

https://datatracker.ietf.org/doc/html/rfc4422#appendix-A.1

cmd/channels/email/main.go
```diff
@@ -43,6 +43,10 @@ type Email struct {
 	User       string `json:"user"`
 	Password   string `json:"password"` // #nosec G117 -- exported password field
 	Encryption string `json:"encryption"`
+	// AuthMethod is one of sasl.Plain, sasl.Login, sasl.OAuthBearer or sasl.External.
```
oxzi: Please don't list the currently supported SASL Mechanisms - which might become outdated -, but describe the field.


```suggestion
	// AuthMethod is a SASL Authentication Mechanism.
```

cmd/channels/email/main.go
```diff
@@ -287,6 +350,31 @@ func (ch *Email) Send(reversePath string, recipients []string, msg []byte) error
...
+		return nil, nil
+	}
+
+	switch method {
+	case sasl.Plain:
+		return sasl.NewPlainClient("", username, secret), nil
+	case sasl.Login:
+		return sasl.NewLoginClient(username, secret), nil
+	case sasl.OAuthBearer:
+		return sasl.NewOAuthBearerClient(&sasl.OAuthBearerOptions{Username: username, Token: secret}), nil
```
oxzi: What about the Host and Port struct fields? Are they omitted on purpose?

Consider the example in the RFC:
https://datatracker.ietf.org/doc/html/rfc7628#section-4.1

oxzi (commented):
Thanks for creating this PR.

Please explain your change briefly in the commit message, which gets imported as the PR description. At the moment, it only references an issue which just asks about the LOGIN SASL Mechanism, but this PR does way more.

Also, please squash the review commits.

cmd/channels/email/main.go
DsgnrFH: Changed it

cmd/channels/email/main.go
DsgnrFH: right, merged it

cmd/channels/email/main.go
DsgnrFH: noted

cmd/channels/email/main.go
DsgnrFH: I inlined it again.

cmd/channels/email/main.go
DsgnrFH: It needs cert support. Theres no way to present one right now

cmd/channels/email/main.go
DsgnrFH: I made it optional

cmd/channels/email/main.go
DsgnrFH: Applied

cmd/channels/email/main.go
DsgnrFH: Added both, thanks

cmd/channels/email/main.go
yhabteab: > FTR: LOGIN is quite obsolete and PLAIN should be used instead. The RFC expired in the process back in 2003.

The OP of the issue explicitly requested to add support for LOGIN, so I don't think, we're in a position to enforce that. If users want to use that auth method, then let them use (no additional work needed on our side).

cmd/channels/email/main.go
yhabteab: Ok. Adding TLS config support in this PR won't work due to time constraint but we'll definitely add a support for it in the future, so I've created a separate issue for it https://github.com/Icinga/icinga-notifications/issues/523. You can now remove all the `sasl.External` related code from this PR.

cmd/channels/email/main.go
```diff
@@ -169,8 +218,32 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
...
 	}
 
-	if (tmpEm.User == "") != (tmpEm.Password == "") {
-		return fmt.Errorf("user and password fields must both be set or empty")
+	// Channels configured before the authentication method became selectable carry no auth_method field and were
+	// authenticated with PLAIN whenever an SMTP password was set. As a default cannot express both cases, the raw
+	// config is checked for the field's presence to keep those channels working the way they did.
+	var stored struct {
+		AuthMethod *string `json:"auth_method"`
+	}
```
yhabteab: > As a default cannot express both cases

Do not provide a default value for it then and make the `auth_method` optional and make the type of `Email.AuthMethod` a string pointer just like you did here. That way, you won't need this intermediate struct but can just be used exactly how you're currently used `stored.AuthMethod`.

yhabteab (changes requested):

cmd/channels/email/main.go
DsgnrFH: Okay

DsgnrFH:
I also squashed the commits

cmd/channels/email/main.go
```diff
@@ -169,8 +197,24 @@ func (ch *Email) SetConfig(jsonStr json.RawMessage) error {
...
+	// Channels configured before the authentication method became selectable carry no auth_method field and were
+	// authenticated with PLAIN whenever an SMTP password was set.
+	if tmpEm.AuthMethod == "" && (tmpEm.User != "" || tmpEm.Password != "") {
+		tmpEm.AuthMethod = sasl.Plain
+	}
+
+	if tmpEm.AuthMethod != "" {
+		switch tmpEm.AuthMethod {
+		case AuthMethodNone:
+			// Nothing
```
yhabteab:
```suggestion
		case AuthMethodNone:
			// Nothing to validate for this method.

```

cmd/channels/email/main.go
```diff
@@ -147,6 +152,29 @@ func (ch *Email) GetInfo() *plugin.Info {
 				EncryptionTLS:      "TLS",
 			},
 		},
+		{
+			Name: "auth_method",
+			Type: "option",
```
yhabteab: Please fix your IDE to prevent from pushing such things.

```suggestion
			Name:     "auth_method",
			Type:     "option",
```

cmd/channels/email/main.go
```diff
@@ -147,6 +152,29 @@ func (ch *Email) GetInfo() *plugin.Info {
...
+			Help: map[string]string{
+				"en_US": "The method has to be supported by the SMTP server. PLAIN and LOGIN require the SMTP user " +
+					"and password, while OAUTHBEARER expects a bearer token in the SMTP password field. " +
+					"Without a method, the SMTP server is contacted without authenticating.",
+				"de_DE": "Die Methode muss vom SMTP Server unterstützt werden. PLAIN und LOGIN benötigen SMTP " +
+					"Benutzer und Passwort, OAUTHBEARER erwartet ein Bearer Token im SMTP Passwortfeld. " +
+					"Ohne Methode wird der SMTP Server ohne Authentifizierung kontaktiert.",
+			},
+			Options: map[string]string{
+				AuthMethodNone:   "None",
```
yhabteab: Please make this in uppercase as well, otherwise, it looks odd when rendered in the UI.

Image

yhabteab (changes requested):

cmd/channels/email/main.go
yhabteab: Is removed from this PR now.

cmd/channels/email/main.go
DsgnrFH: Good suggestion

cmd/channels/email/main.go
```diff
@@ -147,6 +152,29 @@ func (ch *Email) GetInfo() *plugin.Info {
...
+				"en_US": "SMTP Authentication Method",
+				"de_DE": "SMTP Authentifizierungsmethode",
+			},
+			Help: map[string]string{
+				"en_US": "The method has to be supported by the SMTP server. PLAIN and LOGIN require the SMTP user " +
+					"and password, while OAUTHBEARER expects a bearer token in the SMTP password field. " +
+					"Without a method, the SMTP server is contacted without authenticating.",
+				"de_DE": "Die Methode muss vom SMTP Server unterstützt werden. PLAIN und LOGIN benötigen SMTP " +
+					"Benutzer und Passwort, OAUTHBEARER erwartet ein Bearer Token im SMTP Passwortfeld. " +
+					"Ohne Methode wird der SMTP Server ohne Authentifizierung kontaktiert.",
```
yhabteab:
```suggestion
				"de_DE": "Die Methode muss vom SMTP Server unterstützt werden. PLAIN und LOGIN benötigen den SMTP " +
					"Benutzer und das Passwort, während OAUTHBEARER ein Bearer Token im SMTP Passwortfeld erwartet. " +
					"Ist die NONE Methode gewählt, wird der SMTP Server ohne Authentifizierung kontaktiert.",
```

cmd/channels/email/main.go
```diff
@@ -147,6 +152,29 @@ func (ch *Email) GetInfo() *plugin.Info {
...
+			Type:     "option",
+			Required: true,
+			Label: map[string]string{
+				"en_US": "SMTP Authentication Method",
+				"de_DE": "SMTP Authentifizierungsmethode",
+			},
+			Help: map[string]string{
+				"en_US": "The method has to be supported by the SMTP server. PLAIN and LOGIN require the SMTP user " +
+					"and password, while OAUTHBEARER expects a bearer token in the SMTP password field. " +
+					"Without a method, the SMTP server is contacted without authenticating.",
```
yhabteab:
```suggestion
					"If the NONE method is chosen, the SMTP server will be contacted without authentication.",
```

yhabteab (changes requested):
Sorry, I didn't send this with my earlier comments. Apart from the below two comments, looks fine to me now.

cmd/channels/email/main.go
DsgnrFH: Thanks, I had to translate either on my own, or with help of Deepl and it was mixed

cmd/channels/email/main.go
DsgnrFH: Thanks for it too, I will update it now

yhabteab (approved):

cmd/channels/email/main.go
oxzi: The upstream `go-sasl` library started dropping LOGIN for the same reasons: https://github.com/emersion/go-sasl/commit/b788ff22d5a6b3970cde181998f52658a475bffc and https://github.com/emersion/go-sasl/issues/19. When we have implemented this and upstream drops it, then it means additional work for us.

yhabteab: > The upstream `go-sasl` library started dropping LOGIN for the same reasons

Started :). It's open since 2020, so there's a reason for that too, I guess.

> When we have implemented this and upstream drops it, then it means additional work for us.

That would be a trivial change. Also we might just even fork the LOGIN client in our email channel instead of dropping it just because the library we use doesn't support it anymore. I'm just saying, if our users have SMTP servers running that require LOGIN, why should we enforce them to switch to PLAIN? Just so they can use our mail channel? I don't think, we are not the one that should enfoce them to switch to something else just because we think, it's obsolete now. The referenced issue clearly shows a need for this since he explicitly requested for LOGIN method.

PR #511, opened 2026-09-07
Support is_managed as an escalation condition

### LINKED ISSUE #478: Support *is managed* as a new escalation condition
This is based on Icinga 2 currently supporting expire times for acknowledgements. My idea is to not provide this option with the new integration in place and instead rely on escalations to act as alternative.

Many users use the expiry time for acknowledgements to avoid having a host or service that is permanently muted because no-one feels responsible anymore.

To let escalations be an alternative to this, we have to support a new condition so that it can be coupled together with the incident's age we already can act on. The suggestion is a new condition `is_managed` which takes a bool (i.e. `y|n`).

DsgnrFH:
Resolves #478

### DIFF
--- a/internal/incident/incident.go
+++ b/internal/incident/incident.go
@@ -641,7 +641,11 @@ func (i *Incident) evaluateEscalations(eventTime time.Time) ([]*rule.Escalation,
 		i.EscalationState = make(map[int64]*EscalationState)
 	}
 
-	filterContext := &rule.EscalationFilter{IncidentAge: eventTime.Sub(i.StartedAt.Time()), IncidentSeverity: i.Severity}
+	filterContext := &rule.EscalationFilter{
+		IncidentAge:      eventTime.Sub(i.StartedAt.Time()),
+		IncidentSeverity: i.Severity,
+		IsManaged:        i.HasManager(),
+	}
 
 	var escalations []*rule.Escalation
 	retryAfter := rule.RetryNever
--- a/internal/rule/condition.go
+++ b/internal/rule/condition.go
@@ -2,10 +2,11 @@ package rule
 
 import (
 	"fmt"
-	"github.com/icinga/icinga-go-library/notifications/event"
-	"github.com/icinga/icinga-notifications/internal/filter"
 	"math"
 	"time"
+
+	"github.com/icinga/icinga-go-library/notifications/event"
+	"github.com/icinga/icinga-notifications/internal/filter"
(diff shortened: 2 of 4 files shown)


internal/incident/incident.go
```diff
@@ -36,14 +36,18 @@ type Incident struct {
...
 	//
 	// For example, if there are escalations configured for incident_age>1h and incident_age>2h.
 	//   - If the incident is less than an hour old, this field contains the incident start time plus one hour.
 	//   - If the incident is between 1h and 2h old, this field contains the incident start time plus two hours.
 	//   - If the incident is already older than 2h, no future escalations can be reached solely based on the incident
 	//     aging, so the field is left nil.
+	//
+	// Escalations referring to is_managed are an exception, as their outcome doesn't only depend on time. Once
+	// such escalation cannot be reached by the incident aging alone anymore, this field is set to a short
+	// interval in the future, so that the escalation is polled until it either triggers or the incident recovers.
```
yhabteab: Why did you change this? The referenced issue never mentioned about this new condition being time dependent. Please revert this and all related changes. The `is_managed` condition should evaluate to true when the respective incident was managed at the time of its evaluation just like the `IncidentSeverity` condition. That's it. You don't have retry anything by your own.

yhabteab (changes requested):

internal/incident/incident.go
DsgnrFH: Okay, then I got confused with the issue at first. I will fix it

DsgnrFH: Squashed it too

internal/rule/condition.go
```diff
@@ -14,6 +14,19 @@ const RetryNever = time.Duration(math.MaxInt64)
...
+
+func parseManagedValue(value any) (bool, error) {
+	switch v := fmt.Sprint(value); v {
+	case "y":
+		return true, nil
+	case "n":
+		return false, nil
+	default:
+		return false, fmt.Errorf(`invalid is_managed value %q, expected either "y" or "n"`, v)
+	}
```
yhabteab: Please always move helper function to the end of the file, especially those that are not exported.

DsgnrFH:
The conflicts have to be solved first :)

yhabteab:
I'll take a look at the failing tests. You can just work on your other PR.

yhabteab (approved):

PR #517, opened 2026-09-08
retention: add basic test cases for all pruners

yhabteab:
This PR fixes one bug in `ResetPruner.Exec` with 043c350 discovered by the newly added tests and adds some basic testing for all the pruners to ensure the used SQL queries run fine with all the supported DBMs versions.

### DIFF
--- a/internal/retention/prune.go
+++ b/internal/retention/prune.go
@@ -312,12 +312,16 @@ func (orp *OrphanRowPruner) assembleSelect(limit uint64) string {
 func (orp *OrphanRowPruner) assembleDelete(driverName string, limit uint64) string {
 	switch driverName {
 	case database.MySQL:
-		joins, conds := orp.joinList()
-		// MariaDB does support JOIN based ANTI-JOINs but doesn't support LIMIT in DELETE statements with JOINs
-		// on older versions, so we have to use a subquery instead.
+		// MariaDB does support JOIN based ANTI-JOINs but doesn't support LIMIT in DELETE statements with
+		// JOINs on older versions (and even latest MySQL versions), so we have to use a subquery instead.
+		// And the nested subquery is required to avoid the (MySQL only) "You can't specify target table
+		// for update in FROM clause" error.
 		return fmt.Sprintf(
-			`DELETE FROM %[1]s WHERE %[2]s IN (SELECT main.%[2]s FROM %[1]s main %[3]s WHERE %[4]s) LIMIT %[5]d`,
-			orp.Table, orp.PKorFK, joins, conds, limit)
+			`DELETE FROM %[1]s WHERE %[2]s IN (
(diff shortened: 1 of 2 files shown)


yhabteab:
The GHA actions are stuck now because the Icinga 2 repo seems to be very busy with lots of PR sthat were just opened/merged.

yhabteab:
Apparently, the tests also triggered a MySQL specific error 46d42a9.

internal/retention/prune.go
```diff
@@ -303,11 +313,16 @@ func (orp *OrphanRowPruner) assembleDelete(driverName string, limit uint64) stri
...
+		// MariaDB does support JOIN based ANTI-JOINs but doesn't support LIMIT in DELETE statements with
+		// JOINs on older versions (and even latest MySQL versions), so we have to use a subquery instead.
+		// And the nested subquery is required to avoid the (MySQL only) "You can't specify target table
+		// for update in FROM clause" error.
 		return fmt.Sprintf(
-			`DELETE FROM %[1]s WHERE %[2]s IN (SELECT main.%[2]s FROM %[1]s main %[3]s WHERE %[4]s) LIMIT %[5]d`,
-			orp.Table, orp.PKorFK, joins, conds, limit)
+			`DELETE FROM %[1]s WHERE %[2]s IN (
+				SELECT sub_tmp_table.%[2]s FROM (
+					SELECT main.%[2]s FROM %[1]s main %[3]s WHERE %[4]s LIMIT %[5]d
```
oxzi: Isn't this query exactly like `assembleSelect`, which could be called here as in the PostgreSQL case below?

internal/retention/prune.go
```diff
@@ -147,26 +147,36 @@ type ResetPruner struct {
...
+					updated, err = execCascade(ctx, db, tx, rp, limit, olderThan)
+					return
+				})
+			},
+			retry.Retryable,
+			backoff.DefaultBackoff,
+			getPrunerRetrySettings(l),
+		)
+		if err != nil {
+			return 0, err
```
oxzi, line 172: Might be a bit stupid, but shouldn't it be `return total, err` as each update happens in its own transaction. For example, if the second retryable transaction fails, the first had already succeeded. Not like its worth much tho.

oxzi (commented):
Thanks, good catch! Only two potential nits.

internal/retention/prune.go
yhabteab: We're returning an error, and as it's the case and the way it should be, every other return value are irrelevant and gets ignored, and this doesn't make sense to return anyhting other than the zero value of the type.

internal/retention/prune.go
yhabteab: You're right. Fixed.

oxzi (approved):
⛵

PR #520, opened 2026-09-09
GHA: bump DB version matrix & replace PostgreSQL procedure with function

yhabteab:
This PR bumps the DB version matrix for both DBMs. Additionally, it also replaces our existing PROCEDURE add in https://github.com/Icinga/icinga-notifications/pull/422 with a FUNCTION for PostgreSQL as our schema file is no longer compatible with PostgreSQL versions older than 11 anymore.

```bash

Run psql -U postgres -w -h$ICINGA_NOTIFICATIONS_DATABASE_HOST -d $ICINGA_NOTIFICATIONS_DATABASE_DATABASE < /home/runner/work/icinga-notifications/icinga-notifications/schema/pgsql/schema.sql
CREATE EXTENSION
CREATE TYPE
CREATE TYPE
CREATE TYPE
CREATE TYPE
CREATE TYPE
CREATE FUNCTION
CREATE OPERATOR
ERROR:  syntax error at or near "PROCEDURE"
LINE 1: CREATE OR REPLACE PROCEDURE assert_correct_schema_version(ex...
                          ^
ERROR:  syntax error at or near "PROCEDURE"
LINE 1: COMMENT ON PROCEDURE assert_correct_schema_version IS 'Asser...
                   ^
CREATE TABLE
CREATE TABLE
CREATE INDEX
CREATE TABLE
CREATE INDEX
CREATE TABLE
CREATE INDEX
```

Also, looking at the PostgreSQL documentation also shows that [this page](https://www.postgresql.org/docs/11/sql-createprocedure.html) describing the PROCDURE feature exists only since PostgreSQL version 11.

### DIFF
--- a/.github/workflows/tests_with_database.yml
+++ b/.github/workflows/tests_with_database.yml
@@ -22,6 +22,8 @@ jobs:
         image:
           - mysql:8.0
           - mysql:8
+          - mysql:9.0
+          - mysql:9
           - mysql:latest
           - mariadb:10.5
           - mariadb:10.6
@@ -29,6 +31,8 @@ jobs:
           - mariadb:11.0
           - mariadb:11.1
           - mariadb:11.2
+          - mariadb:12.0
+          - mariadb:12
           - mariadb:latest
 
     env:
@@ -84,7 +88,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: ["9.6", "10", "11", "12", "13", "14", "15", "latest"]
+        version: ["9.6", "10", "11", "12", "13", "14", "15", "16", "17", "latest"]
 
     env:
       ICINGA_NOTIFICATIONS_DATABASE_TYPE: pgsql
@@ -119,7 +123,7 @@ jobs:
         env:
           PGPASSWORD: ${{ env.ICINGA_NOTIFICATIONS_DATABASE_PASSWORD }}
         run: |
-          psql -U postgres -w -h$ICINGA_NOTIFICATIONS_DATABASE_HOST -d $ICINGA_NOTIFICATIONS_DATABASE_DATABASE < ${{ github.workspace }}/schema/pgsql/schema.sql
+          psql -v ON_ERROR_STOP=1 -U postgres -w -h$ICINGA_NOTIFICATIONS_DATABASE_HOST -d $ICINGA_NOTIFICATIONS_DATABASE_DATABASE < ${{ github.workspace }}/schema/pgsql/schema.sql
 
       - name: Download dependencies
         run: go get -v -t -d ./...
--- a/doc/02-Installation.md
+++ b/doc/02-Installation.md
@@ -70,7 +70,7 @@ To apply these changes, run `systemctl reload postgresql`.
 After creating the database, import the Icinga Notifications schema using the following command:
 
 ```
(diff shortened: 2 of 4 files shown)


nilmerg (changes requested):
Raising a major by two steps is hard to justify IMHO due to a minor syntax difference. Please consider doing this instead:

```sql
CREATE OR REPLACE FUNCTION assert_correct_schema_version(expected_version text)
    RETURNS text
    LANGUAGE plpgsql
AS $$
DECLARE
    actual_version text := (SELECT version FROM notifications_schema ORDER BY timestamp DESC LIMIT 1);
BEGIN
    IF actual_version IS NULL THEN
        RAISE 'Schema version not found in notifications_schema table.';
    ELSIF actual_version != expected_version THEN
        RAISE 'Schema version mismatch: expected %, got %. Please apply all previous upgrade scripts in order before applying this one.', expected_version, actual_version;
    ELSE
        RETURN 'Schema version matches expected version: ' || expected_version;
    END IF;
END;
$$;
```

yhabteab:
> Raising a major by two steps is hard to justify IMHO due to a minor syntax difference. Please consider doing this instead:

Both PostgreSQL versions are way too old now and as listed in the PR description, even RHEL users can easily (if not already done) upgrade to even newer version than 11. But if you think, it's acceptable to have an output like this with each and every future upgrade script, then I don't mind changing it.

```sql
CREATE OR REPLACE FUNCTION assert_correct_schema_version(expected_version text)
    RETURNS void
    LANGUAGE plpgsql
    STABLE
    STRICT
    PARALLEL RESTRICTED
AS $$
DECLARE
    actual_version text;
BEGIN
    SELECT version INTO actual_version FROM notifications_schema ORDER BY timestamp DESC LIMIT 1;

    IF actual_version IS NULL THEN
        RAISE 'Schema version not found in notifications_schema table.';
    ELSIF actual_version != expected_version THEN
        RAISE 'Schema version mismatch: expected %, got %. Please apply all previous upgrade scripts in order before applying this one.', expected_version, actual_version;
    END IF;
END;
$$;
```

```bash
$ cat something.sql
SELECT assert_correct_schema_version('v1.0');

CREATE TABLE tmp_table (id bigint);
DROP TABLE tmp_table;
```

```bash
$ psql -U postgres notifications_test < something.sql
 assert_correct_schema_version
-------------------------------

(1 row)

CREATE TABLE
DROP TABLE
```

nilmerg (dismissed):
I think it's acceptable and the safer approach right now. And it's our own upgrade files anyway, so we may also change it to a procedure in the future.

yhabteab:
I've removed the COMMENT on the FUNCTION as PostgreSQL 9.6 doesn't seem to support it.
https://github.com/Icinga/icinga-notifications/actions/runs/34344465306/job/102442664481?pr=520

```bash
CREATE TYPE
CREATE TYPE
CREATE TYPE
CREATE TYPE
CREATE FUNCTION
CREATE OPERATOR
CREATE FUNCTION
ERROR:  syntax error at or near "IS"
LINE 1: COMMENT ON FUNCTION assert_correct_schema_version IS 'Assert...
                                                          ^
Error: Process completed with exit code 3.
```

nilmerg (approved):

Issue #523, opened 2026-09-09
Support SMTP TLS config for email channel


yhabteab:
Currently, if `STARTTLS` or just `TLS` is used as encription, the email will just use a TLS client without any certificates. This should be changed to allow configuring proper client certificate and CA. While implementing this, the `sasl.External` SMTP authentication method should also be added that can only be used if a valid client cert is provided.

PR #526, opened 2026-09-15
Conditional encryption options for the email channel

### LINKED ISSUE #514: Allow channel plugins to conditionally render config options in the UI
Currently, channel plugins can only provide a static set of configuration options for the user to fill out. This means that all options are always displayed, regardless of the values of other options. We want to allow channel plugins to conditionally render and hide configuration options based on the values of other options. This will make the configuration UI cleaner and easier to use without overwhelming the user with unnecessary options. To achieve this, I propose the following changes:

First, add a new type in [plugin package](https://github.com/Icinga/icinga-go-library/blob/main/notifications/plugin/plugin.go) called `ChildOption` which describes a config element that is conditionally displayed based on the value of a parent option. As opposed to the existing `ConfigOption` type, this new type has an additional field called `parent_values` which is a list of parent element values that will trigger the display of this child option. Here's a draft of how the `ChildOption` type could look like:

```go
// ChildOption describes a config element that is conditionally displayed based on the value of a parent option.
//
// This allows to create a more dynamic and user-friendly configuration interface, where certain options are only
// shown when relevant based on the user's previous selections. Even though the API allows for nested children, the
// UI will only render one level of children or even reject them entirely. Therefore, it is prohibited to define a
// child option with its own children.
type ChildOption struct {
	ConfigOption

	// ParentValues is a list of parent element values that will trigger the display of this child option.
	//
	// If the parent element is a select element, this list should contain the keys of the parent option that
	// should trigger this child option to be displayed. If the parent element is a checkbox, this list should
	// contain the boolean value (true or false). All other parent element types are not supported for child options
	// and will be rejected by Icinga Notifications Web.
	//
	// If empty, the child option will always be displayed regardless of the parent option's value. This is useful for
	// child options that should always be displayed, but are still logically grouped under a parent element.
	ParentValues []any `json:"parent_values"`
}
```

Next, the existing `ConfigOption` type gains a new field called `children` which is a slice of `ChildOption` entries. This field allows channel plugins to specify child options that are conditionally displayed based on their parent element's value. The `children` field is optional and can be omitted if there are no child options for a given config element. It's not allowed to have a child element with its own children. So, you've to either enforce this on the plugin side (which is not ideal since that will break existing plugins) or in Icinga Notifications Web when rendering the elements. 

```go
// Children contains a set of [ChildOption] rendered in the UI only when a specific parent option value is selected.
//
// This allows for conditional configuration options based on the user's selection in the parent option.
// For example, if a parent option is a select element with multiple choices, each choice can have its own
// set of child options that are displayed only when that choice is selected. A concrete example would be a
// parent option for selecting HTTP request method (GET, POST, PUT, DELETE), and each method could have its
// own set of child options for configuring headers, body content, etc.
//
// A child option must not have its own children, as this would create a recursive structure that is not supported.
Children []ChildOption `json:"children,omitempty"`
```

This example JSON snippet from the `Email` channel plugin demonstrates how to use the new `OptionGroup` feature. For this, I've patched the `Email#GetInfo()` method to include the new `groups` field in the `ConfigOption` entries for `ca_file`, `cert_file`, and `key_file`.

<details>
<summary>Here's the Patch</summary>

```diff
diff --git a/cmd/channels/email/main.go b/cmd/channels/email/main.go
index 13b94c8..5974955 100644
--- a/cmd/channels/email/main.go
+++ b/cmd/channels/email/main.go
@@ -146,6 +146,47 @@ func (ch *Email) GetInfo() *plugin.Info {
 				EncryptionStartTLS: "STARTTLS",
 				EncryptionTLS:      "TLS",
 			},
+			Children: []plugin.ChildOption{
+				{
+					Name: "ca_cert",
+					Type: "string",
+					Label: map[string]string{
+						"en_US": "CA Certificate",
+						"de_DE": "CA Zertifikat",
+					},
+					Help: map[string]string{
+						"en_US": "Optional CA certificate to verify the SMTP server's certificate.",
+						"de_DE": "Optionales CA Zertifikat zur Überprüfung des Zertifikats des SMTP Servers.",
+					},
+					ParentValues: []any{EncryptionStartTLS, EncryptionTLS},
+				},
+				{
+					Name: "client_cert",
+					Type: "string",
+					Label: map[string]string{
+						"en_US": "Client Certificate",
+						"de_DE": "Client Zertifikat",
+					},
+					Help: map[string]string{
+						"en_US": "Optional client certificate for mutual TLS authentication.",
+						"de_DE": "Optionales Client Zertifikat für gegenseitige TLS Authentifizierung.",
+					},
+					ParentValues: []any{EncryptionStartTLS, EncryptionTLS},
+				},
+				{
+					Name: "client_key",
+					Type: "secret",
+					Label: map[string]string{
+						"en_US": "Client Key",
+						"de_DE": "Client Schlüssel",
+					},
+					Help: map[string]string{
+						"en_US": "Optional client key for mutual TLS authentication.",
+						"de_DE": "Optionaler Client Schlüssel für gegenseitige TLS Authentifizierung.",
+					},
+					ParentValues: []any{EncryptionStartTLS, EncryptionTLS},
+				},
+			},
 		},
 	}
```

</details>

This is not the complete JSON output of the `Email` channel plugin, I've removed some of the other configuration options for brevity. The important part is the new `children` field in the `ConfigOption` entry for `encryption`. When the user selects either "STARTTLS" or "TLS" for it, these three elements (`ca_cert`, `client_cert`, `client_key`) can be displayed in the UI, otherwise they will be hidden. This allows for a cleaner and more user-friendly configuration experience.

```json
{
  "name": "Email",
  "version": "0.2.0",
  "author": "Icinga GmbH",
  "config_attrs": [
    {
      "name": "encryption",
      "type": "option",
      "label": {
        "de_DE": "SMTP Transportverschlüsselung",
        "en_US": "SMTP Transport Encryption"
      },
      "required": true,
      "options": {
        "none": "None",
        "starttls": "STARTTLS",
        "tls": "TLS"
      },
      "children": [
        {
          "name": "ca_cert",
          "type": "string",
          "label": {
            "de_DE": "CA Zertifikat",
            "en_US": "CA Certificate"
          },
          "help": {
            "de_DE": "Optionales CA Zertifikat zur Überprüfung des Zertifikats des SMTP Servers.",
            "en_US": "Optional CA certificate to verify the SMTP server's certificate."
          },
          "parent_values": [
            "starttls",
            "tls"
          ]
        },
        {
          "name": "client_cert",
          "type": "string",
          "label": {
            "de_DE": "Client Zertifikat",
            "en_US": "Client Certificate"
          },
          "help": {
            "de_DE": "Optionales Client Zertifikat für gegenseitige TLS Authentifizierung.",
            "en_US": "Optional client certificate for mutual TLS authentication."
          },
          "parent_values": [
            "starttls",
            "tls"
          ]
        },
        {
          "name": "client_key",
          "type": "secret",
          "label": {
            "de_DE": "Client Schlüssel",
            "en_US": "Client Key"
          },
          "help": {
            "de_DE": "Optionaler Client Schlüssel für gegenseitige TLS Authentifizierung.",
            "en_US": "Optional client key for mutual TLS authentication."
          },
          "parent_values": [
            "starttls",
            "tls"
          ]
        }
      ]
    }
  ]
}
```

starlingun:
Email channel configuration now includes several options for encryption: certificate of authority, client certificate and client key. Those are child options of the encryption type, TLS or STARTTLS, which will allow them to be rendered depending on user's choice.

Closes #514

Depends on 
- https://github.com/Icinga/icinga-go-library/pull/250


starlingun:
@cla-bot check

starlingun:
Closes #514

Jan-Schuppik (approved):
Looks good, as described in the issue 👍

Jan-Schuppik:
Don't forget to change the IGL (after the required IGL branch has been merged and before this is merged)

yhabteab (changes requested):
The examples I gave in #514 were just, well, examples to demonstrate clearly how that feature is going to be used. Though, the email channel doesn't support TLS settings yet (see #523) and you're just adding the options here without making any use of them. So, I suggest that you change the other settings as commented by @nilmerg in https://github.com/Icinga/icinga-notifications/issues/514#issuecomment-5696581830 in this PR, and if you still want to add TLS support for this channell, then you'll have to do it separately and assign #523 to yourself accordingly.

Also, I don't know how or why you did that, but you've changed the file mode of the `cmd/channels/email/main.go` file. Please undo that.

cmd/channels/email/main.go
```diff
@@ -169,11 +149,35 @@ func (ch *Email) GetInfo() *plugin.Info {
 					"Ist die NONE Methode gewählt, wird der SMTP Server ohne Authentifizierung kontaktiert.",
 			},
 			Options: map[string]string{
-				AuthMethodNone:   "NONE",
+				AuthMethodNone:   AuthMethodNone,
```
yhabteab: Please revert this.

cmd/channels/email/main.go
```diff
@@ -169,11 +149,35 @@ func (ch *Email) GetInfo() *plugin.Info {
...
+				{
+					Name: "password",
+					Type: "secret",
+					Label: map[string]string{
+						"en_US": "SMTP Password",
+						"de_DE": "SMTP Passwort",
+					},
+					ParentValues: []any{sasl.Plain, sasl.Login, sasl.OAuthBearer},
+				},
+			},
```
yhabteab, line 182: Since these two fields are rendered conditionally now, you can make them required (set the Required field to true).

yhabteab (changes requested):

cmd/channels/email/main.go
starlingun: done

yhabteab (approved):

